PR #14, opened 2020-11-30
Fix typo in consult.el

hkjels:


minad:
Oh yes, but it still does not sound right to me like this. I improved the docstring.

PR #17, opened 2020-12-01
Add annotations for some ‘customize-*’ commands.

okamsn:
Annotate these commands:
- customize-face
- customize-face-other-window
- customize-group
- customize-group-other-window
- customize-option
- customize-option-other-window
- customize-set-variable
- customize-variable
- customize-variable-other-window

Create annotation function ‘consult-annotate-customization-group’.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -947,6 +947,15 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
 (defcustom consult-annotate-alist
   '((execute-extended-command . consult-annotate-command-binding)
     (consult-apropos . consult-annotate-symbol)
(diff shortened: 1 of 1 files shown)


minad:
Thank you. Looks almost good. Maybe remove consult-annotate-group and replace by consult-annotate-symbol. Add the 'group property check to consult-annotate-symbol.

minad:
> Maybe remove consult-annotate-group and replace by consult-annotate-symbol. Add the 'group-documentation property check to consult-annotate-symbol.

I guess it is better to do it like this to ensure that we always use the correct property. I merge as is.

minad:
Merged in https://github.com/minad/consult/commit/b3dd023450b0143a7c801d835598bfa2d35f8b30

PR #22, opened 2020-12-01
Ignore registers without cdrs in ‘consult--register-candidates’.

okamsn:
Trying to use a garbage register (one without a cdr) will only raise an error that the register has no text.  Nothing is gained by listing them in the minibuffer, so they can be removed from the candidates list.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -665,15 +665,20 @@ Otherwise replace the just-yanked text with the selected text."
 
 (defun consult--register-candidates ()
   "Return alist of register descriptions and register names."
-  (mapcar
-   (lambda (r)
-     (setq r (car r))
-     (cons (format "%s: %s"
-                   (single-key-description r)
-                   (register-describe-oneline r))
-           r))
-   (or (sort (copy-sequence register-alist) #'car-less-than-car)
-       (user-error "All registers are empty"))))
+  (cl-remove-if
+   #'null
+   (mapcar
+    (lambda (r)
+      ;; Sometimes, registers are made without a `cdr'.  They don't do anything,
+      ;; and should be ignored.
+      (when (cdr r)
+        (setq r (car r))
+        (cons (format "%s: %s"
+                      (single-key-description r)
+                      (register-describe-oneline r))
+              r)))
+    (or (sort (copy-sequence register-alist) #'car-less-than-car)
+        (user-error "All registers are empty")))))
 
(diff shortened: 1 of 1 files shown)


consult.el
```diff
@@ -665,15 +665,20 @@ Otherwise replace the just-yanked text with the selected text."
...
-           r))
-   (or (sort (copy-sequence register-alist) #'car-less-than-car)
-       (user-error "All registers are empty"))))
+  (cl-remove-if
+   #'null
+   (mapcar
+    (lambda (r)
+      ;; Sometimes, registers are made without a `cdr'.  They don't do anything,
+      ;; and should be ignored.
+      (when (cdr r)
```
minad, line 674: Please use seq-remove instead of cl- and filter before the mapcar. Then we can avoid the when here.

okamsn: How is it now?

Also, what is the benefit of using `seq-remove/filter` over the `cl-lib` versions?

consult.el
```diff
@@ -665,20 +665,20 @@ Otherwise replace the just-yanked text with the selected text."
...
+     (setq r (car r))
+     (cons (format "%s: %s"
+                   (single-key-description r)
+                   (register-describe-oneline r))
+           r))
+   (or
+    ;; Sometimes, registers are made without a `cdr'.
+    ;; Such registers don't do anything, and can be ignored.
+    (sort (seq-filter (lambda (r) (cdr r))
+                      (copy-sequence register-alist))
```
minad, line 679: Thank you. Just noticed - I think we can get rid of the copy here thanks to filter. I can change that before merging. Don't want to bother you with nitpicks ;)

minad: #'cdr

consult.el
minad: Only stylistic and for consistency. I think seq can be treated as more fundamental since seq is somehow Emacs stdlib. I am personally not a fan of overusing the cl lib. I am using only a few things like cl-defun for kw args. You might have noticed that I replaced all uses of cl-loop with other iteration constructs. I like your loopy more than cl. But even more preferable to such dsls is composing functions. I am preferring the functional style maybe combined with a bit of mutation. But I am also a haskeller. In cases where composing multiple functions is not good enough for performance I am ok with imperative style dolist or cl-loop if it looks nicer.

minad:
merged

PR #23, opened 2020-12-01
Create ‘consult-kmacro’.

okamsn:
This is basically the same as the `selectrum-kmacro` version.

Consistent with `consult-register`, information about the keyboard macro is part of the candidate instead of being prefixed via the `selectrum-candidate-display-prefix` property. This makes history less useful for candidates containing counters, as the recorded candidate will no longer be correct after running the macro again.

Do you have a preferred method for dealing with such issues?

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -997,19 +997,20 @@ ones made entirely of mouse clicks) are not shown."
         ;;
         ;; Get actual index, since we prepended ‘kmacro-ring’
         ;; with ‘last-kbd-macro’ in selection.
-        (let ((actual-index (1- chosen-kmacro-index)))
-          ;; Temporarily change the variables to retrieve the correct
-          ;; settings.  Mainly, we want the macro counter to persist, which
-          ;; automatically happens when cycling the ring.
-          (cl-destructuring-bind
-              (last-kbd-macro kmacro-counter kmacro-counter-format-start)
-              (nth actual-index kmacro-ring)
-            (kmacro-call-macro (or arg 1) t)
-            ;; Once done, put updated variables back into the ring.
-            (setf (nth actual-index kmacro-ring)
-                  (list last-kbd-macro
-                        kmacro-counter
-                        kmacro-counter-format))))))))
+        (let* ((actual-index (1- chosen-kmacro-index))
+               (actual-kmacro (nth actual-index kmacro-ring))
+               ;; Temporarily change the variables to retrieve the correct
+               ;; settings.  Mainly, we want the macro counter to persist, which
+               ;; automatically happens when cycling the ring.
+               (last-kbd-macro (car actual-kmacro))
+               (kmacro-counter (cadr actual-kmacro))
+               (kmacro-counter-format (caddr actual-kmacro)))
+          (kmacro-call-macro (or arg 1) t)
+          ;; Once done, put updated variables back into the ring.
+          (setf (nth actual-index kmacro-ring)
+                (list last-kbd-macro
+                      kmacro-counter
+                      kmacro-counter-format)))))))
 
 ;;;; consult-preview-mode - Enabling preview for consult commands
 
--- a/consult.el
+++ b/consult.el
@@ -950,40 +950,36 @@ ones made entirely of mouse clicks) are not shown."
   (interactive "p")
   (if (or last-kbd-macro kmacro-ring)
       (let* ((numbered-kmacros
-              (cl-loop
-               ;; The most recent macro is not on the ring, so it must be
-               ;;  explicitly included.
-               for kmacro in (cons (if (listp last-kbd-macro)
-                                       last-kbd-macro
-                                     (list last-kbd-macro
-                                           kmacro-counter
-                                           kmacro-counter-format))
-                                   kmacro-ring)
-               ;; 1. Format the macros.  Mouse clicks are removed
-               ;;    by the format function.
-               ;;
-               ;; 2. Give the macros with non-empty strings a number, which
-               ;;    we use for rotating the ring.
-               for index = 0 then (1+ index)
-               for formatted-kmacro
-               = (condition-case nil
-                     (format-kbd-macro (if (listp kmacro)
-                                           (cl-first kmacro)
-                                         kmacro)
-                                       1)
-                   ;; Recover from error from ‘edmacro-fix-menu-commands’.
-                   ;; In Emacs 27, it looks like mouse events are
-                   ;; silently skipped over.
-                   (error
-                    "Warning: Cannot display macros containing mouse clicks"))
-
-               unless (string-empty-p formatted-kmacro)
-               collect (cons (concat (when (consp kmacro)
-                                       (format "%d,%s: "
-                                               (cl-second kmacro)
-                                               (cl-third kmacro)))
-                                     formatted-kmacro)
-                             index)))
+              (delete
+               nil
+               (seq-map-indexed
+                (lambda (kmacro index)
+                  (let ((formatted-kmacro
+                         (condition-case nil
+                             (format-kbd-macro (if (listp kmacro)
+                                                   (cl-first kmacro)
+                                                 kmacro)
+                                               1)
+                           ;; Recover from error from
+                           ;; ‘edmacro-fix-menu-commands’.  In Emacs 27, it
+                           ;; looks like mouse events are silently skipped over.
+                           (error
+                            "Warning: Cannot display macros containing mouse clicks"))))
+                    (unless (string-empty-p formatted-kmacro)
+                      (cons (concat (when (consp kmacro)
+                                      (format "%d,%s: "
+                                              (cl-second kmacro)
+                                              (cl-third kmacro)))
+                                    formatted-kmacro)
+                            index))))
+                ;; The most recent macro is not on the ring, so it must be
+                ;;  explicitly included.
+                (cons (if (listp last-kbd-macro)
+                          last-kbd-macro
+                        (list last-kbd-macro
+                              kmacro-counter
+                              kmacro-counter-format))
+                      kmacro-ring))))
              ;; The index corresponding to the chosen kmacro.
              (chosen-kmacro-index
               (consult--read "Keyboard macro: "
--- a/consult.el
+++ b/consult.el
@@ -185,6 +185,9 @@ nil shows all `custom-available-themes'."
 (defvar consult-minor-mode-history nil
   "History for the command `consult-minor-mode'.")
 
+(defvar consult-kmacro-history nil
+  "History for the command `consult-kmacro.'")
+
 ;;;; Internal variables
 
 (defvar consult--gc-threshold 67108864
@@ -217,6 +220,11 @@ nil shows all `custom-available-themes'."
 (defvar package-archive-contents)
 (declare-function package-desc-summary "package")
 (declare-function package--from-builtin "package")
+
+(defvar kmacro-ring)
+(defvar kmacro-counter)
+(defvar kmacro-counter-format)
+(defvar kmacro-counter-format-start)
 
 ;;;; Helper functions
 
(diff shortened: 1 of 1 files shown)


minad:
Including the prefix should be avoided generally but is ok for registers since they are named.

minad:
What you write regarding histories makes sense. I am using a prefix hack for disambiguation in consult-line. This should be kept in mind. Either remove the history there or find a better disambiguation solution. Some of the other histories are also not useful I guess and should be removed. For example consult mark history. You can also check!

consult.el
```diff
@@ -218,6 +221,11 @@ nil shows all `custom-available-themes'."
 (declare-function package-desc-summary "package")
 (declare-function package--from-builtin "package")
 
+(defvar kmacro-ring)
+(defvar kmacro-counter)
+(defvar kmacro-counter-format)
+(defvar kmacro-counter-format-start)
+
```
minad, line 228: Require kmacro

consult.el
```diff
@@ -932,6 +940,81 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
   (interactive)
   (consult--buffer #'switch-to-buffer #'find-file #'bookmark-jump))
 
+;;;###autoload
+(defun consult-kmacro (arg)
+  "Run a chosen keyboard macro.  With prefix ARG, run the macro that many times.
+
+Macros containing mouse clicks can't be displayed properly.  To
+keep things simple, macros with an empty display string (e.g.,
```
minad, line 948: Shorter description please. Remove clicks since they are not displayed properly or sth?

consult.el
```diff
@@ -932,6 +940,81 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
...
+ones made entirely of mouse clicks) are not shown."
+  (interactive "p")
+  (if (or last-kbd-macro kmacro-ring)
+      (let* ((numbered-kmacros
+              (cl-loop
```
minad, line 953: As argued before I prefer no cl-loop here

consult.el
```diff
@@ -932,6 +940,81 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
...
+            (kmacro-call-macro (or arg 1) t nil)
+          ;; Otherwise, run a kmacro from the ring.
+          ;;
+          ;; Get actual index, since we prepended ‘kmacro-ring’
+          ;; with ‘last-kbd-macro’ in selection.
+          (let ((actual-index (1- chosen-kmacro-index)))
+            ;; Temporarily change the variables to retrieve the correct
+            ;; settings.  Mainly, we want the macro counter to persist, which
+            ;; automatically happens when cycling the ring.
+            (cl-destructuring-bind
```
minad, line 1007: I prefer pcase or manual accessing if not too complicated

consult.el
```diff
@@ -932,6 +940,81 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
...
+               for formatted-kmacro
+               = (condition-case nil
+                     (format-kbd-macro (if (listp kmacro)
+                                           (cl-first kmacro)
+                                         kmacro)
+                                       1)
+                   ;; Recover from error from ‘edmacro-fix-menu-commands’.
+                   ;; In Emacs 27, it looks like mouse events are
+                   ;; silently skipped over.
+                   (error
```
minad, line 977: Can we just ignore such errors?

consult.el
```diff
@@ -932,6 +940,81 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
...
+                    "Warning: Cannot display macros containing mouse clicks"))
+
+               unless (string-empty-p formatted-kmacro)
+               collect (cons (concat (when (consp kmacro)
+                                       (format "%d,%s: "
+                                               (cl-second kmacro)
+                                               (cl-third kmacro)))
```
minad, line 984: The counter should not be part of the string. Use a space as suffix/prefix with display property to annotate.

consult.el
```diff
@@ -932,6 +940,81 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
...
+Macros containing mouse clicks can't be displayed properly.  To
+keep things simple, macros with an empty display string (e.g.,
+ones made entirely of mouse clicks) are not shown."
+  (interactive "p")
+  (if (or last-kbd-macro kmacro-ring)
+      (let* ((numbered-kmacros
+              (cl-loop
+               ;; The most recent macro is not on the ring, so it must be
+               ;;  explicitly included.
+               for kmacro in (cons (if (listp last-kbd-macro)
```
minad, line 956: Delete duplicates from the list?

consult.el
```diff
@@ -932,6 +940,81 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
...
+            (cl-destructuring-bind
+                (last-kbd-macro kmacro-counter kmacro-counter-format-start)
+                (nth actual-index kmacro-ring)
+              (kmacro-call-macro (or arg 1) t)
+              ;; Once done, put updated variables back into the ring.
+              (setf (nth actual-index kmacro-ring)
+                    (list last-kbd-macro
+                          kmacro-counter
+                          kmacro-counter-format))))))
+    (user-error "No keyboard macros defined")))
```
minad, line 1016: Move the error to the beginning without identing the rest. Early bailout.

consult.el
```diff
@@ -932,6 +940,81 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
...
+             ;; The index corresponding to the chosen kmacro.
+             (chosen-kmacro-index
+              (consult--read "Keyboard macro: "
+                             numbered-kmacros
+                             :require-match t
+                             :sort nil
+                             :history 'consult-kmacro-history
+                             :lookup (lambda (candidates x)
+                                       (cdr (assoc x candidates))))))
+        (if (= chosen-kmacro-index 0)
```
minad, line 996: Could yo move the logic into the :lookup function? Or is there a reason why not?

minad (changes requested):

consult.el
minad: I think it is good like that.

consult.el
okamsn: The function `edmacro-fix-menu-commands`, which is called by `format-kbd-maro`, will signal the error "Macros with mouse clicks are not supported by this command" in Emacs 26.3. The signalling of this error will prevent the command `consult-kmacro` from completing, so `consult-kmacro` should respond to it.

I chose to just include the warning in the candidate list, so that the problematic macros could still be run with `consult-kmacro`. Other options include removing those candidates from the list.

Later versions of Emacs just skip over the mouse clicks, and so don't signal the error.

I don't know how often people use macros with mouse clicks, so I am not sure of the best way to respond to the error.

minad: Ok. I am not sure I understand everything you write but I think we should do the following: filter out all macros with mouse clicks in the beginning. And then don't catch any errors. Then we are fine. Nobody uses mouse clicks with macros and if someone does we could still add support later if the issue comes up. But I don't want to focus on special cases now. We should exclude it.

consult.el
okamsn: I was unsure about removing duplicates, since they are distinct items in the ring. 

Users might want to take advantage of the position of macros in the ring, such as rotating the macro to make switching between macro take less effort. Additionally, because of counters, macros might only be duplicates temporarily.

minad: I am not sure what is better. Removing duplicates could be better since after narrowing in the minibuffer the order of the macros is somehow not relevant anymore, since the context is lost. Regarding the counter, you could exclude that from the duplicate check? After selecting the macro are you somehow rotating the ring? If that is the case I think it does not make sense to remove duplicates and you want to present the ring unsorted. But right now I think you do not, you just pick the selected macro and execute it?

okamsn: What I meant was more that adding a duplicate macro is an direct action, so the existence of duplicates can be purposeful/intentional.

However, I now think that your reasoning about special cases applies to this too. People will ask for it, if they want to keep duplicates.

okamsn:
I have tried to implement these changes. How is it now?

minad:
Good! I will take a closer look!

consult.el
minad: I think we can argue here that the interface at hand is simply not meant to work nicely with duplicates. It is just for selecting a macro. Therefore I guess we are good when we delete.

consult.el
```diff
@@ -932,6 +936,70 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
   (interactive)
   (consult--buffer #'switch-to-buffer #'find-file #'bookmark-jump))
 
+;;;###autoload
+(defun consult-kmacro (arg)
+  "Run a chosen keyboard macro.  With prefix ARG, run the macro that many times.
+
+Macros containing mouse clicks aren't displayed."
+  (interactive "p")
+  (if (not (or last-kbd-macro kmacro-ring))
```
minad, line 945: This should be a `unless` and then do not indent the let* since user-error throws. This is what I meant earlier. I prefer if the indentation level is not as deep.

consult.el
```diff
@@ -932,6 +936,70 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
...
+            (seq-uniq (let (result (index -1))
+                        (dolist ( kmacro (cons (if (listp last-kbd-macro)
+                                                   last-kbd-macro
+                                                 (list last-kbd-macro
+                                                       kmacro-counter
+                                                       kmacro-counter-format))
+                                               kmacro-ring)
+                                  (nreverse result))
+                          (cl-incf index)
+                          (unless (seq-some #'mouse-event-p (car kmacro))
```
minad, line 957: I would prefer seq-filter, followed by mapcar, followed by seq-uniq. Usually I juggle a bit around until it feels right. This is probably different from writing cl-loops. If you start functional and decompose everything, it stays pretty readable and is often not worse than cl-loop. I don't know how you feel about this?

minad (changes requested):
Almost ready, only minor changes required.

minad:
@okamsn I merged it and made rewrote it slightly. Please take a look if it works for you!

minad:
@okamsn Another question - since the kmacro-counter-format is %d by default, do you think it would make sense to hide the format if it is the default?

okamsn:
I don't mind seeing the default format. I find it to be a helpful reminder that I can change the format if I need to (which I forget, sometimes).

One thing I noticed is  that your version adds the indices at the end of the process, after filtering candidates. The indices need to be added before the filtering, otherwise they refer to the wrong position in `kmacro-ring`.

Also, thank you for teaching me about `thread-last`. It seems really useful.

minad:
> One thing I noticed is that your version adds the indices at the end of the process, after filtering candidates. The indices need to be added before the filtering, otherwise they refer to the wrong position in kmacro-ring.

Thank you for rechecking. I am right now playing around with it and I think I would prefer if the default 0(%d) is not shown, e.g., or macros which do not use the counter. For the other commands I tried to reduce the clutter as much as possible.

okamsn:
> I think I would prefer if the default 0(%d) is not shown, e.g., or macros which do not use the counter.

That makes sense to me. I think you can check if a kmacro is used with the following.

``` elisp
(when (or (seq-contains keys 'f3)
          (cl-search [?\C-x ?\C-k ?\C-i] keys))
      (propertize " "
                  'display
                  (format "%d(%s) " counter format)))
```

okamsn:
> I could also check if the counter is zero and the format is the kmacro-default-counter-format? Or why use this more complicated conditional?

The counter can start from any value, so a value of zero does not necessarily mean that the counter isn't being used.

A better approach, in my opinion, is to check whether the keys that run the command for inserting the counter are present in the macro. That way we know for sure that the counter is being used.

These are only the default keybindings for commands that insert the counter, but it is a start. I don't yet know how to find all of the keybindings for a command.

minad:
> A better approach, in my opinion, is to check whether the keys that run the command for inserting the counter are present in the macro. That way we know for sure that the counter is being used.

Ok!

> These are only the default keybindings for commands that insert the counter, but it is a start. I don't yet know how to find all of the keybindings for a command.

I am not sure if there is an easy way. There is where-is-internal but I feel it is slow and you must specify the keymap and you must be in the correct buffer if the keymap is buffer local. See what we do in marginalia https://github.com/minad/marginalia/blob/bcb6bbbde6cf3e2ebc66cfe1b0cd5b5d5585396c/marginalia.el#L129

PR #27, opened 2020-12-02
Add example for installing with Straight.el

ashton314:
You've only got one example of installation; since this is billed as working well with Selectrum, I think a lot of people (like myself) would be interested in installing this package with Straight.el.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,15 @@ order to use the enhanced commands, you must configure the keybindings yourself.
   )
 ~~~
 
+If you're using [Straight.el](https://github.com/raxod502/straight.el), you can get similiar behavior like so:
+
+~~~elisp
+(straight-use-package '(consult :type git :host github :repo "minad/consult"))
+(consult-annotate-mode)
+(setf (alist-get 'execute-extended-command consult-annotate-alist) #'consult-annotate-command-full)
+~~~
+
+
 ### Configuration settings
 
 | Variable                   | Def | Description                                             |
--- a/README.md
+++ b/README.md
@@ -153,13 +153,15 @@ order to use the enhanced commands, you must configure the keybindings yourself.
   )
 ~~~
 
-If you're using [Straight.el](https://github.com/raxod502/straight.el), you can get similiar behavior like so:
+If you're using [Straight.el](https://github.com/raxod502/straight.el), install like so:
 
 ~~~elisp
 (straight-use-package '(consult :type git :host github :repo "minad/consult"))
 (consult-annotate-mode)
 (setf (alist-get 'execute-extended-command consult-annotate-alist) #'consult-annotate-command-full)
 ~~~
+
+You will need to add the key bindings yourself.
 
 
 ### Configuration settings


README.md
```diff
@@ -153,6 +153,17 @@ order to use the enhanced commands, you must configure the keybindings yourself.
...
 
+If you're using [Straight.el](https://github.com/raxod502/straight.el), install like so:
+
+~~~elisp
+(straight-use-package '(consult :type git :host github :repo "minad/consult"))
+(consult-annotate-mode)
+(setf (alist-get 'execute-extended-command consult-annotate-alist) #'consult-annotate-command-full)
+~~~
+
+You will need to add the key bindings yourself.
```
minad, line 164: Thank you! This is good addition for the readme.

Two things:
1. Can you please squash your commits such that it is only one commit in the end before merging? I prefer that for simple additions.
2. Can you rewrite the config a little bit, such a you mark somehow e.g. with an ellipsis where the use-package configuration from above should go? Such that we don't have to document things twice and such that the configurations do not differ? You added this setf as an example, I suppose? I've also seen there is something like ":straight t" for use-package. Could something like that be used? I am not a straight user myself, so I cannot tell what the best way is.

minad (changes requested):

README.md
ashton314: I personally don't use `use-package` any more; just pure `straight.el`. There *is* a way to use `straight.el` with `use-package`; I'll add that example.

minad: @ashton314 Another question since I am using package.el+use-package.el - are there more package managers to consider or is straight the next most important package manager after package.el+elpa/melpa? I know that doom (and maybe spacemacs?) use straight.el, so I probably has a large user base. But I would like to avoid adding instructions for all the different possible setups. It is certainly okay to document the two or three most used alternatives. Selectrum is a bit more bold in that regards, but it is also by the author of straight.el, so this is understandable :)

minad: > I personally don't use use-package any more; just pure straight.el. There is a way to use straight.el with use-package; I'll add that example.

Ok, could explain this a bit better. I see use-package for configuration and package managers as orthogonal concerns. What I've read there seems to be mostly agreement regarding this. And since use-package is maintained by John Wiegley, will be upstreamed to Emacs proper soon, works great for me (one of the best packages behind the scenes for me to get some structure into my init.el). Therefore I would prefer if the configuration here also uses use-package. Is there something else which straight offers instead?

ashton314: That's a good question. My evolution through emacs packages has been built-in package.el -> use-package -> straight.el; I might not be the best person to ask what the most popular packages are.

> Melpa
Is this on Melpa now? Straight wasn't able to find the package on Melpa last time I tried. If that's the case, I'll revise the instructions to not worry about the git repo.

> Selectrum is a bit more bold in that regardsa

😃 That's for sure.

minad: > That's a good question. My evolution through emacs packages has been built-in package.el -> use-package -> straight.el; I might not be the best person to ask what the most popular packages are.

Ok, but I guess straight is good to have here! But since you don't use use-package anymore I would like to hear why!

> Is this on Melpa now? Straight wasn't able to find the package on Melpa last time I tried. If that's the case, I'll revise the instructions to not worry about the git repo.

No, the Melpa pipeline is somehow clogged unfortunately. I just asked on reddit https://www.reddit.com/r/emacs/comments/k5ebpj/is_there_any_chance_to_add_more_maintainers_to/

minad: @ashton314 It seems raxod502 also uses use-package, so we should use that too in any case here. See https://github.com/raxod502/radian/blob/06f2a30258deacb719a195221747ce1d528fc2fe/emacs/radian.el#L540

minad:
It seems to me one only needs ":straight t" in the use-package specification. I guess there is no need to document that here.

Luis-Henriquez-Perez:
> It seems to me one only needs ":straight t" in the use-package specification.

I think that's if it's in melpa. I had to specify the recipe to install this package since straight didn't know about it.

> Since you don't use use-package anymore, I would like to hear why.

I couldn't help but interject. I know you didn't ask me, but I'd like to share why I stopped using use-package as well.

1. It's a not lisp
use-package creates a DSL specifically designed for installing packages. As a consequence, things like indentation and navigation through the code via lisp editing constructs. For some packages, for example, I want to autoload more than a handful of functions. Since use-package is a macro emacs doesn't realize how to properly indent the list of commands after the commands keyword. It's harder to navigate the use-package construct via things like lispy and to operate on them because they are now a declarative `cl-loop` type language.

2. Ambiguity as to what goes into a `use-package` form
I *strongly* prefer to program with small blocks of code as opposed to dealing with forms that are well over a page long. With packages where I have a lot of configuration such as `evil`, there's no way I can put all my configuration in a use-package form and keep it small. This creates this annoying dilemma for me as to what I should keep in and keep out. It's arbitrary.

3. Duality of Code Function
With the introduction of use-package there's now always two ways of doing the same thing. Do you do `(use-package some-package :config (+ 1 1))` or `(with-eval-after-load 'some-package (+ 1 1))`? If you have both then you're using two different languages. It becomes ambiguous of why use one over the other. And someone reading your code needs to understand the mechanism of use-package well to know what is happening and in what order.

4. Code saved via use-package is greatly exaggerated
Just using plain old looping and other lisp constructs, you can initialize most packages in less than five lines. It really isn't a big deal. In fact sometimes use-package forms are more lengthy.

```elisp
(setq gcmh-idle-delay 5)
(setq gcmh-verbose void-debug-p)
(setq gcmh-high-cons-threshold VOID-GC-CONS-THRESHOLD)
(setq gcmh-low-cons-threshold VOID-GC-CONS-THRESHOLD-MIN)

(autoload #'gcmh-mode "gcmh" nil t nil)
(void-add-hook 'emacs-startup-hook #'gcmh-mode) ;; My own custom hook function
```
```elisp
(use-package gcmh
  :hook gcmh-mode
  :init
  (setq gcmh-idle-delay 5)
  (setq gcmh-verbose void-debug-p)
  (setq gcmh-high-cons-threshold VOID-GC-CONS-THRESHOLD)
  (setq gcmh-low-cons-threshold VOID-GC-CONS-THRESHOLD-MIN))
```
Lesson:
Design is hard and tricky. Macros are insanely powerful, but it's hard to make an abstraction that is genuinely useful and elegant. Just seems more convenient on the surface doesn't mean it's good or better.  Be particularly wary of anything that tries to add a DSL on lisp (cl-loop, I'm looking at you).

minad:
@Luis-Henriquez-Perez Thank you for the long explanation. It helps for my understanding! But let me make it clear directly in the beginning that my decision on this point still holds due to the reasons stated above. While use-package criticism has its place it is not this readme of a package which does not have much to do with package management. The readme is simply stating the simplest way to configure the package according to its authors.

> 1. It's a not lisp ... As a consequence, things like indentation and navigation through the code via lisp editing constructs.

From what I've seen, Elisp does not have big problems with that. There is the `declare indent` construct to fix many macros. Macros are all used over the place. To me, macros are one of the identifying criteria of what makes a programming language a lisp. But I am generally wary of DSLs too and like you not a fan of cl-loop but for different reasons.

> 2. Ambiguity as to what goes into a use-package form

Agree, something like this is a problem. I rather had it with builtin packages which are not really meant to be configured with use-package (still many people do). But I have no problem moving everything inside a use-package block. It gives my init.el much more structure. It is somehow pressing me inside a structure but I prefer that over Emacs bankruptcy.

> 3. Duality of Code Function ... With the introduction of use-package there's now always two ways of doing the same thing.

In elisp there is a plethora of ways of doing the same thing. Due to the late binding nature of elisp the design space for dynamic extensions for example is enormous. You can do advice-add in many variants, you can use hooks, you can overwrite functions, you can eval, etc etc. From a static analysis point of view, this is horrible. Astonishingly it still works :)
But having two different syntactical language constructs - this is really a minor thing in contrast.

> And someone reading your code needs to understand the mechanism of use-package well to know what is happening and in what order.

Sure you have to understand use-package but it is okay and not more complicated than the existing auto-loading mechanism. I think the culprit here is the fundamentally complicated design of elisp regarding autoloading. The thin use-package shim on top adds not much to the complexity. As far as I can tell more people had problems with configuring autoloading properly without use-package than they had if they use it.

> 4. Code saved via use-package is greatly exaggerated

This is not why I use it. I use it because it forces me to keep things together, because it ensures that autoloading is correct, because it offers a neat keybinding dsl etc.

ashton314:
@minad I think my avoidance of use-package is just an artifact of the time when I was trying out `straight.el`; I think `use-pacakge`'s support for and documentation regarding the `straight.el` backend has improved, so I'll give it a try.

Sorry I kinda dropped off the planet for a few days; I'm still in school and finals start next week. Crunch time!

Thanks for making these packages; I hope the Mepla pipeline gets unstuck soon. Maybe consider adding a line to the README about installing directly from GitHub as a temporary solution?

minad:
> Thanks for making these packages; I hope the Mepla pipeline gets unstuck soon. Maybe consider adding a line to the README about installing directly from GitHub as a temporary solution?

Some my packages are under review now for melpa. It is not going quickly, but we will have the packages in maybe a week or two, I think. No need to add something temporary to the readme.

a13:
> It's a not lisp

why?

> specifically designed for installing packages

not true, it doesn't have anything to do with installation specifically, it's a configuration DSL, which is great if you want to configure Emacs (I'd not write packages using it)


> Since use-package is a macro emacs doesn't realize how to properly indent the list of commands after the commands keyword.

you can configure that

> It's harder to navigate the use-package construct via things like lispy

there are other ways to edit things


> With packages where I have a lot of configuration such as evil

one can always strip out things and make a new package, like evil-collection does

> It becomes ambiguous of why use one over the other.

that's easy, if there's a keyword for that - use it! (if there's no - extend use-package with a keyword for that)

> And someone reading your code needs to understand the mechanism of use-package well to know what is happening and in what order.

`pp-macroexpand-expression` is the answer

> Code saved via use-package

it's not about code saving at all (though in lot of cases use-package makes things shorter), but more about declarativity, to make things more error-proof and nicely organized


in your example it's better to use `:custom` since defcustoms can have setter fns 

```
(use-package gcmh
  :hook gcmh-mode
  :custom
  (gcmh-idle-delay 5)
  (gcmh-verbose void-debug-p)
  (gcmh-high-cons-threshold VOID-GC-CONS-THRESHOLD)
  (gcmh-low-cons-threshold VOID-GC-CONS-THRESHOLD-MIN))
```

>;; My own custom hook function

yep, that's how you start to invent your own abstractions (which you have to support yourself) instead of using a popular and working (mostly) solution


> Design is hard and tricky. Macros are insanely powerful, but it's hard to make an abstraction that is genuinely useful and elegant.

use-package is not perfect at all, but are there any better alternatives? I don't think so, `leaf` is just a use-package clone filled with myriads of keywords. Make your own set of macros/functions? No way, I was there many years ago and don't want to get back to all these NIH stuff.

As a software engineer in most cases I prefer to use standard solutions instead of NIH wheels. Use-package works for me (with minor quirks), and the rest doesn't matter

a13:
> builtin packages which are not really meant to be configured with use-package 

But why? For use-package there's no difference, whether it's a built-in, if you installed it manually or via package/quelpa/straight/whatever. It can be required, it uses the same core elisp functions, etc

minad:
@a13 I agree with you as you see above.

>  But why? For use-package there's no difference, whether it's a built-in, if you installed it manually or via package/quelpa/straight/whatever. It can be required, it uses the same core elisp functions, etc

You are right, you can configure builtins using use-package. I also do that. But I sometimes had problems with builtin packages, since I had to set `:ensure nil` in order to prevent looking up the package on elpa/melpa. This is where I missed "better support" for builtins.

PR #33, opened 2020-12-04
Preliminary code for unfolding org headline branch.

Luis-Henriquez-Perez:


minad:
If something like this could be done for the general outline mode I would like that very much. I mean this folding/unfolding is not org-mode specific.

Luis-Henriquez-Perez:
I introduced the function `consult--outline-show-branch-maybe`. Which is like the function I had before but generalized to outlines and with some improvements. It needs to be called after the going to the point. I'd like some direction on how exactly you want this done. I took a very simple route and just added a call to this function after `consult-outline` and `consult-line`. We could do this via advice, and allow the user to unadvise if they don't want this. I don't know, up to you.

I need to make some modifications for the case with `counsel-line` so that it ends up at the specific point you searched for. Right now if you use it will end up at the unfolded headline.

minad:
@Luis-Henriquez-Perez I wonder how general this is. There are many libraries which offer code folding, origami, outshine, vimish-fold etc. Do they all use the same outline mechanism internally? Furthermore we would need something which opens and closes all the headings after moving to another position during preview - Save current outline state, open, restore.

Luis-Henriquez-Perez:
> There are many libraries which offer code folding, origami, outshine, vimish-fold etc. Do they all use the same outline mechanism internally?

Outshine folds work with outline commands. I'm not certain about outshine and vimish fold.

> Furthermore we would need something which opens and closes all the headings after moving to another position during preview - Save current outline state, open, restore.

Ideally, the unfolding shouldn't happen during preview. The preview should stay the same. It should happen after the user has selected where they want to go.

minad:
> Outshine folds work with outline commands. I'm not certain about outshine and vimish fold.

I resolved this now. I am using the same mechanism as isearch uses. This should be compatible with everything or at least it is not the fault of consult, if something is not compatible.

> Ideally, the unfolding shouldn't happen during preview. The preview should stay the same. It should happen after the user has selected where they want to go.

With the current implementation the unfolding happens also during preview and is reset afterwards. I like it this way. But we can also make the behavior configurable if you prefer to not have unfolding during preview. PR welcome in that case I guess :)

PR #34, opened 2020-12-04
Add command for counsel-org-tag.

Luis-Henriquez-Perez:


minad:
First, the function must respect the consult naming scheme. Is it possible to upstream this to org btw? It seems to be misplaced here and a perfect fit for org itself?

minad:
Can you rebase this on the *main* branch and reopen as a new PR. The problem is that I changed the main branch from master.

PR #36, opened 2020-12-05
Only add counter information when kmacro would insert counter.

okamsn:
This uses `where-is-internal` to find the keybindings of `kmacro-insert-counter` and `kmacro-start-macro-or-insert-counter`.

In testing it, it only adds the counter information when that macro would insert the counter. For example, if I locally bind `kmacro-insert-counter` to `C-c i`, then `consult-kmacro` will treat that key sequence as inserting the counter in that particular buffer, but not others. It also works similarly for Evil bindings.

Does this work for you?

EDIT:

It seems to only check against the current Evil state, so it might be not be correct if the kmacro changes the Evil state. But, I think it would be odd to define a macro in one state and then try to use it a different Evil state. I think it is more likely that Evil users would just use Evil macros.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -935,9 +935,16 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
      (mapcar (pcase-lambda (`((,keys ,counter ,format) . ,index))
                (cons
                 (concat
-                 (propertize " "
-                             'display
-                             (format "%d(%s) " counter format))
+                 ;; If the counter is 0 and the counter format is its default,
+                 ;; then there is a good chance that the counter isn't actually
+                 ;; being used.  This can only be wrong when a user
+                 ;; intentionally starts the counter with a negative value and
+                 ;; then increments it to 0.
+                 (unless (and (eql counter 0)
+                              (string= format "%d"))
+                   (propertize " "
+                               'display
+                               (format "%d(%s) " counter format)))
                  (format-kbd-macro keys 1))
                 index))))
    ;; Remove duplicates


minad:
Can you please rebase on top of master?

Regarding the code - I am not sure if this is robust. where-is-internal returns a list of vectors. Is the kmacro in the kmacro-ring always stored as a vector? The problem is that Emacs internally uses two different representation for keypresses, strings and vectors. So I am not sure if this will always work. You could convert between vectors and strings in order to ensure that it will. Furthermore I am unsure what you do regarding this cl-search which searches for a subsequence, I don't understand why this is correct.

okamsn:
> Can you please rebase on top of master?

I have tried to do that. Does the result work?

> Regarding the code - I am not sure if this is robust. where-is-internal returns a list of vectors. Is the kmacro in the kmacro-ring always stored as a vector? The problem is that Emacs internally uses two different representation for keypresses, strings and vectors. So I am not sure if this will always work. You could convert between vectors and strings in order to ensure that it will, if this is necessary - is it? 

No, that is not needed. `cl-search` works on sequences, of which strings and vectors are two kinds. You can try this by running `(cl-search [?t ?h ?e] "Hello there")`, which should return `6`.

> Furthermore I am unsure what you do regarding this cl-search which searches for a subsequence, I don't understand why this is correct.

The goal is to check whether any of the keybindings that would insert the macro counter are present in the key presses that make up the keyboard macro. By default, those bindings are `[f3]` and `[?\C-x ?\C-k ?\C-i]`, but this approach also works well with the custom bindings that I have tried.

Does this make sense?

minad:
Regarding rebase - can you also squash? Such that it is only a single git commit on top of master.

Why do you search for a subsequence? It would be possible for the insert key to be a suffix under some prefix key sequence which would not lead to an insertion or am I missing something? Then this would not work under all circumstances.

okamsn:
> Why do you search for a subsequence? It could be possible for the insert key to be a suffix under some prefix key sequence which would not lead to an insertion or am I missing something? Then this would not work under all circumstances.

You are correct. I hadn't thought about that case. I will think about this some more.

minad:
@okamsn I think it is pretty difficult to get this right actually. Or alternatively fall back to the simpler heuristic I thought about before (checking for 0 and default format). Could you explain again why the counter is not necessarily zero?

okamsn:
> Could you explain again why the counter is not necessarily zero?

Counter values can be set to any integer with `M-x kmacro-set-counter`, so it is possible to begin with a negative counter value and then increment it until it is 0.

minad:
> Counter values can be set to any integer with M-x kmacro-set-counter, so it is possible to begin with a negative counter value and then increment it, eventually setting it to 0.

Yes but why? If I am not using a counter in my macro, I won't do that. And since I am not sure if we can solve this in a general way I would simply hide the macros with default format and 0. This is a simple and easy to understand heuristic - or do you expect it to break all the time?

okamsn:
I can't accurately speak on how others use keyboard macros, as I only know how I use them.

> Yes but why? If I am not using a counter in my macro, I won't do that. 

I have infrequently started marco counters with a negative value and then incremented the counter through 0 into the positive values. I'm only acknowledging that if someone wants to use counters that way, they can. I don't know if others find value in being able to do that.

>  do you expect it to break all the time?

I do not, because I don't know how common it is to use negative counter values. I also don't know how common it is that the bindings of `kmacro-insert-counter` or `kmacro-start-macro-or-insert-counter` are sub-sequences of other valid key sequences in a particular buffer. I am guessing that the latter is less common than the former, but you are correct, I do not know that for certain.

minad:
> I have infrequently started marco counters with a negative value and then incremented the counter through 0 into the positive values. I'm only acknowledging that if someone wants to use counters that way, they can. I don't know if others find value in being able to do that.

Ok, I understand now better what you want to say. Sure, in some cases the counter will be hidden if it hits zero for such a macro. But given the simple heuristic - you always know during selection that if a macro has no visible counter prefix, the counter must be zero and the format must be the default format. Therefore I am not worried about this problem. I would prefer that over adding a more complicated mechanism.

We could also combine both heuristics - checking for the key and checking for the counter. But usually I go with the simplest solution, your version is also acceptably simple, but breaks for weird keysequences, which are probably not used by anybody. My solution "breaks" for negative counters. But then I argue that it is not really breaking but simply hiding the prefix if it is the default value. You are more aiming for a solution which detects if a counter is used by a macro.

What do you prefer? It depends a bit on the perspective of you how you look at it.

okamsn:
I think that until this "key detection" approach can work more predictably, it is better to do what you have suggested and check for default values.

minad:
Agree! Thank you!

PR #37, opened 2020-12-05
Assign evil's :jump property to relevant commands

leungbk:
----

#

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -201,8 +201,9 @@ nil shows all `custom-available-themes'."
 ;;;; Pre-declarations for external packages
 
 (defvar selectrum-should-sort-p)
-(declare-function selectrum-read "selectrum")
-(declare-function selectrum-get-current-candidate "selectrum")
+(declare-function selectrum-read "ext:selectrum")
+(declare-function selectrum-get-current-candidate "ext:selectrum")
+(declare-function evil-set-command-property "ext:evil-common")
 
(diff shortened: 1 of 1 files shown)


minad:
What is this evil jump property, I may ask? I will probably not merge something like this since it unnecessarily pollutes the code and I won't be able to maintain these properties. My suggestion would be to add a configuration example to the [wiki](https://github.com/minad/consult/wiki).

leungbk:
The `:jump` property ensures that the current value of `point` is placed on a jump list used by `evil`'s `C-i` and `C-o` commands.

It's pretty easy to maintain. `evil` is quite popular, boasting more than one million downloads on MELPA, so many `consult` users will benefit without having to be aware of the wiki, while `consult` users who don't use `evil` are unaffected. @minad @okamsn what do you think?

minad:
@leungbk Thank you for your answer. From my point of view it is wrong to define these within unrelated package. These properties should go into your init.el. Since you already have to bind each consult command to some key, it does not hurt too much to also add some evil properties on top of that. Consult is not meant to provide a "works out of the box experience". It is rather one component of a set of packages which are meant to work together: icomplete/selectrum+consult+embark+marginalia+orderless/prescient for example. If you want all those to work nicely with evil you will have to do some configuration work yourself - or you might want to write a package/init-example which combines those and preconfigures everything.

PR #38, opened 2020-12-06
Add function to open file in operating system's default application

oantolin:
Here's the function I mentioned in the wishlist issue #6.


oantolin:
I've used this function on Windows (windows-nt, not cygwin) and Linux for a couple of years, but I have never tested it on Mac OS or on Cygwin. Counsel uses the same commands on those systems, so I'm pretty sure it is OK, but thought I should mention this for full disclosure.

PR #43, opened 2020-12-08
Have the consult-yank* commands report the category 'kill-ring

oantolin:
I'd find it useful for the yank commands to some special category so I can configure embark to display them vertically instead of in a grid view. But even without that selfish motivation, it's probably a good idea to have them report some category.

I choose the name `kill-ring` for the category, but maybe you have a better suggestion.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -654,6 +654,7 @@ The arguments and expected return value are as specified for
   "Open kill ring menu and return selected text."
   (consult--read "Ring: "
                  (delete-dups (seq-copy kill-ring))
+                 :category 'kill-ring
                  :require-match t))
 
 ;; Insert selected text.


minad:
I wonder about one thing - the common criticism regarding the yank commands using the minibuffer is that the minibuffer is not well suited for the multiline yanks. Is this resolved by switching to the Embark occur buffer, such that you get multilines there? If yes this is great and would also make the browse-kill-ring obsolete!

oantolin:
Yes, the embark occur buffer does show multiline candidates in full. The problem is that there is no separation between them. That is why in my personal version of consult-line I added a horizontal line at the end of each candidate (just as a display property, not actually included in the candidate). Would you be interested in that?

This is my old function, it's supposed to be like `consult-yank-replace`:


```elisp
(defun completing-yank ()
 "Insert item from kill-ring, selected with completion."
 (interactive)
 (when (eq last-command 'yank)
   (let ((inhibit-read-only t))
     (delete-region (point) (mark t))))
 (insert-for-yank
  (let* ((sep (propertize
               (concat "\n" (make-string (1- (window-width)) ?—))
               'face 'shadow))
         (embark-occur-initial-view-alist '((t . list))))
    (completing-read "Yank: "
     (mapcar (lambda (str) (propertize str 'display (concat str sep)))
             kill-ring)
     nil t))))
```

I could add the separators to `consult--yank-read`.

minad:
Awesome! I would rather like to do this differently - via candidate transformers :D

oantolin:
I thought candidate transformers were for the opposite direction. Like, in consult-line the candidates need to include the line number, and then a transformer can strip it before someone consumes it.

Here, the candidates can be just the kill-ring entries, just propertized to display as if they had a horizontal line at the end. There is no need to transform those propertized strings at all.

oantolin:
I made a separate pull request for the horizontal separators #44. See if you like them.

PR #44, opened 2020-12-08
Add a horizontal line below each consult--yank-read candidate

oantolin:
- The candidates do not include the horizontal line, it is only
displayed via text properties.

- The line comes at the end of the candidate, so I believe this should
not affect what selectrum displays at all. (This should be tested.)
So, this doesn't improve things for selectrum, but it should keep them
the same.

- Both icomplete-vertical and embark-occur display the full multiline
candidate. So consult + either can be regarded as a substitute for
browse-kill-ring. (We mentioned embark-occur already, I forgot that
icomplete-vertical shows multiline candidates too.)

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -653,7 +653,12 @@ The arguments and expected return value are as specified for
 (defun consult--yank-read ()
   "Open kill ring menu and return selected text."
   (consult--read "Ring: "
-                 (delete-dups (seq-copy kill-ring))
+                 (let* ((sep (propertize
+                              (concat "\n" (make-string (1- (window-width)) ?—))
+                              'face 'shadow)))
+                   (mapcar (lambda (str)
+                             (propertize str 'display (concat str sep)))
+                           (delete-dups kill-ring)))
                  :require-match t))
 
 ;; Insert selected text.


oantolin:
Here the code to add separators between kill ring entries. Try it and see if you like it. Unless I'm mistaken it should have no effect at all under selectrum, but it should show the full kill ring entries with separators under icomplete-vertical or embark-occur. And it should make no difference in terms of what gets inserted in the buffer.

minad:
Generally I am not eager to add something like this since it seems to presentational to me and does not work at all with the single-line per candidate display, to which I am accustomed as of now. I didn't know Icomplete-vertical supports multi-line candidates.

Therefore I mentioned again the half-baked idea of candidate transformers. The simplest/ungeneral option of simply adding properties to the candidates still seems viable to me. For consult-line have a property added to the candidate prefix. This candidate prefix gets then stripped off before showing the candidate somewhere else (if this is what we wanted to achieve). Similarily for the kill-ring candidates we could add some property to the candidate, marking it as a multiline candiate which shuld be presented with lines in between in the Embark occur buffer. Alternatively you could simply detect if there are newlines in the candidates and act accordingly?

Overall, I don't feel this PR is the right way to go. I am hoping for a better solution, but I know that I am pushing the problem somehow towards Embark, which is not very nice ;) 

> Both icomplete-vertical and embark-occur display the full multiline
candidate. So consult + either can be regarded as a substitute for
browse-kill-ring. (We mentioned embark-occur already, I forgot that
icomplete-vertical shows multiline candidates too.)

Interesting! I wonder if something like this is possible in Selectrum too. Maybe @clemera can comment on that. But it could as well be that many things in Selectrum are tied to the design criterion that candidates are single line. Furthermore I wonder how it works out in Icomplete-vertical if you have many lines per candidate. Then only one candidate and only the beginning will be visible?

oantolin:
You said about embark:

>  Alternatively you could simply detect if there are newlines in the candidates and act accordingly?

This is doable. I also thought of maybe giving even and odd numbered candidates different background colors (always, whether or not they contain newlines).

You asked about icomplete-vertical:

> Then only one candidate and only the beginning will be visible?

Yes.

minad:
Let me write this as a separate comment:

I should add that I am not putting a final veto here, it is just that I am not comfortable with the change. If it turns out that there is no better way to get something like this or puts too much of a burden on Embark we can still merge this or something similar. As of now my pov is mostly biased to get a good minibuffer display. Therefore I expect further discussions in the future which are related to getting a good display in the Embark view *and* getting a good display in the minibuffer. It is all not unexpected. I also thought about having two different kinds of candidate variants, e.g. also via a property, such that each candidate can specify an alternative multiline view. I think we should explore these ideas. But I haven't thought any of those through carefully. Maybe they won't work for some reasons.

minad:
> This is doable. I also thought of maybe giving even and odd numbered candidates different background colors (always, whether or not they contain newlines).

These are both nice ideas! Either having a line between the candidates or having a stripe pattern, it is a matter of taste though.

> Then only one candidate and only the beginning will be visible?
> Yes.

Maybe I like the selectrum approach better for this then. Would it also possible to configure this for icomplete-vertical or is this very hard? If I want multiline I can always switch to the Embark occur view! But this is again a matter of taste I am sure :)

oantolin:
> I should add that I am not putting a final veto here, it is just that I am not comfortable with the change.

No problem! I thought you might not like it so I said "Try it and see if you like it" from the beginning.

By the way, it's not just `icomplete-vertical` that displays multiline candidates in full, so does `icomplete`! And, of course, `icomplete-vertical` gets the feature for free, there is no special code in it for multiline candidates.

minad:
Regarding what I said earlier:

The ugliness of these property ideas is however that we are deeply going into the territory of making our own standards for exchanging data between the minibuffer and embark. If we can avoid that it would be great! It is questionable however if doing a simple generic solution like adding lines between the candidates looks always nice.

One other thing one can consider - I think it is okay if we do a generic solution and add some kind of enhancements via properties etc, which will degrade gracefully. So if you use Embark with all the nice enhancements everything will look perfect, but there is no necessity for the properties to be there. Even without them they view should look reasonable.

----
Your comments:

>  No problem! I thought you might not like it so I said "Try it and see if you like it" from the beginning.

Sorry, probably I read over that line.

> By the way, it's not just icomplete-vertical that displays multiline candidates in full, so does icomplete! And, of course, icomplete-vertical gets the feature for free, there is no special code in it for multiline candidates.

And would it be possible to change that? Probably only with deeper hacks and you probably wouldn't like it as far as I understand? It is not that I would like to have that at all cost, but it is nice to understand what is possible and what is not.

clemera:
I just tested this and this gets bad effects in helm and selectrum, I think it's hard to make such display enhancements work well for all use cases and frameworks. Something like `icomplete` (not vertical) shouldn't display multiline candidates at all if you ask me ;)   BTW shouldn't `consult-yank` disable sort? This is the first time I tested it and I was surprised that the candidates weren't in order.

oantolin:
> I just tested this and this gets bad effects in helm and selectrum

Well, I guess that kills it! I'll retract this, then.

To get the lines in icomplete-vertical, one can just advise the `consult-yank*` family to `let ((icomplete-vertical-separator ...)) ...)`.

I do want some visual indication in Embark occur buffers, but the more I think about the more I prefer the stripes idea: it doesn't waste vertical space.

About icomplete and multiline candidates: it would be easy to display them as just the first lines. I could put some code to do that in icomplete-vertical, and it would be something you could use even horizontally. I'll put it on the todo list.

clemera:
> About icomplete and multiline candidates: it would be easy to display them as just the first lines. I could put some code to do that in icomplete-vertical, and it would be something you could use even horizontally. I'll put it on the todo list.

I removed that line about icomplete but you were quicker to answer...In selectrum we display the matched line followed by a truncated version of the lines merged into a single line, maybe that would work for icomplete, too.

oantolin:
> I removed that line about icomplete but you were quicker to answer...In selectrum we display the matched line followed by a truncated version of the lines merged into a single line, maybe that would work for icomplete, too.

Oh, that's smarter than just showing the first line and an ellipsis, which was my plan.

minad:
@oantolin would ist be possible to do thin lines between the candidates instead of the stripes. Like underlines or border lines. Not taking full lines, not wasting space.

clemera:
Showing the first matched line within the multi line candidate would probably be enough and maybe some indication how many lines the candidates contains. Maybe one dot for every newline?

oantolin:
> @oantolin would it be possible to show thin lines between the candidates instead of the stripes. Like underlines or border lines. Not taking full lines, not wasting space.

Good question! I don't know how efficiently it can be done. One thing one could do is look for the last line and display it underlined.

PR #47, opened 2020-12-09
replace consult-minibuffer-history by consult-history

minad:
This is a first draft regarding adopting the history command by @oantolin https://github.com/oantolin/completing-history.
I replace the command `consult-minibuffer-history` by a more general variant.

For now it is a bit simpler than the original and I would like to discuss a few things:

* ~Why not use minibuffer-history directly, why is there minibuffer-history-variable etc?~
  **EDIT**: I understand this now. It makes sense to have the indirection to allow overwriting the history from outside.
* ~In contrast to the original, I do not enable-recursive-minibuffers for now. I have them generally on but I am not sure if I should instead turn them on always for consult--read. I think I have to play around with them being turned off in order to understand if it makes sense. I am so accustomed to having them turned on...
**EDIT**: I have problems with this, I am really not accustomed to non-recursive-minibuffers...~
**EDIT2**: Resolved this. It makes sense to have this setting.
* ~Would it make sense to add an optional argument to the command which takes a history variable such that users can build there own history commands if there is a need, such that the auto-detection is not used? However one has to distinguish rings from minibuffer history variables.~
**EDIT**: I added a history argument for now. This makes it trivial to implement `consult-minibuffer-history` again.
* ~It would also be an option to keep the `consult-minibuffer-history` command. I am not sure if it is better to have a dwim command or have multiple more specialised commands. Very often I like dwim commands. Maybe have both?~
**EDIT**: I don't think there is a need for `consult-minibuffer-history`. If there is, just use `consult-history` with a custom argument.
* ~Due to the design of consult, the command is bound nowhere, so this won't be an exact substitute for the command by @oantolin. But I think the command could be bound to a C-c key and if it is available it can be used if not it reports "History is empty". We could make the error more precise, showing "History not available, configure consult-history-rings" for example.~
**EDIT**: I implemented the more precise error.
* I contrast to the original, I do not modify inhibit-read-only, I don't see why this command should get a special treatment. Should it?
* In contrast to the original, I do not query the command-history. It was queried when repeating. Keep it? Note that there is still consult-command-history, which executes commands.

Ping @oantolin @clemera @manuel-uberti for opinions. It seems I have many more questions than answers here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,9 @@ information (e.g. `M-x`, `describe-face`, `describe-symbol`, `helpful-function`,
 ### Histories
 
   * `consult-command-history`: Select a command from the `command-history`.
-  * `consult-minibuffer-history`: Insert a string from the `minibuffer-history`.
+  * `consult-history`: Insert a string from the current buffer history.
+    This command can be invoked from the minibuffer. In that case the history
+    stored in the minibuffer-history-variable is used.
 
 ### Jumping and Search
 
@@ -127,7 +129,8 @@ order to use the enhanced commands, you must configure the keybindings yourself.
 ;; Example configuration for Consult
 (use-package consult
   ;; Replace bindings. Lazily loaded due to use-package.
-  :bind (("C-c o" . consult-outline)
+  :bind (("C-c h" . consult-history)
+         ("C-c o" . consult-outline)
          ("C-x b" . consult-buffer)
          ("C-x 4 b" . consult-buffer-other-window)
          ("C-x 5 b" . consult-buffer-other-frame)
@@ -167,14 +170,14 @@ order to use the enhanced commands, you must configure the keybindings yourself.
 
 | Variable                   | Def | Description                                             |
 |----------------------------|-----|---------------------------------------------------------|
-| consult-annotate-alist     | …   | Functions which should get a richer completion display. |
 | consult-line-numbers-widen | t   | Show absolute line numbers when narrowing is active.    |
+| consult-mode-histories     | …   | Mode-specific history variables                         |
 | consult-preview-buffer     | t   | Enable buffer preview during selection                  |
-| consult-preview-theme      | t   | Enable theme preview during selection                   |
-| consult-preview-yank       | t   | Enable yank preview during selection                   |
-| consult-preview-mark       | t   | Enable mark preview during selection                    |
 | consult-preview-line       | t   | Enable line preview during selection                    |
+| consult-preview-mark       | t   | Enable mark preview during selection                    |
 | consult-preview-outline    | t   | Enable outline preview during selection                 |
+| consult-preview-theme      | t   | Enable theme preview during selection                   |
+| consult-preview-yank       | t   | Enable yank preview during selection                    |
 | consult-themes             | nil | List of themes to be presented for selection            |
 
 ## Acknowledgements
--- a/consult.el
+++ b/consult.el
@@ -41,6 +41,7 @@
 (require 'kmacro)
 (require 'outline)
 (require 'recentf)
+(require 'ring)
 (require 'seq)
 (require 'subr-x)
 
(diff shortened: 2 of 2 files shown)


s-kostyaev:
Where can I read about `recursive-minibuffers` concept?

minad:
@s-kostyaev https://www.gnu.org/software/emacs/manual/html_node/elisp/Recursive-Mini.html https://www.gnu.org/software/emacs/manual/html_node/emacs/Recursive-Edit.html See also my https://github.com/minad/recursion-indicator. I find that indicator helpful to see the recursion level you are in.

minad:
I am more satisfied with the design now, maybe except for the names: `consult-history`, `consult-mode-history` etc.

oantolin:
I quickly read your code and it looks good to me! A couple of quick points:

- I'm pretty sure I had a good reason to add the `inhibit-read-only`,but I don't remember what that reason was. (I should have commented that!) I say we leave it as it is now, and usage will quickly point out why this was necessary (and what a better fix is). I remember the problem was with some completely normal every day mode, maybe shell-mode? Something like that.

- I think your `consult-history` command doesn't support `repeat-complex-command`, right? I guess that's what the separate `consult-command-history` is for. Maybe `consult-history` can support `repeat-complex-command` anyway, despite the existence of the other command?

minad:
@oantolin Thank you very much for looking at it and for making the original version!

> I'm pretty sure I had a good reason to add the inhibit-read-only,but I don't remember what that reason was. (I should have commented that!) I say we leave it as it is now, and usage will quickly point out why this was necessary (and what a better fix is). I remember the problem was with some completely normal every day mode, maybe shell-mode? Something like that.

I suspected something like this, but I am still a bit wary of using the sledgehammer ;)

> I think your consult-history command doesn't support repeat-complex-command, right? I guess that's what the separate consult-command-history is for. Maybe consult-history can support repeat-complex-command anyway, despite the existence of the other command?

Yes, it does not. I can add this back. Can you explain why this is useful? I am confused why you want the command history if the last command is the repeat command. But there are many thing which confuse me about repeating commands tbh, this-original-command, this-real-command, this-really-really-original-command, this-repeatable-command I am looking at you...

oantolin:
> Yes, it does not. I can add this back. Can you explain why this is useful? I am confused why you want the command history if the last command is the repeat command.

I had that special support for `repeat-complex-command` (`C-x M-:`) because I didn't write a command analogous to `consult-command-history`, instead I wanted to just support a completing-read history search for all commands I knew about where it would be useful. So instead of a dedicated command like `consult-command-history`, I would use `C-x M-:` and then my `completing-history-insert-item` command.

You'll notice the only extra thing needed is `prin1-to-string`, just like in your separate `consult-command-history`.

It would duplicate functionality, I guess, because you could either use `consult-command-history` or the `repeat-complex-command`/`consult-history` combo to get the same effect.

On the other hand, as it is now, it feels like `consult-history` is "broken" if used from `repeat-complex-command`.

> But there are many thing which confuse me about repeating commands tbh, this-original-command, this-real-command, this-really-really-original-command, this-repeatable-command I am looking at you...

:D I know, it confuses me too.

oantolin:
Oh, I realize now you were probably asking why I check for `repeat-complex-command` with `last-command`. That is actually slightly broken, it only works if you call the `completing-history-insert-item` right after calling `repeat-complex-command`. If you edit the minibuffer in between you don't get the special logic and the history you are shown only has the functions, not the entire s-expression.

It should really check if repeat-complex-command is the command responsible for opening the minibuffer, i.e., is it `marginalia--this-command`, but of course we don't have that here.

I see three options, since I know you don't like things that are broken by design (so my `last-command` hack is out):

1. Do nothing to the code and document that people wanting to use `consult-history` from within `repeat-complex-command` should instead use `consult-command-history`.

2. Detect `repeat-complex-command` more robustly: by looking at the prompt, for example.

minad:
Thanks for the explanation! Is it okay if we merge this as if since I think it is already an improvement over the status quo. And potentially add more logic to make the command more convenient as soon as someone of us has time for this/considers it important enough. I will add a comment linking to this discussion and document a bit more.

oantolin:
Sure, go ahead. And I'll update the Readme at completing-history (and my personal config, too!).

minad:
@oantolin I decided to adopt the behavior you had for repeat-complex-command. It is good enough.

PR #52, opened 2020-12-10
consult-info ported from ivy/counsel

minad:
@Luis-Henriquez-Perez I ported over your counsel/ivy PR as a test. How does it differ from the selectrum wiki version? And I wonder if the child node regexps cannot be improved somehow, they show a lot of entries.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -1170,6 +1170,59 @@ Macros containing mouse clicks aren't displayed."
                     kmacro-counter
                     kmacro-counter-format))))))
 
+(defun consult--info-top-nodes ()
+  "Return list of top-level info nodes to use for `consult-info'.
+Elements of the list are strings of Info file names without
+extensions (e.g. \"emacs\" for file \"emacs.info.gz\"). Info
+files are found by searching directories in `Info-directory-list'."
+  (info-initialize)
+  (let (files nodes)
+    (dolist (dir (or Info-directory-list Info-default-directory-list))
+      (when (file-directory-p dir)
+        (setq files (append files (directory-files dir nil "\\.info")))))
+    (save-match-data
+      (dolist (file files)
+        (string-match "\\(.+?\\)\\..+" file)
+        (push (match-string 1 file) nodes)))
+    (delete-dups (nreverse nodes))))
+
+(defvar consult--info-node-line-regexp
+  "\\* +\\([^\n]*.+[^\n]*\\):[\011 ]+\\([^\n]*\\)\\.\\(?:[\011\n ]*(line +\\([0-9]+\\))\\)?"
+  "Regexp used to find info-node lines.")
+
+(defun consult--info-child-nodes (top-node)
+  "Return a list of child nodes and their description as strings."
+  (save-selected-window
+    (with-temp-buffer
+      (info top-node (current-buffer))
+      (let (collection Info-history start end line)
+        (dolist (node (Info-index-nodes))
+          (Info-goto-node node)
+          (goto-char (point-min))
+          (while (search-forward "\n* " nil t)
+            (unless (search-forward "Menu:\n" (1+ (point-at-eol)) t)
+              (setq start (point-at-bol)
+                    end (or (save-excursion
+                              (goto-char (point-at-bol))
+                              (re-search-forward "(line +[0-9]+)" nil t))
+                            (point-at-eol))
+                    line (replace-regexp-in-string
+                          "\n" "" (buffer-substring start end)))
+              (push line collection))))
+        (nreverse collection)))))
+
+;;;###autoload
+(defun consult-info (top)
+  ;;"Prompt for node to jump to based on TOP-NODE."
+  "Search the child-nodes of a select top-level Info node.
+First prompt for the specific node to search in (ie. emacs, elisp,
+org-mode). Then search the child nodes (along with their description) of that
+node."
+  (interactive (list (consult--read "Top Node: " (consult--info-top-nodes))))
+  (let ((line (consult--read "Child Node:" (consult--info-child-nodes top))))
+    (string-match "\\*[[:space:]].+:[[:space:]]+\\(.+\\)\\." line)
+    (Info-goto-node (format "(%s)%s" top (match-string 1 line)))))
+
 ;;;; consult-preview-mode - Enabling preview for consult commands
 
 (defun consult--preview-update-selectrum ()


Luis-Henriquez-Perez:
I can't be sure of the differences or the reasons for them because I didn't write the selectrum wiki version. I think @clemera did but it doesn't say on the wiki.

However, looking at the selectrum wiki version and comparing it to my own I notice:

- it adds a custom variable: `selectrum-info-default-other-window` that controls where the info node is displayed. I don't know if this is necessary as the user can just customize `display-buffer-alist` to control how info nodes are opened. Anyway, the command shouldn't be dealing with window placement in the first place.

- `selectrum--info-get-child-node` preserves match data mine did not. This is probably an improvement.

- `selectrum--info-get-child-node` gets the list of nodes and prompts with `completing-read` whereas the my version just returned the list of nodes. I personally don't like this change, that function should just do one thing--return a list; it should not additionally deal with prompting the user via `completing-read`. Also, probably because it's doing too many things, it's really long.

- `selectrum--info-get-child-node` also deals with annotations and propertizing them which I did not do. If possible `marginalia` should deal with this. Annotations and propertizing seem like a aesthetic issue which this command itself should not be responsible for.

- It has some `selectrum-mode` specific code which is probably not a good idea to have if we want the command to be portable.

- it introduces `cl-loop`. I don't see how the introduction `cl-loop` improves the original in any way (in fact it makes it worse IMO), so we should probably do without it.

- It includes the variable `Info-additional-directory-list` which is a variable I didn't know about. So this seems like an improvement.

This is just an overview from me reading the code. To truly see the differences I need to run the commands. I will do this later today and report back (hopefully with screenshots) and determine specifically whether it has a better regexp that shows less entries.

minad:
@Luis-Henriquez-Perez Thank you! Would you like to work a bit on this PR? It is just a direct port of your cousel version. My problem with this command right now is mostly that I don't understand which kind of child nodes are shown. The candidate list looks very messy - is the regex too general? Is there nothing we can do to extract the sections somehow better?

Luis-Henriquez-Perez:
> Would you like to work a bit on this PR?
@minad Sure.

> My problem with this command right now is mostly that I don't understand which kind of child nodes are shown. The candidate list looks very messy - is the regex too general? Is there nothing we can do to extract the sections somehow better?

After I run the commands and compare them, I'll get back to you with an answer for these questions (although if I had to answer I'd say probably yes and yes).

minad:
Cool, looking forward to it!

okamsn:
> I can't be sure of the reasons for the differences because I didn't write the selectrum wiki version. I think @clemera did but it doesn't say on the wiki.

No, it was me.

> 
> However, looking at the selectrum wiki version and comparing it to my own I notice:
> 
> - it adds a custom variable: `selectrum-info-default-other-window` that controls where the info node is displayed. I don't know if this is necessary as the user can just customize `display-buffer-alist` to control how info nodes are opened. Anyway, the command shouldn't be dealing with window placement in the first place.
> 

Would you describe how to do that? I'm not really familiar with this feature.

> - `selectrum--info-get-child-node` preserves match data mine did not. This is probably an improvement.
> 
> - `selectrum--info-get-child-node` gets the list of nodes and prompts with `completing-read` whereas the my version just returned the list of nodes. I personally don't like this change, that function should just do one thing--return a list; it should not additionally deal with prompting the user via `completing-read`. Also, probably because it's doing too many things, it's really long.

It was my first approach to be able to just jump to the child node. I normally just use different shortcuts for different Info manuals.

> - `selectrum--info-get-child-node` also deals with annotations and propertizing them which I did not do. If possible `marginalia` should deal with this. Annotations and propertizing seem like a aesthetic issue which this command itself should not be responsible for.
> 
> - It has some `selectrum-mode` specific code which is probably not a good idea to have if we want the command to be portable.

The code allows for matching against node descriptions, which I find to be very useful. The Selectrum-specific code needs to be changed anyway, since I think user lemera is considering removing the `full-candidate` text property that the command currently uses.

> - it introduces `cl-loop`. I don't see how the introduction `cl-loop` improves the original in any way (in fact it makes it worse IMO), so we should probably do without it.

I was learning about `cl-loop`, so I made a few commands that use it on the Selectrum wiki. I like `cl-loop`, but I am learning that it is divisive.

minad:
@okamsn Thank you for looking at it too!

Luis-Henriquez-Perez:
@okamsn 

> Would you describe how to do that? I'm not really familiar with this feature.

Sure.

I highly recommend `display-buffer-alist`. Right now it seems like the best way
to control how windows are displayed--as in, whether they are opened in the same window, or a different window and how big the window should be, and where it should be etc.

When there is a buffer that you want to "tame", you add a corresponding entry in`display-buffer-alist'. For example, this is what I have in my config for org
source buffers. The car of the item you add is a regular expression which matches the buffer name. The CDR are elements that describes how you want the buffer displayed. I won't go more into it as the [manual does a better job than me](https://www.gnu.org/software/emacs/manual/html_node/elisp/Frame-Layouts-with-Side-Windows.html#Frame-Layouts-with-Side-Windows).

```elisp
(alet '("\\*Org Src"
	(display-buffer-at-bottom)
	(window-height . 0.5))
  (push it display-buffer-alist))
```

So for info buffers, if you wanted to display it in the side, you can do this.

```elisp
(alet '("\\*info\\*" display-buffer-in-side-window
	(side . left) (slot . 0)
	(preserve-size . (nil . t)))
  (push it display-buffer-alist))
```
> I was learning about cl-loop, so I made a few commands that use it on the Selectrum wiki. I like cl-loop, but I am learning that it is divisive.

Sorry if I seemed harsh about cl-loop. The reasons I don't use it is because it
does not use lisp syntax, so you loose many of the benefits of using lisp by
using cl-loop. And since Emacs is designed to indent lisp, `cl-loop` breaks indentation. Furthermore, the "conveniences" that cl-loop provides with its clauses are generally not much shorter than the
equivalent lisp code. If you like cl-loop I would recommend giving
[loopy](https://github.com/okamsn/loopy#introduction) a try. It provides a
similar abstraction but it retains lisp syntax. One consequence of this is you
clauses arbitrarily nested in lisp forms which is something you can't do in
`cl-loop`.

I haven't completely formed an opinion about how useful loopy is. Certainly, it
is better than `cl-loop`. If it is useful, it's usefulness lies in complex loops where you need to collect into many variables; and where the control flow is not fluid (as in you have breaks, skips).

minad:
@okamsn @Luis-Henriquez-Perez While I think loopy/iterate macro is an improvement over cl-loop I would like to avoid introducing dependencies here.

Luis-Henriquez-Perez:
> While I think loopy/iterate macro is an improvement over cl-loop I would like to avoid introducing dependencies here.

Sure. However, I meant in he should try it in general; not that it should be used for this package.

minad:
@Luis-Henriquez-Perez I don't understand - in case you missed, @okamsn is the author of loopy?!

Luis-Henriquez-Perez:
>  I don't understand - in case you missed, @okamsn is the author of loopy?!

@minad Yes, I said that before I realized he was the author. I apologize for the confusion.

Luis-Henriquez-Perez:
This is a gif of consult info as it currently stands. 

output-2020-12-14-16:32:01

This is the selectrum info command.

output-2020-12-14-16:39:25

I definitely think the selectrum info display certainly is more polished.

@minad Can you expand more about how you'd like the info commands to display the entries? Is `selectrum-info` more of what you'd like?

minad:
Oh yes, selectrum info ist much nicer!

Luis-Henriquez-Perez:
@minad Ok. I'll work on adopting selectrum info's display. The annotations won't be included though right? Not sure how marginalia works but seems like it will take care of that?

minad:
> The annotations won't be included though right? Not sure how marginalia works but seems like it will take care of that?

I think the annotations are shown directly by selectrum-info in the gif? For now you should not rely on marginalia providing additional annotations and jut get the basic consult-info command to work. If we feel that annotations will help we can add them later.

okamsn:
> > The annotations won't be included though right? Not sure how marginalia works but seems like it will take care of that?
> 
> I think the annotations are shown directly by selectrum-info in the gif?

Yes, `selectrum-info` finds topic headings and their descriptions in one pass.

> For now you should not rely on marginalia providing additional annotations and jut get the basic consult-info command to work. If we feel that annotations will help we can add them later by doing some minor adjustments.

In my opinion,  the topic descriptions are helpful to have in the candidate, because they provide more context and more to match against. For example, in the below image, when I search for "regexp", I get several topics that are clearly related to my search, even when they do not have "Regexp" in their title. I now forget what topic in particular I had trouble finding before adding the descriptions, but I do know that it is useful that they are part of the candidate.

example

minad:
I am closing this because of conflicts with the main branch, please reopen if you still want to follow through with this :)

PR #54, opened 2020-12-11
Split package, such that consult.el depends only on Emacs core components

minad:
The goal is to separate completion-system-specific logic completely from the rest. Selectrum-specific logic has been moved to consult-selectrum.el. Furthermore flycheck-specific has been moved to consult-flycheck.el.

Note, that there is still 'consult-buffer` which contains selectrum-specific code. This is not addressed here and ~will be adressed later, see discussion in #10.~ has been addressed in #55.

I think I am happy with now - the code is indeed cleaner since no completion-system specifics are mixed with the rest of the code. There is also no regression in behavior. The only disadvantage is that some users have to install three packages now. In contrast, for users who don't use flycheck/selectrum this is an advantage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,10 +99,12 @@ completions with richer information (e.g. `M-x`, `describe-face`,
     and recursive editing.
   * `consult-outline`: Jump to a heading of the outline. Supports live preview
     and recursive editing.
-  * `consult-flycheck`: Jump to flycheck error. Supports live preview
-    and recursive editing.
   * `consult-imenu`: Jump to imenu item. Supports live preview
     and recursive editing.
+  * `consult-flycheck`: Jump to flycheck error. Supports live preview
+    and recursive editing. This command requires to install the additional
+    `consult-flycheck.el` package since the main `consult.el` package
+    only depends on Emacs core components.
 
 ### Miscellaneous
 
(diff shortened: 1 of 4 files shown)


minad:
@purcell @manuel-uberti @s-kostyaev Please take a look.

s-kostyaev (approved):
LGTM. Although, I don't use any of this external packages.

manuel-uberti:
I don't use Flycheck, so I'll let others chime in on this, but if one wants to use `consult-selectrum` without the previews would that be possible?

minad:
> I don't use Flycheck, so I'll let others chime in on this, but if one wants to use consult-selectrum without the previews would that be possible?

Yes, it will work in any case. If you don't activate consult-preview-mode, you won't have previews. If you don't install consult-selectrum, you can also not have previews and furthermore you miss some other minor selectrum-specific tweaks. If using selectrum, you should use consult-selectrum just to be sure.

manuel-uberti:
Great stuff, then. Thank you for working on this.

minad:
@purcell Do you agree with the changes here? If I merge this can you please ensure that MELPA gets consult-flycheck and consult-selectrum packages?

purcell:
Yes, great, I can synch things on the MELPA side when merged.

purcell:
To be clear, the way to go is to merge this first, at which point MELPA would end up with a multi-file `consult` package if it builds immediately: this would still work for users. And once the files are in place I can add/update recipes as necessary.

minad:
> To be clear, the way to go is to merge this first, at which point MELPA would end up with a multi-file consult package if it builds immediately: this would still work for users. And once the files are in place I can add/update recipes as necessary.

Yes, that is clear. I merged this PR!

minad:
@purcell I made a PR here: https://github.com/melpa/melpa/pull/7287/commits/9cfecb5655cb9f638f1f207317a4f8f2a7d16341

purcell:
MELPA recipes are updated now and should build in the next cycle. :-)

minad:
@purcell Great! Thank you!

purcell:
Sorry, I had commits for this waiting locally and just pushed them, so I didn't see your PR first.

minad:
@purcell That's okay. I guess your version is better since it uses :exclude.

PR #58, opened 2020-12-12
Use different approach to install narrow setup

clemera:
This simplifies the way the keymap gets setup which allows removing the `consult--minibuffer-map-hook` and the menu filter for `SPC` also makes sure the `C-h k` works as expected and if `SPC` is bound to something else than `self-insert-command` it will continue to do so.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -286,7 +286,7 @@ PREVIEW is the preview function."
          (propertize (concat prefix consult-narrow-separator " ") 'display "")
          strings))
 
-(defun consult-insert ()
+(defun consult-insert-for-narrow ()
   "Insert narrowing prefix, see `consult-narrow-separator'."
   (interactive)
   (insert (concat consult-narrow-separator " ")))
@@ -304,7 +304,7 @@ CHARS is the list of narrowing prefix strings."
                            (&optional _)
                            (let ((str (minibuffer-contents)))
                              (when (member str chars)
-                               'consult-insert)))))
+                               'consult-insert-for-narrow)))))
           (use-local-map keymap)))
     (funcall body)))
 
--- a/consult.el
+++ b/consult.el
@@ -223,10 +223,6 @@ does not occur in candidate strings."
 
 ;;;; Internal variables
 
-(defvar consult--minibuffer-map-hook nil
-  "Get minibuffer keymap.
-For each completion system, a function must be added here.")
-
 (defvar consult--gc-threshold 67108864
   "Large gc threshold for temporary increase.")
 
@@ -290,32 +286,27 @@ PREVIEW is the preview function."
          (propertize (concat prefix consult-narrow-separator " ") 'display "")
          strings))
 
+(defun consult-insert ()
+  "Insert narrowing prefix, see `consult-narrow-separator'."
+  (interactive)
+  (insert (concat consult-narrow-separator " ")))
+
 (defun consult--narrow-install (chars body)
   "Install narrowing in BODY.
 
 CHARS is the list of narrowing prefix strings."
-  (let* ((keymap (or (run-hook-with-args-until-success 'consult--minibuffer-map-hook)
-                     ;; Use minibuffer-local-completion-map by default
-                     minibuffer-local-completion-map))
-         (stack)
-         (setup (lambda ()
-                   (push (lookup-key keymap " ") stack)
-                   (unless (cdr stack)
-                     (define-key keymap " "
-                       (lambda ()
-                         (interactive)
-                         (let ((str (minibuffer-contents)))
-                           (if (member str chars)
-                               (insert (concat consult-narrow-separator " "))
-                             (call-interactively 'self-insert-command))))))))
-         (exit (lambda ()
-                  (define-key keymap " " (pop stack)))))
-    (add-hook 'minibuffer-setup-hook setup)
-    (add-hook 'minibuffer-exit-hook exit)
-    (unwind-protect
-        (funcall body)
-      (remove-hook 'minibuffer-setup-hook setup)
-      (remove-hook 'minibuffer-exit-hook exit))))
+  (minibuffer-with-setup-hook
+      (lambda ()
+        (let ((keymap (make-composed-keymap (current-local-map) nil)))
+          (define-key keymap " "
+            `(menu-item "" nil :filter
+                        ,(lambda
+                           (&optional _)
+                           (let ((str (minibuffer-contents)))
+                             (when (member str chars)
+                               'consult-insert)))))
+          (use-local-map keymap)))
+    (funcall body)))
 
 (defmacro consult--with-narrow (chars &rest body)
   "Setup narrowing in BODY.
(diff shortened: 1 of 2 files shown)


minad:
@clemera This looks better! Please remove the hook also in consult-selectrum.el. Does this work with recursive minibuffers, does it work with both icomplete-vertical and selectrum?

clemera:
I haven't explicitly tested with `icomplete` because it's always such a hassle for me to set things up for it. But I will do just to be sure.

clemera:
Works with `emacs -Q` in icomplete-vertical, too.

minad:
I use the following to switch back and forth

~~~
;; icomplete vertical display
(use-package icomplete-vertical
  :bind (:map icomplete-minibuffer-map
              ("SPC" . self-insert-command)
              ("<down>" . icomplete-forward-completions)
              ("<up>" . icomplete-backward-completions))
  :config
  (setq-default icomplete-compute-delay 0))

(defun toggle-completion ()
  (interactive)
  (if selectrum-mode
      (progn
        (selectrum-mode -1)
        (fido-mode 1)
        (icomplete-vertical-mode 1))
    (fido-mode -1)
    (icomplete-vertical-mode -1)
    (selectrum-mode 1)))
~~~

clemera:
Thanks, I should probably add that to my config because it's likely I will have to test things for it in the future ;) Also yes to recursive minibuffer question above by the way.

clemera:
Please don't merge yet I have to check a potential problem I just discovered.

clemera:
I though `make-composed-keymap` would allow to not modify the original map but it didn't. I switched to `copy-keymap` which doesn't has this problem.

minad:
Ok, copy keymap is maybe a bit more expensive. Should not matter hopefully. But what I find scary about your approach is that you actually replace the local keymap - isn't that too intrusive?

clemera:
Turned out I just was using it wrong. Now it's unexpensive again and doesn't modify the original map.

clemera:
> But what I find scary about your approach is that you actually replace the local keymap - isn't that too intrusive?

You mean this is a problem when someone modifies the keymap after you copied it? With my last change the original keymap is set as the parent so this should work as expected.

clemera:
And note this is only affects the current session.

minad:
> You mean this is a problem when someone modifies the keymap after you copied it? With my last change the original keymap is set as the parent so this should work as expected.

I wonder about weird side effects. Maybe something expects the local keymap to be identical to some keymap instance. But if you use the original keymap as parent keymap I don't expect too many issues. What do you dislike about the current method? The hook and that it actually performs a modification? Instead of using a hook I tried to modify the local map, but that didn't work.

> And note this only affects the current session it doesn't change the keymap permanently.

Yes, that is clear.

clemera:
Yes I did not like that every framework needs to install a setup in the hook. And also that `SPC` might be bound to something else than `self-insert-command`.

minad:
Ok, makes sense. The self-insert-command issue could be fixed by making a lookup first. I also tried to use overriding-local-map but I also could not get that to work. I would actually prefer overriding-local-map if that is possible.

clemera:
Using `overriding-local-map` could break things badly, because it changes how keys are looked up, I wouldn't recommend to use it for this.

minad:
Yes it messes with the order completely. Originally I expected it just to sit on top and take precedence.

consult.el
```diff
@@ -290,32 +286,27 @@ PREVIEW is the preview function."
...
-    (add-hook 'minibuffer-exit-hook exit)
-    (unwind-protect
-        (funcall body)
-      (remove-hook 'minibuffer-setup-hook setup)
-      (remove-hook 'minibuffer-exit-hook exit))))
+  (minibuffer-with-setup-hook
+      (lambda ()
+        (let ((keymap (make-composed-keymap nil (current-local-map))))
+          (define-key keymap " "
+            `(menu-item "" nil :filter
```
minad, line 302: What is menu-item?

minad:
@clemera Is this ready from your side? If bad side effects occur we can still revert, so I would be okay to merge this.

clemera:
Yes, I'm done.

consult.el
minad: Ok! Thanks!

minad:
@clemera It does not seem to work for me with icomplete. I reverted it again for now.

clemera:
Strange seems to work fine for me with `icomplete` and `emacs -Q`

minad:
@clemera This is weird. The problem seems to be if I start with selectrum-mode enabled and then use my toggler.

clemera:
I also noticed some other problems with the current approach. The stack isn't reused by further invokations and after I started edebug to check for this I noticed the hooks continue to run even when not in any `consult-buffer` session.

clemera:
> EDIT: No it is not.

Does this mean it works or it does not work but the toggler was not the problem?

minad:
> Does this mean it works or it does not work but the toggler was not the problem?

No problem with the toggler. Still does not work for me.

clemera:
What does not work? The insertion of the separator prefix?

minad:
Yes, the keymap is not used. The command for space is not called. This is a problem I had before when I tried to use the local map with icomplete. 

I wrote above:

> Instead of using a hook I tried to modify the local map, but that didn't work.

clemera:
Does it work for you without `icomplete`, I mean plain minibuffer completion?

minad:
@clemera Yes, it does.

clemera:
Okay, than it might be something with you `icomplete` config? It works with `emacs -Q` (v 27).

minad:
@clemera Sorry, I think the bug is in my config. I have this SPC setting. I don't recall why I put that here. But I remember when we discussed this dynamic candidates stuff that you also had this setting.

~~~ elisp
(use-package icomplete-vertical
  :bind (:map icomplete-minibuffer-map
              ("SPC" . self-insert-command)
              ("<down>" . icomplete-forward-completions)
              ("<up>" . icomplete-backward-completions))
  :config
  (setq-default icomplete-compute-delay 0))
~~~

clemera:
Yes I just looked at 

```elisp
(defun icomplete-minibuffer-setup ()
  "Run in minibuffer on activation to establish incremental completion.
Usually run by inclusion in `minibuffer-setup-hook'."
  (when (and icomplete-mode (icomplete-simple-completing-p))
    (set (make-local-variable 'completion-show-inline-help) nil)
    (use-local-map (make-composed-keymap icomplete-minibuffer-map
    					 (current-local-map)))
    (add-hook 'pre-command-hook  #'icomplete-pre-command-hook  nil t)
    (add-hook 'post-command-hook #'icomplete-post-command-hook nil t)
    (run-hooks 'icomplete-minibuffer-setup-hook)))
```

They use a similar approach like this PR. We can run the hook late to make it work.

minad:
@clemera Thanks! It works!

clemera:
Okay, then this should be done hopefully.

minad:
Yes, I just merged it!

clemera:
Great, thanks!

Issue #48, opened 2020-12-10
In eshell's completion-at-point, consult-completion-in-region places the minibuffer point in the middle of the completion string


tomfitzhenry:
# Steps to reproduce

GNU Emacs 27.1
consult: b5cbc619533d61db85714ef4b795795f37be8553

```
$ emacs -q
(require 'consult)
(setq completion-in-region-function 'consult-completion-in-region)
```

In eshell, run:
```
touch foobar
touch football
cat foo<TAB>
```

# Expected
I'm prompted "Completion: foo|" where `|` indicates where I expect point to be.

# Actual
I'm prompted "Completion: |foo".

This is a problem because I then type "b" (to try to complete "foob" to "foobar"), the minibuffer actually shows "bfoo" which fails to complete.

tomfitzhenry:
Workaround: Remove this clause: https://github.com/minad/consult/blob/b5cbc619533d61db85714ef4b795795f37be8553/consult.el#L627-L632

minad:
@oantolin ideas?

oantolin:
Huh. Like @tomfitzhenry pointed out, it's read-file-name that's doing that! I don't know why, but it chooses to put point before the argument `initial` rather than at the end of the minibuffer. That doesn't seem like a very good default. We could just not use `read-file-name`, but it is nice to have your special keybindings from `minibuffer-local-filename-completion-map` present.

How about just forcing the move to the end of line? We can wrap the call to `read-file-name` in `(minibuffer-with-setup-hook #'end-of-line _____)`.

minad:
@tomfitzhenry are you using selectrum or icomplete btw?

tomfitzhenry:
> @tomfitzhenry are you using selectrum or icomplete btw?

I can reproduce this issue with `emacs -q` (so not even icomplete is enabled), and with `icomplete`. I haven't tried selectrum.

oantolin:
It's definitely something that `read-file-name-default` does. Selectrum uses its own setting for `read-file-name-function`, so it's possible it behaves differently.

tomfitzhenry:
> Huh. Like @tomfitzhenry pointed out, it's read-file-name that's doing that! I don't know why, but it chooses to put point before the argument initial rather than at the end of the minibuffer. That doesn't seem like a very good default.

I'll try to reproduce this using just `read-file-name`, and raise that issue upstream. Working around bugs long-term isn't sustainable.

minad:
[...]

PR #63, opened 2020-12-13
Fix issue #48

oantolin:
When completing files with consult-completion-in-region, the point in
the minibuffer gets placed initially at the beginning of the last path
component. With this change it starts at the end of minibuffer
contents, as for other types of completion.


minad:
Maybe add a comment?

oantolin:
> Maybe add a comment?

Oh, yes, I should have done that! I'll do it in a little while. I'm a little git ignorant here: do I make a new pull request for that?

minad:
I already added a comment. I hope it makes sense.

PR #66, opened 2020-12-14
consult-flymake

tomfitzhenry:
consult-flymake is like consult-flycheck, but for Flymake.

Relatedly, I've written https://github.com/tomfitzhenry/flymake-flycheck to use flycheck from flymake.

### DIFF
--- a/consult-flymake.el
+++ b/consult-flymake.el
@@ -52,8 +52,9 @@
     (user-error "No Flymake diagnostics"))
   (let* ((diagnostics (flymake-diagnostics))
          (buffer-width (apply #'max (mapcar (lambda (x) (length (format "%s" (flymake-diagnostic-buffer x)))) diagnostics)))
+         (line-width (apply #'max (mapcar (lambda (x) (length (format "%s" (consult-flymake--diag-line x)))) diagnostics)))
          (category-width (apply #'max (mapcar (lambda (x) (length (consult-flymake--diag-type-name x))) diagnostics)))
-         (fmt (format "%%-%ds %%-%ds %%s" buffer-width category-width)))
+         (fmt (format "%%-%ds %%-%ds %%-%ds %%s" buffer-width line-width category-width)))
     (mapcar
      (lambda (diag)
        (with-current-buffer (flymake--diag-buffer diag)
@@ -67,6 +68,7 @@
              (_ "n"))
            (format fmt
                    (flymake-diagnostic-buffer diag)
+                   (consult-flymake--diag-line diag)
                    (propertize
                     (consult-flymake--diag-type-name diag)
                     'face
@@ -90,6 +92,13 @@
                             ("n" . "Note"))
                   :lookup #'consult--lookup-list
                   :preview (and consult-preview-flymake #'consult--preview-position))))
+
+(defun consult-flymake--diag-line (diag)
+  "Return the line number of DIAG."
+  (save-excursion
+    (with-current-buffer (flymake--diag-buffer diag)
+      (goto-char (flymake--diag-beg diag))
+      (line-number-at-pos))))
 
 (provide 'consult-flymake)
 ;;; consult-flymake.el ends here
--- a/consult-flymake.el
+++ b/consult-flymake.el
@@ -67,13 +67,14 @@
      (lambda (diag)
        (with-current-buffer (flymake-diagnostic-buffer diag)
          (goto-char (flymake-diagnostic-beg diag)))
-       (let ((type (flymake-diagnostic-type diag)))
+       (let* ((type (flymake-diagnostic-type diag))
+              (category (get type 'flymake-category)))
          (cons
           (consult--narrow-candidate
-           (pcase type
-             (:error "e")
-             (:warning "w")
-             (_ "n"))
+           (pcase category
+             ('flymake-error ?e)
+             ('flymake-warning ?w)
+             (_ ?n))
            (format fmt
                    (flymake-diagnostic-buffer diag)
                    (consult-flymake--diag-line diag)
@@ -95,9 +96,9 @@
                   :category 'flymake-error
                   :require-match t
                   :sort nil
-                  :narrow '(("e" . "Error")
-                            ("w" . "Warning")
-                            ("n" . "Note"))
+                  :narrow '((?e . "Error")
+                            (?w . "Warning")
+                            (?n . "Note"))
                   :lookup #'consult--lookup-list
                   :preview (and consult-preview-flymake #'consult--preview-position))))
 
--- a/consult-flymake.el
+++ b/consult-flymake.el
@@ -38,12 +38,19 @@
   :type 'boolean
   :group 'consult-preview)
 
-(defsubst consult-flymake--diag-type-name (diag)
-  "Return the type name of DIAG."
-  (let ((type (flymake-diagnostic-type diag)))
-    (format "%s"
-            (flymake--lookup-type-property
-             type 'flymake-type-name type))))
+(defsubst consult-flymake--diag-line (diag)
+  "Return the line number of DIAG."
+  (save-excursion
+    (with-current-buffer (flymake-diagnostic-buffer diag)
+      (goto-char (flymake-diagnostic-beg diag))
+      (line-number-at-pos))))
+
+(defsubst consult-flymake--type-property (type prop)
+  "Return the PROP of TYPE."
+  (let* ((category (get type 'flymake-category))
+         (default (get category prop))
+         (val (get type prop)))
+    (or val default)))
 
 (defun consult-flymake--candidates ()
   "Return Flymake errors as alist."
@@ -53,8 +60,9 @@
   (let* ((diagnostics (flymake-diagnostics))
          (buffer-width (apply #'max (mapcar (lambda (x) (length (format "%s" (flymake-diagnostic-buffer x)))) diagnostics)))
          (line-width (apply #'max (mapcar (lambda (x) (length (format "%s" (consult-flymake--diag-line x)))) diagnostics)))
-         (category-width (apply #'max (mapcar (lambda (x) (length (consult-flymake--diag-type-name x))) diagnostics)))
-         (fmt (format "%%-%ds %%-%ds %%-%ds %%s" buffer-width line-width category-width)))
+         (type-name-width (apply #'max (mapcar (lambda (x) (length (consult-flymake--type-property
+                                                                    (flymake-diagnostic-type x) 'flymake-type-name))) diagnostics)))
+         (fmt (format "%%-%ds %%-%ds %%-%ds %%s" buffer-width line-width type-name-width)))
     (mapcar
      (lambda (diag)
        (with-current-buffer (flymake-diagnostic-buffer diag)
@@ -70,9 +78,9 @@
                    (flymake-diagnostic-buffer diag)
                    (consult-flymake--diag-line diag)
                    (propertize
-                    (consult-flymake--diag-type-name diag)
+                    (consult-flymake--type-property type 'flymake-type-name)
                     'face
-                    (flymake--lookup-type-property (flymake-diagnostic-type diag) 'mode-line-face))
+                    (consult-flymake--type-property type 'mode-line-face))
                    (flymake-diagnostic-text diag)))
           (point-marker))))
      (flymake-diagnostics))))
@@ -92,13 +100,6 @@
                             ("n" . "Note"))
                   :lookup #'consult--lookup-list
                   :preview (and consult-preview-flymake #'consult--preview-position))))
-
-(defsubst consult-flymake--diag-line (diag)
-  "Return the line number of DIAG."
-  (save-excursion
-    (with-current-buffer (flymake-diagnostic-buffer diag)
-      (goto-char (flymake-diagnostic-beg diag))
-      (line-number-at-pos))))
 
 (provide 'consult-flymake)
 ;;; consult-flymake.el ends here
--- a/consult-flymake.el
+++ b/consult-flymake.el
@@ -1,0 +1,95 @@
+;;; consult-flymake.el --- Provides the command `consult-flymake'  -*- lexical-binding: t; -*-
+
+;; Author: Daniel Mendler, Consult and Selectrum contributors
+;; Maintainer: Daniel Mendler
+;; Created: 2020
+;; License: GPL-3.0-or-later
+;; Version: 0.1
+;; Package-Requires: ((consult "0.1") (emacs "26.1"))
+;; Homepage: https://github.com/minad/consult
+
+;; This file is not part of GNU Emacs.
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Provides the command `consult-flymake'.  This is an extra package,
+;; to avoid loading Flymake.
+
+;;; Code:
+
+(require 'consult)
+(require 'flymake)
+
+(defcustom consult-preview-flymake t
+  "Enable Flymake preview during selection."
+  :type 'boolean
+  :group 'consult-preview)
+
+(defun consult-flymake--diag-type-name (diag)
+  "Return the type name of DIAG."
+  (let ((type (flymake-diagnostic-type diag)))
+    (format "%s"
+            (flymake--lookup-type-property
+             type 'flymake-type-name type))))
+
+(defun consult-flymake--candidates ()
+  "Return Flymake errors as alist."
+  (consult--forbid-minibuffer)
+  (unless (flymake-diagnostics)
+    (user-error "No Flymake diagnostics"))
+  (let* ((diagnostics (flymake-diagnostics))
+         (buffer-width (apply #'max (mapcar (lambda (x) (length (format "%s" (flymake-diagnostic-buffer x)))) diagnostics)))
+         (category-width (apply #'max (mapcar (lambda (x) (length (consult-flymake--diag-type-name x))) diagnostics)))
+         (fmt (format "%%-%ds %%-%ds %%s" buffer-width category-width)))
+    (mapcar
+     (lambda (diag)
+       (with-current-buffer (flymake--diag-buffer diag)
+         (goto-char (flymake--diag-beg diag)))
+       (let ((type (flymake-diagnostic-type diag)))
+         (cons
+          (consult--narrow-candidate
+           (pcase type
+             (:error "e")
+             (:warning "w")
+             (_ "n"))
+           (format fmt
+                   (flymake-diagnostic-buffer diag)
+                   (propertize
+                    (consult-flymake--diag-type-name diag)
+                    'face
+                    (flymake--lookup-type-property (flymake-diagnostic-type diag) 'mode-line-face))
+                   (flymake-diagnostic-text diag)))
+          (point-marker))))
+     (flymake-diagnostics))))
+
+;;;###autoload
+(defun consult-flymake ()
+  "Jump to Flymake diagnostic."
+  (interactive)
+  (consult--jump
+   (consult--read "Flymake diagnostic: "
+                  (consult--with-increased-gc (consult-flymake--candidates))
+                  :category 'flymake-error
+                  :require-match t
+                  :sort nil
+                  :narrow '(("e" . "Error")
+                            ("w" . "Warning")
+                            ("n" . "Note"))
+                  :lookup #'consult--lookup-list
+                  :preview (and consult-preview-flymake #'consult--preview-position))))
+
+(provide 'consult-flymake)
+;;; consult-flymake.el ends here
--- a/consult-flymake.el
+++ b/consult-flymake.el
@@ -57,8 +57,8 @@
          (fmt (format "%%-%ds %%-%ds %%-%ds %%s" buffer-width line-width category-width)))
     (mapcar
      (lambda (diag)
-       (with-current-buffer (flymake--diag-buffer diag)
-         (goto-char (flymake--diag-beg diag)))
+       (with-current-buffer (flymake-diagnostic-buffer diag)
+         (goto-char (flymake-diagnostic-beg diag)))
        (let ((type (flymake-diagnostic-type diag)))
          (cons
           (consult--narrow-candidate
@@ -96,8 +96,8 @@
 (defsubst consult-flymake--diag-line (diag)
   "Return the line number of DIAG."
   (save-excursion
-    (with-current-buffer (flymake--diag-buffer diag)
-      (goto-char (flymake--diag-beg diag))
+    (with-current-buffer (flymake-diagnostic-buffer diag)
+      (goto-char (flymake-diagnostic-beg diag))
       (line-number-at-pos))))
 
 (provide 'consult-flymake)
--- a/consult-flymake.el
+++ b/consult-flymake.el
@@ -38,7 +38,7 @@
   :type 'boolean
   :group 'consult-preview)
 
-(defun consult-flymake--diag-type-name (diag)
+(defsubst consult-flymake--diag-type-name (diag)
   "Return the type name of DIAG."
   (let ((type (flymake-diagnostic-type diag)))
     (format "%s"
@@ -93,7 +93,7 @@
                   :lookup #'consult--lookup-list
                   :preview (and consult-preview-flymake #'consult--preview-position))))
 
-(defun consult-flymake--diag-line (diag)
+(defsubst consult-flymake--diag-line (diag)
   "Return the line number of DIAG."
   (save-excursion
     (with-current-buffer (flymake--diag-buffer diag)


consult.el
```diff
@@ -1312,5 +1316,33 @@ Prepend PREFIX in front of all items."
...
 
+(defun consult--flymake-candidates ()
+  "Return Flymake errors as alist."
+  (consult--forbid-minibuffer)
+  (let* ((fmt (format "%s" "foo")))
+    (mapcar
+     (lambda (diag)
+       (goto-char (flymake-diagnostic-beg diag))
+       (cons
+        (format "%s %s %s"
```
tomfitzhenry, line 1327: I should propertize this string,.

minad: Yes, ideally with the flymake fonts. It would be great if you can align the formatting with the table formatting of the flymake buffer. Please take a look at how the flycheck command does it.

minad:
I wonder what the point of this flycheck/flymake bridge is. If you already have flycheck installed why do you want the flymake frontend? What are the advantages then? You don't avoid the dependencies and the flycheck frontend is less "shiny" than the flycheck frontend as far as I see it.

tomfitzhenry:
I predominantly use Flymake over Flycheck because eglot supports Flymake. But I'd like still like good diagnostic functions for modes not supported by eglot, and Flycheck happens to be a source of many good checkers.

I'd like the same UI for these, hence my motivation for adapting one to the other.

I looked into both frameworks, and found their differences to be small, except that Flymake is built in, which is a strong plus for myself.

I need to add this rationale to a README in https://github.com/tomfitzhenry/flymake-flycheck .

manuel-uberti:
FWIW, I tested the code. I had to use `consult--jump` instead of `consult--goto` to make it work, but thanks a lot for this.

minad:
@tomfitzhenry Makes sense to have a unified UI. 

But I am not sure if having something built-in is a strong plus. There are too many examples of packages which were superseded by external packages (flymake, project, ido, ...). Then there are package which are well-intended but did not really work out (semantic). Personally I would prefer a smaller core Emacs.

Regarding the rationale being similar to consult-vs-counsel, I am not sure. The point of consult is that `completing-read` is a core API, it is even implemented in C code, since it is a fundamental UI interaction feature. This means if you work around that one you have to reinvent *everything* which uses this API.  Luckily with consult/marginalia/embark we only have to implement fewer commands since we can enhance most existing commands with annotations, actions etc after the fact without modifying them.

In contrast, flymake is just a package being maintained as part of the Emacs code base. I don't see why that makes it architecturally or technically any better than an external package. Sure, it makes sense to have a commonly agreed API somewhere - possibly in core, like the flymake API or externally like the flycheck API. The situation as of now is just that we have two widely used APIs, but as far as I see, flycheck has many more checkers available?

minad:
@manuel-uberti @tomfitzhenry But let me make it clear - while I am using and preferring flycheck myself, I think it is a good idea to support flymake too as part of consult. As far as I see it, this PR just needs a bit more polishing similar to what I did with consult-flycheck (narrowing support, polished rendering, using markers such that recursive editing works etc). Then I will pull this in!

manuel-uberti:
In my case, my preference for `project` and `flymake` is due to the fact that I find them simpler than their counterparts and that I like to rely on external packages as little as possible. And not only when it comes to Emacs, by the way.

This not to say that Projectile and Flycheck are bad. On the contrary, they are great tools which I used a lot, but I found it worth to give what's built-in a shot before jumping on something external, and maybe take a chance to improve on that. If something built-in works for me, I see no need to pull in a package with its dependencies. 

But again, this is very personal and dictated by my own experience. YMMV.

s-kostyaev:
> In my case, my preference for `project` and `flymake` is due to the fact that I find them simpler than their counterparts and that I like to rely on external packages as little as possible. And not only when it comes to Emacs, by the way.
> 
> This not to say that Projectile and Flycheck are bad. On the contrary, they are great tools which I used a lot, but I found it worth to give what's built-in a shot before jumping on something external, and maybe take a chance to improve on that. If something built-in works for me, I see no need to pull in a package with its dependencies.
> 
> But again, this is very personal and dictated by my own experience. YMMV.

Same works for me. Also, I had some problems with `projectile` and `flycheck` on Emacs master branch. Once I had this issues I have replaced it with built-in alternatives. When I miss checker, I create it for flymake myself.

minad:
@manuel-uberti Well the problem with the built-in things is also that they have a different development approach - more focused on stability, slower development, fewer contributors, more hassle to contribute etc. I think that works great for core functionality, but not so much for more experimental things.

@s-kostyaev I wonder about the problems. I also had bugs with projectile from time to time, but Bozhidar is very responsive. Things can be fixed in a day if you write a patch.

s-kostyaev:
> @manuel-uberti Well the problem with the built-in things is also that they have a different development approach - more focused on stability, slower development, fewer contributors, more hassle to contribute etc. I think that works great for core functionality, but not so much for more experimental things.
> 
> @s-kostyaev I wonder about the problems. I also had bugs with projectile from time to time, but Bozhidar is very responsive. Things can be fixed in a day if you write a patch.

That was when `projectile` was without maintainer.

minad:
> That was when flycheck was without maintainer.

Ok. At that point I would probably also considered to switch. Usually I drop unmaintained/buggy packages very quickly since it can be a huge time sink - additionally to the huge time sink which Emacs already is due to its procrastination potential ;)

manuel-uberti:
> @manuel-uberti Well the problem with the built-in things is also that they have a different development approach - more focused on stability, slower development, fewer contributors, more hassle to contribute etc. I think that works great for core functionality, but not so much for more experimental things.

Absolutely, but I've come to prefer stability over experimental things, which doesn't mean I stopped having fun with new Emacs packages altogether. :)

minad:
@manuel-uberti I fully agree in particular regarding essentials. This is the main reason why I never started with helm - I immediately had issues with it, but this is years ago.

consult.el
tomfitzhenry: Done: flymake fonts are used, and columns are aligned.

manuel-uberti:
@tomfitzhenry Tried your latest commits. Still good to me. :+1:

tomfitzhenry:
Should I squash commits?

consult.el
```diff
@@ -44,6 +44,7 @@
 
 (require 'bookmark)
 (require 'cl-lib)
+(require 'flymake)
```
minad, line 47: Please move the command to an extra file `consult-flymake.el` such that flymake is not always loaded. This makes sense for the flycheck users. The file `consult-flycheck.el` can still be distributed as part of the MELPA `consult` package in contrast to `consult-flycheck`.

consult.el
```diff
@@ -1312,5 +1317,40 @@ Prepend PREFIX in front of all items."
...
 (add-hook 'consult-preview-mode-hook #'consult--icomplete-preview-setup)
 
+(defun consult--flymake-candidates ()
+  "Return Flymake errors as alist."
+  (consult--forbid-minibuffer)
+  (let* ((diagnostics (flymake-diagnostics))
+	 (buffer-width (apply #'max (mapcar (lambda (x) (length (format "%s" (flymake-diagnostic-buffer x)))) diagnostics)))
+	 (category-width (apply #'max (mapcar (lambda (x) (length (symbol-name (flymake-diagnostic-type x)))) diagnostics)))
+         (fmt (format "%%-%ds %%-%ds %%s" buffer-width category-width)))
+    (mapcar
```
minad, line 1327: Yes, it is kind of like reimplementing tabulated mode, but it is good to have no dependency here and it is better for performance like this. Marginalia does the same.

consult.el
```diff
@@ -1312,5 +1317,40 @@ Prepend PREFIX in front of all items."
...
+     (lambda (diag)
+       (with-current-buffer (flymake--diag-buffer diag)
```
minad, line 1329: Fix indentation using emacs please.

consult.el
```diff
@@ -1312,5 +1317,40 @@ Prepend PREFIX in front of all items."
...
+	(point-marker)))
+     (flymake-diagnostics))))
+
+;;;###autoload
+(defun consult-flymake ()
+  "Jump to Flymake diagnostic."
+  (interactive)
+  (consult--jump
+   (consult--read "Flymake diagnostic: "
+                  (consult--with-increased-gc (consult--flymake-candidates))
```
minad, line 1348: Please add narrowing support via :narrow and consult--narrow-candidate. See consult-flycheck.

minad (changes requested):

consult.el
tomfitzhenry: Why is unnecessarily loading Flymake a problem? It only activates if flymake-mode enables, and thus wouldn't impact Flycheck users.

consult.el
tomfitzhenry: SGTM.

consult.el
tomfitzhenry: Done.

consult.el
minad: Generally it is better to avoid loading any unnecessary dependency. We will do the same if we add consult-org-* commands. They will go to a consult-org.el file. The dependencies of consult.el should be rather lightweight as of now.

tomfitzhenry: That makes sense. How about lazily requiring flymake, per https://github.com/minad/consult/pull/66/commits/af1401cf8cbe0edd14906d2a3dae49f8221f49c4 ?

This allows non-Flymake users to avoid having to load Flymake, but allows `consult-flymake` to be able and discoverable by all users.

consult.el
```diff
@@ -1319,6 +1318,7 @@ Prepend PREFIX in front of all items."
 
 (defun consult--flymake-candidates ()
   "Return Flymake errors as alist."
+  (require 'flymake)
```
minad, line 1321: No. This is not good enough, since this breaks package-lint. It has to go to an extra file.

consult.el
minad: @tomfitzhenry In case you worry. This will not break lazy loading. It will just work. Everything will work if it is in an extra file. The only advantage is that flymake is not loaded for users who don't need it. Furthermore this lazy requiring breaks package-lint is therefore not a good idea.

consult.el
tomfitzhenry: Take two, how about using declare-function declarations, to satisfy package-lint. This is a commonly used pattern, e.g. https://github.com/emacs-mirror/emacs/blob/c6c4e746036c5ba714bfce565de13d713f3cb8b5/lisp/emacs-lisp/package.el#L1304

consult.el
tomfitzhenry: Ah, yes. Fair point. Autoload will continue to work.  Ok, I'll move it to a separate file.

consult.el
manuel-uberti: Wouldn't it be easier to add an extra file like `consult-flycheck` does?

tomfitzhenry: Ignore this. I agree moving this to a separate file is a better option.

minad: No, this is not good enough since this breaks if flymake changes. if you are doing that you are basically circumventing static analysis.

consult.el
tomfitzhenry: Done.

consult.el
tomfitzhenry: Done.

tomfitzhenry:
Ah, the file move broke something. Let me fix.

consult.el
```diff
@@ -1312,5 +1312,6 @@ Prepend PREFIX in front of all items."
 
 (add-hook 'consult-preview-mode-hook #'consult--icomplete-preview-setup)
 
+
```
manuel-uberti, line 1315: Is this necessary?

consult.el
tomfitzhenry: I think this is done. I've converted tabs to spaces, and indented using indent-pp-sexp?

consult.el
tomfitzhenry: Fixed.

manuel-uberti:
I'd squash the commit, but I'd let @minad have the final word here of course. :)

minad (changes requested):
Yes, please squash. Line numbers are missing in the view, this should be added.

tomfitzhenry:
> Yes, please squash. 

I've squashed the commits.

> Line numbers are missing in the view, this should be added.

Are line numbers useful for a Flymake UI in the minibuffer? After all, there's nothing the user can do with that information while in the minibuffer. I can see why line numbers are included `flymake-show-diagnostics-buffer`, since that might be a separate window, and some user may look at that window to find the line number and `goto-line` to it, but this wouldn't be true of a minibuffer UI.

tomfitzhenry:
I need to work on the rendering of Flymake categories in the minibuffer. For eglot diagnostics, the error is rendered as "eglot-error". It should be "error".

minad:
Line numbers are useful to select an item, for orientation and for consistency with the standard flymake view. Furthermore I like if the command is at least somehow consistent with the flycheck command.

minad:
Regarding the eglot error - does it work if you use the expression from the flymake view? You did something different here.

tomfitzhenry:
Here's what it now looks like, with correct diagnostic type naming (i.e. the same rendering as `flymake-show-diagnostics-buffer`), and with line numbers.

image

The implementation of line number isn't as efficient as I'd like: computing the line length involves `(line-number-at-pos)` which is O(size of file). `consult-flymake` does this for each diagnostic, to compute `line-width`. Flymake only tracks `beg`, rather than `line`, so I don't think we can avoid this inefficiency.

minad:
@tomfitzhenry We can avoid the inefficiency by first sorting the errors by position, then iterating over the errors and lines together. This is the approach I took by consult-outline. Then the complexity will be O(nlogn+m) where n=errors m=lines. But since the numbers of errors is not too large I am not too worried about the O(nm) behavior.

But it would still be nice if you do not call line-number-at-pos repeatedly, remember the line positions when doing the first iteration over the line numbers.

consult-flymake.el
```diff
@@ -0,0 +1,104 @@
...
+
+(require 'consult)
+(require 'flymake)
+
+(defcustom consult-preview-flymake t
+  "Enable Flymake preview during selection."
+  :type 'boolean
+  :group 'consult-preview)
+
+(defun consult-flymake--diag-type-name (diag)
```
minad, line 41: should be inlined

consult-flymake.el
```diff
@@ -0,0 +1,104 @@
...
+(defun consult-flymake--candidates ()
+  "Return Flymake errors as alist."
+  (consult--forbid-minibuffer)
+  (unless (flymake-diagnostics)
+    (user-error "No Flymake diagnostics"))
+  (let* ((diagnostics (flymake-diagnostics))
+         (buffer-width (apply #'max (mapcar (lambda (x) (length (format "%s" (flymake-diagnostic-buffer x)))) diagnostics)))
+         (line-width (apply #'max (mapcar (lambda (x) (length (format "%s" (consult-flymake--diag-line x)))) diagnostics)))
+         (category-width (apply #'max (mapcar (lambda (x) (length (consult-flymake--diag-type-name x))) diagnostics)))
+         (fmt (format "%%-%ds %%-%ds %%-%ds %%s" buffer-width line-width category-width)))
```
minad, line 57: Iterate once over the diagnostics, remember category and line number?

consult-flymake.el
```diff
@@ -0,0 +1,104 @@
...
+(defun consult-flymake ()
+  "Jump to Flymake diagnostic."
+  (interactive)
+  (consult--jump
+   (consult--read "Flymake diagnostic: "
+                  (consult--with-increased-gc (consult-flymake--candidates))
+                  :category 'flymake-error
+                  :require-match t
+                  :sort nil
+                  :narrow '(("e" . "Error")
```
minad, line 90: I made some recent changes to the :narrow format, sorry! Please adjust that!

consult-flymake.el
```diff
@@ -0,0 +1,104 @@
...
+                            ("w" . "Warning")
+                            ("n" . "Note"))
+                  :lookup #'consult--lookup-list
+                  :preview (and consult-preview-flymake #'consult--preview-position))))
+
+(defun consult-flymake--diag-line (diag)
```
minad, line 96: Same here, should be inlined then.

consult-flymake.el
```diff
@@ -0,0 +1,104 @@
...
+     (lambda (diag)
+       (with-current-buffer (flymake--diag-buffer diag)
+         (goto-char (flymake--diag-beg diag)))
+       (let ((type (flymake-diagnostic-type diag)))
+         (cons
+          (consult--narrow-candidate
+           (pcase type
+             (:error "e")
+             (:warning "w")
+             (_ "n"))
```
minad, line 68: This is not compatible with the elgot error I guess? Would it make sense to use flymake--lookup-type-property instead?

consult-flymake.el
```diff
@@ -0,0 +1,104 @@
...
+           (format fmt
+                   (flymake-diagnostic-buffer diag)
+                   (consult-flymake--diag-line diag)
+                   (propertize
+                    (consult-flymake--diag-type-name diag)
+                    'face
+                    (flymake--lookup-type-property (flymake-diagnostic-type diag) 'mode-line-face))
```
minad, line 75: In the flymake source they also have 'flymake-error here? Why?

~~~ elisp
                       ,(propertize (format "%s"
                                            (flymake--lookup-type-property
                                             type 'flymake-type-name type))
                                    'face (flymake--lookup-type-property
                                           type 'mode-line-face 'flymake-error))
~~~

minad (changes requested):
@tomfitzhenry Looks pretty good now! Only a few minor issues remain.

tomfitzhenry:
> But it would still be nice if you do not call line-number-at-pos repeatedly, remember the line positions when doing the first iteration over the line numbers.

This should be part of Flymake. Until then, Flymake users can accept the inefficiency. If it's important to anyone, they can optimise it.

consult-flymake.el
tomfitzhenry: I just tested this, and it works. I think https://github.com/joaotavora/eglot/blob/38971d3c1d647ab6ad839c3b8585394ea8e53b54/eglot.el#L1513 maps eglot-error to :error.

minad:
> This should be part of Flymake. Until then, Flymake users can accept the inefficiency. If it's important to anyone, they can optimise it.

Agree, but still we can try to write it as good as we can here. But note that I am not expecting the O(nlogn+m) optimization. This should indeed be part of the flymake error structure.

manuel-uberti:
Probably it makes sense to @joaotavora for an opinion on this.

joaotavora:
I didn't follow half of this long thread, other than casually noticing   that @tomfitzhenry has made some kind of flymake-flycheck which seems like a good idea to reuse the zillion checkers that flycheck has.

So I don't understand what exactly @manuel-uberti is asking me about.  If you still want my input, please, for the benefit of my scarce time, summarize the question in an intelligible way. Thanks in advance.

manuel-uberti:
@joaotavora you're right, sorry about that. The problem here seems to be that with regard to errors, Flymake only tracks `beg`, rather than `line` as pointed out [here](https://github.com/minad/consult/pull/66#issuecomment-745247712). You can start from there, but I'll let @tomfitzhenry fill you in with more specifics.

minad:
@joaotavora We have no open question here and from my pov there wasn't a need to summon you - there is only one flymake issue we are stumbling over, since line-number-at-pos is a very inefficient function - it would be neat if flymake could cache the line number in its diagnostics structure. However I am sure this has already been proposed somewhere in the flymake issue tracker. We can investigate there and open an issue if feel it to be necessary. Thank you for coming here!

joaotavora:
> However I am sure this has already been proposed somewhere in the flymake issue tracker.

Really?  Can you point me to the issue?  Flymake is part of Emacs.  If there isn't such an issue, please make one with M-x report-emacs-bug

That said, _everything_, even flipping a bit in a CPU register is inefficient if you try to call it enough times. I still don't understand what you are trying to do.   Are you trying to navigate to diagnostics from the minibuffer?  Do you need to eagerly calculate line numbers for all the diagnostics in the file, or can you do it just in time for the diagnostics you want to show?

consult-flymake.el
```diff
@@ -0,0 +1,104 @@
...
+  (unless (flymake-diagnostics)
+    (user-error "No Flymake diagnostics"))
+  (let* ((diagnostics (flymake-diagnostics))
+         (buffer-width (apply #'max (mapcar (lambda (x) (length (format "%s" (flymake-diagnostic-buffer x)))) diagnostics)))
+         (line-width (apply #'max (mapcar (lambda (x) (length (format "%s" (consult-flymake--diag-line x)))) diagnostics)))
+         (category-width (apply #'max (mapcar (lambda (x) (length (consult-flymake--diag-type-name x))) diagnostics)))
+         (fmt (format "%%-%ds %%-%ds %%-%ds %%s" buffer-width line-width category-width)))
+    (mapcar
+     (lambda (diag)
+       (with-current-buffer (flymake--diag-buffer diag)
```
joaotavora, line 60: Using `flymake--` symbols is a recipe for breakage.

minad:
> Really? Can you point me to the issue? Flymake is part of Emacs. If there isn't such an issue, please make one with M-x report-emacs-bug

Agree, this should be discussed as part of an Emacs bug report. It is off-topic here.

joaotavora:
Anyway, I had a look at the code.  You should probably invent a way do that line and column annotations that you presumably display in front of each diagnostic can be  calculated lazily only for the (presumably small) set of diagnostics that you want to show, and not upfront for the full set of diagnostics.  If the latter happens very frequently, it will be slow.

joaotavora:
Alternatively, caching on your side (mind/consult) also works.  Flymake guarantees that diagnostic objects retain identity, so a weak-keyed  hash table should probably be an easy win.

consult-flymake.el
```diff
@@ -0,0 +1,104 @@
...
+             (:error "e")
+             (:warning "w")
+             (_ "n"))
+           (format fmt
+                   (flymake-diagnostic-buffer diag)
+                   (consult-flymake--diag-line diag)
+                   (propertize
+                    (consult-flymake--diag-type-name diag)
+                    'face
+                    (flymake--lookup-type-property (flymake-diagnostic-type diag) 'mode-line-face))
```
tomfitzhenry, line 75: Can this be done without relying on the private `flymake--lookup-type-property`?

consult-flymake.el
```diff
@@ -0,0 +1,104 @@
...
+(defcustom consult-preview-flymake t
+  "Enable Flymake preview during selection."
+  :type 'boolean
+  :group 'consult-preview)
+
+(defsubst consult-flymake--diag-type-name (diag)
+  "Return the type name of DIAG."
+  (let ((type (flymake-diagnostic-type diag)))
+    (format "%s"
+            (flymake--lookup-type-property
```
tomfitzhenry, line 45: Can this be done without relying on the private flymake--lookup-type-property?

minad:
> Anyway, I had a look at the code. You should probably invent a way do that line and column annotations that you presumably display in front of each diagnostic can be calculated lazily only for the (presumably small) set of diagnostics that you want to show, and not upfront for the full set of diagnostics. If the latter happens very frequently, it will be slow.

Thank you for taking a look. Lazy annotations would be possible, but there is some complexity involved and we can probably avoid this. Instead of lazy annotations I would rather try to compute the list of errors in a very efficient way O(n) in the number of lines.

> Alternatively, caching on your side (mind/consult) also works. Flymake guarantees that diagnostic objects retain identity, so a weak-keyed hash table should probably be an easy win.

Sure, but it is another unnecessary layer of complexity. It is not needed.

consult-flymake.el
tomfitzhenry: Perhaps with just `(get (flymake-diagnostic-type diag) 'flymake-type-name)` with a default.

consult-flymake.el
tomfitzhenry: Perhaps with just `(get (flymake-diagnostic-type diag) 'face)` with a default.

consult-flymake.el
tomfitzhenry: I've removed this, and will remove the other uses of private functions (tracked with other comments).

consult-flymake.el
tomfitzhenry: The default can be no face.

consult-flymake.el
tomfitzhenry: The default can be `(flymake-diagnostic-type diag)`.

consult-flymake.el
tomfitzhenry: Oh, flymake-category is a thing.

joaotavora:
> Instead of lazy annotations I would rather try to compute the list of errors in a very efficient way, e.g., O(n) in the number of lines.

How would you do that? "Lines are just newline characters in a big stream of bytes..

> Sure, but it is another unnecessary layer of complexity. It is not needed.

Seems you are asking for this very complexity to be added to `flymake.el` and exposed as a new interface. I don't think that's a great idea when you have more options in the program making use of Flymake's API.

minad:
> How would you do that? "Lines are just newline characters in a big stream of bytes..

Well, formally you are right. It will be O(b) in the number of bytes b. Nothing can be done about that. So let us just assume b=c*n, where n=number-of-lines.

But note that right now we are O(n*m) with n=number-of-lines and m=number-of-errors. This can be improved to O(mlogm+m).

> Seems you are asking for this very complexity to be added to flymake.el and exposed as a new interface. I don't think that's a great idea when you have more options in the program making use of Flymake's API.

Caching is just the last resort if all other options have been exhausted. This is not the case here - therefore no caching layer is needed.

consult-flymake.el
tomfitzhenry: Fixed

joaotavora:
> Caching is just the last resort if all other options have been exhausted. 

I'm quite confused. It seems you were requesting caching for Flymake, right?

>> it would be neat if flymake could cache the line

> This can be improved to O(mlogm+n).

presumably you mean k * O(mlogm+n), where k is the number of times that "consult" runs that code per flymake iteration right?, well i'd just run the O(m*n) version once and then spend O(m) space in caching the silly line value and forget about it. Next k-1 times it will be O(1) runtime. A weak hash table diag -> line, can do this in 3 lines of code (and no Flymake interfaces).

```elisp
(defvar consult--diag-to-line (make-hash-table :weakness :key))

(defvar consult--diag-line (diag)
   (or (gethash diag consult--diag-to-line) 
         (setf (gethash diag consult--diag-to-line)
             ...existing code that calculates the line...)))
```

manuel-uberti:
[@minad  See? I knew it including João Távora in the conversation would spur interesting comments on this problem. :)]

minad:
@joaotavora

> I'm quite confused. It seems you were requesting caching for Flymake, right?

No, I am not requesting caching from flymake - I request that flymake stores the line number with the diagnostics. This line number is usually provided by the backend, I am just requesting that flymake preserves all the information provided by the backends.

> presumably you mean k * O(mlogm+n), where k is the number of times that "consult" runs that code per flymake iteration right?, 

No.

@manuel-uberti 
> [@minad See? I knew it including João Távora in the conversation would spur interesting comments on this problem. :)]

Generally, I agree that it is a good thing if more people look on some piece of code and add valuable information. Still - we were not at a point, where we had to ping some other busy maintainer for advice just yet. We are all busy, so I think it is better if unnecessary noise is avoided. Nevertheless, @joaotavora seems to be interested in this and does not mind to give advice as far as I can tell from here :)

joaotavora:
> No, I am not requesting caching from flymake - I request that flymake stores the line number with the diagnostic

"storing" is another words for "caching" in my book.  Also Flymake must store markers, and not absolute line numbers, which aren't invalidated if other areas of the buffer are changed. So if Flymake were to keep and give out the backend's information as you request, it would be acting irresponsibly, as it can quickly be invalidated.

> No. Let's assume consult only calls everything just once, k=1.

So is there really a performance/responsiveness problem?  Flymake spends way much more time iterating diagnostics and I don't see it, generally.

> Nevertheless, @joaotavora seems to be interested in this and does not mind to give advice as far as I can tell from here :)

You seem to be requesting stuff from Flymake, I'm telling you there are simpler solutions.  But I'll be on my way.

minad:
> "storing" is another words for "caching" in my book. Also Flymake must store markers, and not absolute line numbers, which aren't invalidated if other areas of the buffer are changed. So if Flymake were to keep and give out the backend's information as you request, it would be acting irresponsibly, as it can quickly be invalidated.

I know that and it is certainly a good reason to not provide line numbers. As I argued, it is also no problem to compute the line number on the fly (also every time without caching). But for this view I think it is acceptable to provide line numbers which are only correct modulo editing. For the actual jumping and navigation we use markers in any case.

> So is there really a performance/responsiveness problem? Flymake spends way much more time iterating diagnostics and I don't see it, generally.

No, there is no responsiveness issue here :)

> You seem to be requesting stuff from Flymake, I'm telling you there are simpler solutions. But I'll be on my way.

Yes, I do request the line numbers to be provided. And I disagree that what you are proposing is simpler. But it is not really important and it is subjective. And I also do not care strongly about it. If I would care strongly, I could still propose a patch to upstream.

joaotavora:
> I know that and it is certainly a good reason to not provide line numbers. As I argued, it is also no problem to 
> compute the line number on the fly (also every time without caching). But for this view I think it is acceptable to 
> provide line numbers which are only correct modulo editing. 

Not to mention certain backends might simply not report things in terms of line numbers.  The Elisp bytecomp backend is a prominent example.

> No, there is no responsiveness issue here :)

That's great.  Then I'd venture to say this is a case of premature optimization.

> Yes, I do request the line numbers to be provided. And I disagree that what you are proposing is simpler

It's simpler to open a new interface in a library for calculating something of dubious validity than doing it in the client using only public interfaces?  You fail to convince me, but open the Flymake issue, maybe you'll convince someone.

consult-flymake.el
tomfitzhenry: Feels like premature optimisation?

consult-flymake.el
tomfitzhenry: `'flymake-error` is a default value, for Flymake diagnostic types that  don't have the property 'mode-line-face, nor are in a category that has a 'mode-line-face. Authors of Flymake diagnostics types should do either of those if they care about styling.

This case is rare, and degrades gracefully: `(propertize "foo" 'face nil)` is a no-op.

minad:
@tomfitzhenry Thank you, I merged it and applied some improvements.

consult-flymake.el
minad: There is a difference between optimization and trivially avoiding duplicate work. Why do we have to argue about everything here? This is not an efficient way of getting things done.

minad: See here on how this should have been written https://github.com/minad/consult/commit/4cd166da289f8bfe5e9bda6b6b7d047b6e16794b

PR #67, opened 2020-12-14
implement consult-help command which allows to search all documentation strings

minad:
too slow to be useful as of now...


minad:
@clemera @oantolin Is this expected that the default refine function is faster than orderless? With @clemera's suggestions I think it is almost usable. There is a large startup time, but I don't see how we can mitigate that. But that's all.

PR #68, opened 2020-12-15
[WIP] Add first implementation of consult-search

ackerleytng:
Seeking comments on code style and implementation as I learn elisp!

Question: How do you develop on consult? Do you symlink this directory into your `~/.emacs.d`?

### DIFF
--- a/consult-search.el
+++ b/consult-search.el
@@ -1,0 +1,92 @@
+;;; consult-search.el --- Providing search via Consult and Projectile  -*- lexical-binding: t; -*-
+
+;; Author: Daniel Mendler, Consult and Selectrum contributors
+;; Maintainer: Daniel Mendler
+;; Created: 2020
+;; License: GPL-3.0-or-later
+;; Version: 0.1
+;; Package-Requires: ((projectile "2.0") (emacs "26.1"))
+;; Homepage: https://github.com/minad/consult
+
+;; This file is not part of GNU Emacs.
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; consult-search integrates command line search tools and presents the results
+;; using completing-read.  This was written to work with consult
+
+;; This only works with ag for now
+
+;;; Code:
+
+(require 'projectile)
+
+(defun consult-search--marked-string ()
+  "Return the marked input (anything highlighted in buffer), nil if nothing is marked."
+  (when (use-region-p)
+    (buffer-substring-no-properties (region-beginning) (region-end))))
+
+(defun consult-search--search-string ()
+  "Get the input for search."
+  (or (consult-search--marked-string)
+      (thing-at-point 'symbol t)))
+
+(defun consult-search--do-ag (s path)
+  "Call ag with search string S on PATH and return all results.  Return a list of results (strings)."
+  (with-temp-buffer
+    (let ((exit-code (process-file "ag" nil t nil "--nocolor" "--nogroup" "--column" s path)))
+      (when (= exit-code 0)
+        (split-string (buffer-string) "\n" t)))))
+
+(defun consult-search--show-search-completions (search-string)
+  "Call ag with SEARCH-STRING on projectile's project root."
+  (completing-read
+   "ag: "
+   (consult-search--do-ag search-string (projectile-project-root))
+   nil t search-string))
+
+(defun consult-search--get-or-read-search-string ()
+  "Either get search string from or ask user for search string.  For use in interactive."
+  (list (or (consult-search--search-string) (read-string "Search for: "))))
+
+(defun consult-search--parse-filename-line-number (string)
+  "Parse filename and line number from STRING, which is assumed to be from ag output.  Input format is expected to be <file path>:<line number>:<column number>:<line from file>.  Return list of <file path>, <line number as int>, <column number as int>."
+  (let ((parts (split-string string ":")))
+    (list (car parts) (string-to-number (cadr parts)) (string-to-number (caddr parts)))))
+
+(defun consult-search--open-user-choice (string)
+  "Open user choice STRING, which is assumed to be from ag output."
+  (let* ((parsed (consult-search--parse-filename-line-number string))
+         (filename (car parsed))
+         (line-number (cadr parsed))
+         (column-number (caddr parsed)))
+    (find-file filename)
+    ;; Go back to the start of the file, then go to line-number
+    (goto-char (point-min)) (beginning-of-line line-number)
+    ;; Go forward to the right column
+    (forward-char (1- column-number)))
+  (let ((parts (split-string string ":")))
+    (list (car parts) (string-to-number (cadr parts)) (string-to-number (caddr parts)))))
+
+(defun consult-ag (search-string)
+  "Search for the marked string, or symbol at point, or if neither of the above are available, a SEARCH-STRING from prompt, then presents the user with search results from ag, and then opens the user's choice in the buffer."
+  (interactive (consult-search--get-or-read-search-string))
+  (let ((user-choice (consult-search--show-search-completions search-string)))
+    (consult-search--open-user-choice user-choice)))
+
+(provide 'consult-search)
+
+;;; consult-search.el ends here


manuel-uberti:
Thank you so much for working on this. I feel like the first steps to a `consult-rg` command are on their way. :)

manuel-uberti:
Instead of Projectile, couldn't you use the built-in `project` to get the project root?

minad:
@ackerleytng I will take a look and make some review soon!

consult-search.el
```diff
@@ -0,0 +1,92 @@
+;;; consult-search.el --- Providing search via Consult and Projectile  -*- lexical-binding: t; -*-
```
minad, line 1: I would like if it could have a less generic name. consult-project-grep or something like this, but no need to change it now - we might come up with something better.

Furthermore I think you can put it directly in consult.el and somehow get rid of the projectile dependency.

consult-search.el
```diff
@@ -0,0 +1,92 @@
...
+;;; Commentary:
+
+;; consult-search integrates command line search tools and presents the results
+;; using completing-read.  This was written to work with consult
+
+;; This only works with ag for now
+
+;;; Code:
+
+(require 'projectile)
```
minad, line 35: It would be great if we can get rid of this dependency and offer instead a consult-project-root-function defcustom such that the user can either configure project.el or projectile.el from outside.

consult-search.el
```diff
@@ -0,0 +1,92 @@
...
+
+(defun consult-search--search-string ()
+  "Get the input for search."
+  (or (consult-search--marked-string)
+      (thing-at-point 'symbol t)))
+
+(defun consult-search--do-ag (s path)
+  "Call ag with search string S on PATH and return all results.  Return a list of results (strings)."
+  (with-temp-buffer
+    (let ((exit-code (process-file "ag" nil t nil "--nocolor" "--nogroup" "--column" s path)))
```
minad, line 50: Can we make this more generic, configurable? Somehow offer grep, rg, ag as possible backends?

consult-search.el
```diff
@@ -0,0 +1,92 @@
...
+      (when (= exit-code 0)
+        (split-string (buffer-string) "\n" t)))))
+
+(defun consult-search--show-search-completions (search-string)
+  "Call ag with SEARCH-STRING on projectile's project root."
+  (completing-read
```
minad, line 56: This should be consult--read since it will allow for additional features like preview and narrowing. I think we might at least add preview support?

consult-search.el
```diff
@@ -0,0 +1,92 @@
...
+   "ag: "
+   (consult-search--do-ag search-string (projectile-project-root))
+   nil t search-string))
+
+(defun consult-search--get-or-read-search-string ()
```
minad, line 61: I like it more if you simply inline very small functions which are used only once. Maybe take a quick look at the consult.el source. I also introduce helper functions to avoid repetition but try to keep the number low.

consult-search.el
```diff
@@ -0,0 +1,92 @@
...
+    (list (car parts) (string-to-number (cadr parts)) (string-to-number (caddr parts)))))
+
+(defun consult-search--open-user-choice (string)
+  "Open user choice STRING, which is assumed to be from ag output."
+  (let* ((parsed (consult-search--parse-filename-line-number string))
+         (filename (car parsed))
+         (line-number (cadr parsed))
+         (column-number (caddr parsed)))
+    (find-file filename)
+    ;; Go back to the start of the file, then go to line-number
```
minad, line 77: No unnecessary comments like this ;)

consult-search.el
```diff
@@ -0,0 +1,92 @@
...
+    (goto-char (point-min)) (beginning-of-line line-number)
```
minad, line 78: Rather put these on separate lines.

minad (commented):
@ackerleytng I added some first comments!

consult-search.el
```diff
@@ -0,0 +1,92 @@
...
+    ;; Go forward to the right column
+    (forward-char (1- column-number)))
+  (let ((parts (split-string string ":")))
+    (list (car parts) (string-to-number (cadr parts)) (string-to-number (caddr parts)))))
+
+(defun consult-ag (search-string)
+  "Search for the marked string, or symbol at point, or if neither of the above are available, a SEARCH-STRING from prompt, then presents the user with search results from ag, and then opens the user's choice in the buffer."
```
manuel-uberti, line 85: It's probably best to format this docstring to have a full sentence on the first line and rest broken down to more sentences.

minad: @manuel-uberti Please add more of your review comments! I would very much appreciate that - in case you have time to help out here!

manuel-uberti: @minad sure! I'll see what I can do. And of course if `rg` comes into place here I'll be glad to try out the code just as I did for `consult-flymake`.

minad: @manuel-uberti Thanks!

minad:
@ackerleytng 

> Seeking comments on code style and implementation as I learn elisp!

I think what you are writing here is already good. I tend to use fewer small functions in elisp since they pollute the namespace/imenu/help-view. And there is not a big benefit in using small functions which are used only once. Furthermore elisp is not as composable as other functional programming languages.

> Question: How do you develop on consult? Do you symlink this directory into your ~/.emacs.d?

I usually load/install the package by hand (either via load or via package-install-file) and then execute the sexps in the editor directly while editing. This allows me to adjust functions, test changes, repeat. From time to time reload everything if it is necessary. This works well for me, but maybe there is a better way to do things.

tomfitzhenry:
Per https://github.com/minad/consult/issues/6#issue-752877729 , consider using xref facilities, xref-search-program, xref-matches-in-files. Emacs already has support for searching using grep or ag. It'd be less code to maintain if that was reused, and the xref to completing-read code could likely be reused for other consult functions in future.

minad:
@tomfitzhenry I also thought about that, I will make a comment in #68. See also https://github.com/travitch/completing-read-xref.el by @travitch. Does it make sense to have this here?

manuel-uberti:
> @tomfitzhenry I also thought about that. See also https://github.com/travitch/completing-read-xref.el by @travitch. Does it make sense to have this here?

FTR, in Emacs 28 one could set `xref-show-definitions-function` to `#'xref--show-defs-minibuffer` to have `completing-read` available.

minad:
@manuel-uberti Thanks! I guess it makes only sense to have the command here if it provides something on top of the standard xref function, e.g., if we have preview or some other advantages.

tomfitzhenry:
> FTR, in Emacs 28 one could set xref-show-definitions-function to #'xref--show-defs-minibuffer to have completing-read available.

You can do this in Emacs 27. It just requires project >=0.5.3 from https://elpa.gnu.org/packages/project.html .

```elisp
(setq xref-search-program 'ripgrep)
(setq xref-show-xrefs-function 'xref--show-defs-minibuffer)
```

Then `(project-find-regexp)` (and typing "defsubst" into the prompt looks like:

image

minad:
@tomfitzhenry what do you suggest? Does it still make sense to provide this here with preview support? The implementation could be taken/inspired by the project.el implementation. Another idea - we could try to modify the preview support such that it can enhance already existing commands. But I am not sure how easy this will be and to which commands it would apply.

tomfitzhenry:
> @tomfitzhenry what do you suggest? Does it still make sense to provide this here with preview support?

I think if consult chooses to implement this, it should only be to add on what isn't provided by xref itself. Perhaps that is preview support (which isn't a strong factor for me). If so, it should do so by implementing xref-show-xrefs-function, and then document how to use this, and possibly provide an interactive function to use it:

```elisp
(defun consult-project-ripgrep ()
  (interactive)
  (let ((xref-search-program 'ripgrep)
        (xref-show-xrefs-function 'consult-show-defs-minibuffer))
    (call-interactively 'project-find-regexp)))
```

I don't think we want to be in the space of customising search programs, or parsing their output.

ackerleytng:
I think @tomfitzhenry 's suggestion pretty much solves it for me in terms of search and filtering, that output is also nicely colored, and also jumps to the column!

Now for one more request - how would I implement something like wgrep-ag where, after showing all the search results in the minibuffer, I can do C-x C-e to bring up another buffer to edit all the results and then commit them back to multiple files? https://github.com/mhayashi1120/Emacs-wgrep

manuel-uberti:
@ackerleytng Do you mean you have an implementation for `consult-xref-show-xrefs-function`?

manuel-uberti:
> Now for one more request - how would I implement something like wgrep-ag where, after showing all the search results in the minibuffer, I can do C-x C-e to bring up another buffer to edit all the results and then commit them back to multiple files? https://github.com/mhayashi1120/Emacs-wgrep

Something similar is already provided I guess: `xref-query-replace-in-results`.

manuel-uberti:
> @ackerleytng Do you mean you have an implementation for `consult-xref-show-xrefs-function`?

Actually, setting `xref-show-xrefs-function` to `#'xref--show-defs-minibuffer` should be enough. @tomfitzhenry what do you think?

ackerleytng:
> @ackerleytng Do you mean you have an implementation for `consult-xref-show-xrefs-function`?

Nope I meant that `xref--show-defs-minibuffer` is good enough.

Also, `xref-query-replace-in-results` doesn't provide the interactivity that `wgrep-ag` does. With `wgrep-ag`, you can filter first, so that the result set is trimmed, then edit only those.

s-kostyaev:
> > @tomfitzhenry what do you suggest? Does it still make sense to provide this here with preview support?
> 
> I think if consult chooses to implement this, it should only be to add on what isn't provided by xref itself. Perhaps that is preview support (which isn't a strong factor for me, but might be for others). If so, it should do so by implementing xref-show-xrefs-function, and then document how to use this, and possibly provide an interactive function to use it:
> 
> ```emacs-lisp
> (defun consult-project-ripgrep ()
>   (interactive)
>   (let ((xref-search-program 'ripgrep)
>         (xref-show-xrefs-function 'consult-xref-show-xrefs-function))
>     (call-interactively 'project-find-regexp)))
> ```
> 
> I don't think we want to be in the space of customising search programs, or parsing their output.

Perfect. Would be nice if consult preview also will works here.

minad:
What is the status here? How should we proceed here? I guess it would still be nice to have a consult variant of `project-find-regexp` with preview. Furthermore we may be able to add the dynamic aspect such that the search already starts after having entered a few keys.

ackerleytng:
I'm trying to get a feel of whether I like the experience of being forced to decide what i want to search for in a separate step from the filtering. If the experience from helm-projectile-ag (search and filter, interleaved) is preferred, then I might want to invest more there. For now seeing the preview of the line itself seems sufficient to me...

minad:
I pushed a first version of `consult-show-xrefs` here - https://github.com/minad/consult/blob/xref/consult-xref.el

manuel-uberti:
Thanks for that. I am using it like this:
```emacs-lisp
(defun consult-project-rg ()
    (interactive)
    (let ((xref-show-xrefs-function #'consult-show-xrefs))
      (call-interactively 'project-find-regexp)))
```
But after hitting <kbd>RET</kbd> at the `Find regexp` to accept the thing at point as the thing to search for, I get this:
```
Debugger entered--Lisp error: (error "xref-bogus-location-message is already defined as ...")
  signal(error ("xref-bogus-location-message is already defined as ..."))
  error("%s is already defined as something else than a gen..." xref-bogus-location-message)
  cl-generic-ensure-function(xref-bogus-location-message)
  cl-generic-define-method(xref-bogus-location-message nil ((this xref-bogus-location)) nil #f(compiled-function (this) "Access the slot `message' from object of class `xref-bogus-location'." #<bytecode 0x8d9a5b983da1077>))
  byte-code("\300\301\302\303\302\304%\210\300\305\302\306\302\307%\210\310\311\312\313!\"\210\310\314\315\313!\"\210\310\316\314\"\210\317\316\320\321#\210\322\313\323\314#\324\313\325\326..." [cl-generic-define-method xref-location-marker nil ((l xref-buffer-location)) #f(compiled-function (l) #<bytecode -0xdab6ade345e11b2>) xref-location-group ((l xref-buffer-location)) #f(compiled-function (l) #<bytecode 0x13149baa216a0d69>) defalias xref-bogus-location-p eieio-make-class-predicate xref-bogus-location xref-bogus-location--eieio-childp eieio-make-child-predicate xref-bogus-location-child-p make-obsolete "use (cl-typep ... \\='xref-bogus-location) instead" "25.1" define-symbol-prop cl-deftype-satisfies eieio-defclass-internal (xref-location) ((message :type string :initarg :message :reader xref-bogus-location-message)) (:documentation "Bogus locations are sometimes useful to\nindicate e...") xref-bogus-location-message ((this xref-bogus-location)) #f(compiled-function (this) "Access the slot `message' from object of class `xref-bogus-location'." #<bytecode 0x8d9a5b983da1077>)] 7)
  require(xref)
  project-find-regexp(#("xref-pop-to-location" 0 20 (fontified t)))
  funcall-interactively(project-find-regexp #("xref-pop-to-location" 0 20 (fontified t)))
  call-interactively(project-find-regexp)
  (let ((xref-show-xrefs-function #'consult-show-xrefs)) (call-interactively 'project-find-regexp))
  consult-project-rg()
  funcall-interactively(consult-project-rg)
  call-interactively(consult-project-rg nil nil)
  command-execute(consult-project-rg)
```

minad:
@manuel-uberti It works for me. I cannot help debugging this right now - it is just a poc, maybe you can figure it out. I am not sure if I want to pursue the route via xref. Xref is only synchronous and to me it feels over-engineered for what it offers. Right now I am looking into the possibility to make the search asynchronous such that grep/rg must not have finished for the candidate selection to get started - see https://github.com/raxod502/selectrum/issues/306. It would still be nice to integrate somehow via xref. What do you think?

manuel-uberti:
Asynchronous search sounds interesting, although for my use cases ripgrep is already fast enough I am happy to let it finish.

About my problem with your solution: I tested it with standard Emacs minibuffer completion, no Selectrum or Icomplete involved.

minad:
> About my problem with your solution: I tested it with standard Emacs minibuffer completion, no Selectrum or Icomplete involved.

Which solution - the asynchronous one? I think this can also work with default completion.

manuel-uberti:
I meant `consult-show-xrefs`.

minad:
See here for an updated version of the asynchronous proposal #91. The downside is that we cannot use anything from the xref infrastructure, but parsing grep/rg output is not hard. We could still generate xref-items if that is desired and in order to allow integration with xref facilities. Alternatively consult-show-xrefs could be offered for people who prefer a synchronous search with preview.

minad:
I am closing this PR in favor of #91. There exist `consult-grep`, `consult-ripgrep` and `consult-git-grep` prototypes which allow to collect the candidates asynchronously. I think it still needs some work until it can be considered "stable", but I think it is on a good way. The asynchronous approach works with selectrum, icomplete, default completion and embark-occur/live-occur. Right now there is no support yet for project roots and no support yet to only search a list of files. It would be nice to hear some feedback regarding the `consult-grep` commands. What are your expectations regarding such a command?

Independent of #91 there exists the xref branch with `consult-show-xrefs` (see https://github.com/minad/consult/blob/xref/consult-xref.el). This could be pursued independently.

Ping @ackerleytng @manuel-uberti @tomfitzhenry @s-kostyaev

PR #69, opened 2020-12-15
fix #57

### LINKED ISSUE #57: Candidate order is changed when filtering in fido-mode
@oantolin When using `consult-line` the candidates are first nicely sorted by line, but as as soon as I start filtering, the order gets messed up. I only see this with icomplete and not with selectrum. Do you also have this and is this expected? I am using the identity as cycle-sort-function and display-sort-function in consult--read as you are probably aware.

oantolin:
Make the no sorting case of consult--read really do no sorting under
fido-mode, or more generally, under the flex completion style, by
temporarily disabling flex's extra sorting by "completion score".

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -530,8 +530,20 @@ NARROW is an alist of narrowing prefix strings and description."
               (lambda (cmd cand state)
                 (funcall preview cmd (and cand (funcall lookup candidates cand)) state)))
        (consult--with-narrow narrow
-         (completing-read prompt candidates-fun
-                          predicate require-match initial history default))))))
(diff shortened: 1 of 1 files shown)


minad:
Tbh I am not a fan of this. I will probably not pull such a workaround. I don't care enough about fido-mode.

oantolin:
Sure, that sounds reasonable. I'll close this then.

oantolin:
For me the fun part was figuring out what the hell was going on, not so much fixing the problem anyway. 😛

minad:
@oantolin Sure, I hoped you would figure it out :)

PR #78, opened 2020-12-18
Handle previews of invisible text

jakanakaevangeli:
Fixes https://github.com/minad/consult/issues/21

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -303,15 +303,15 @@ DISPLAY is the string to display instead of the unique string."
 
 PREVIEW is an expresion which previews the candidate.
 FUN is the body function."
-  (save-excursion
-    (save-restriction
-      (push (lambda (cand) (funcall preview 'preview cand nil)) consult--preview-stack)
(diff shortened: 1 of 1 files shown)


minad:
@jakanakaevangeli Thank you! For now I added the save-excursion fix. I tested this and it works nicely with org mode and outline. However I wonder if it makes sense to generalize this somehow and rather install it via some hooks, maybe also install them by default. The question is if there are other mechanisms which perform hiding/unhiding.

There is consult--jump-1 and the isearch-range-invisible code is always executed after it, so maybe that code could go to the consult-after-jump-hook?

minad:
@jakanakaevangeli See here for the current wip https://github.com/minad/consult/tree/unfolding

minad:
I decided to use the mechanism of ctrlf, which seems cleaner since like this we do not manipulate any global state.

jakanakaevangeli:
Thanks. Congratulations on getting rid of global state.

PR #85, opened 2020-12-22
Run command from current major mode or from an active minor mode

oantolin:
This an alternative implementation to yours. It's very similar. I was already working on it when I saw your email and GitHub comments. I did include a command for minor modes.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -1053,6 +1053,55 @@ The arguments and expected return value are as specified for
         (funcall exit completion exit-status))
       t)))
 
+(defun consult--extract-commands (mode)
+  "Extract commands from MODE."
+  (let ((library-path (symbol-file mode))
+        (mode-rx (regexp-quote
+                  (substring (if (eq major-mode 'c-mode)
+                                 "cc-mode"
+                               (symbol-name major-mode))
+                             0 -5))))
+    (mapcan
+     (lambda (feature)
+       (let ((path (car feature)))
+         (when (or (string= path library-path)
+                   (string-match-p mode-rx (file-name-nondirectory path)))
+           (mapcar #'cdr
+                   (seq-filter
+                    (lambda (item)
+                      (and (consp item)
+                           (eq (car item) 'defun)
+                           (commandp (cdr item))
+                           (not (string-match-p "--" (symbol-name cmd)))))
+                    (cdr feature))))))
+     load-history)))
+
+;;;###autoload
+(defun consult-major-command ()
+  "Run a command from the current major mode."
+  (interactive)
+  (call-interactively
+   (consult--read "M-X: "
+                  (consult--extract-commands major-mode)
+                  :require-match t
+                  :category 'command
+                  :lookup #'intern)))
+
+;;;###autoload
+(defun consult-minor-command ()
+  "Run a command from one of the current active minor modes."
+  (interactive)
+  (call-interactively
+   (consult--read "M-X: "
+                  (mapcan #'consult--extract-commands
+                          (seq-filter (lambda (m)
+                                        (and (boundp m) (symbol-value m)))
+                                      minor-mode-list))
+                  :require-match t
+                  :category 'command
+                  :lookup #'intern)))
+
+
 (defun consult--yank-read ()
   "Open kill ring menu and return selected text."
   (consult--read
--- a/consult.el
+++ b/consult.el
@@ -229,6 +229,7 @@ You may want to add a function which pulses the current line, e.g.,
 (defvar consult--apropos-history nil)
 (defvar consult--theme-history nil)
 (defvar consult--minor-mode-menu-history nil)
+(defvar consult--mode-command-history nil)
 (defvar consult--kmacro-history nil)
 (defvar consult--buffer-history nil)
 (defvar-local consult--imenu-history nil)
@@ -1053,9 +1054,9 @@ The arguments and expected return value are as specified for
         (funcall exit completion exit-status))
       t)))
 
-(defun consult--extract-commands (mode)
-  "Extract commands from MODE."
-  (let ((library-path (symbol-file mode))
+(defun consult--mode-commands (&optional minor-mode)
+  "Extract commands from MINOR-MODE."
+  (let ((library-path (symbol-file (or minor-mode major-mode)))
         (mode-rx (regexp-quote
                   (substring (if (eq major-mode 'c-mode)
                                  "cc-mode"
@@ -1066,41 +1067,46 @@ The arguments and expected return value are as specified for
        (let ((path (car feature)))
          (when (or (string= path library-path)
                    (string-match-p mode-rx (file-name-nondirectory path)))
-           (mapcar #'cdr
-                   (seq-filter
-                    (lambda (item)
-                      (and (consp item)
-                           (eq (car item) 'defun)
-                           (commandp (cdr item))
-                           (not (string-match-p "--" (symbol-name cmd)))))
-                    (cdr feature))))))
+           (mapcar
+            (lambda (x)
+              (cons
+               (consult--narrow-candidate
+                (cond
+                 ((and minor-mode (local-variable-if-set-p minor-mode)) ?l)
+                 (minor-mode ?g)
+                 (t ?m))
+                (symbol-name (cdr x)))
+               (cdr x)))
+            (seq-filter
+             (lambda (cmd)
+               (and (consp cmd)
+                    (eq (car cmd) 'defun)
+                    (commandp (cdr cmd))
+                    (not (string-match-p "--" (symbol-name (cdr cmd))))))
+             (cdr feature))))))
      load-history)))
 
 ;;;###autoload
-(defun consult-major-command ()
-  "Run a command from the current major mode."
+(defun consult-mode-command ()
+  "Run a command from the currently active major and minor modes."
   (interactive)
-  (call-interactively
-   (consult--read "M-X: "
-                  (consult--extract-commands major-mode)
+  (command-execute
+   (consult--read "Mode command: "
+                  (consult--remove-dups
+                   (append
+                    (consult--mode-commands)
+                    (mapcan #'consult--mode-commands
+                            (seq-filter (lambda (m)
+                                          (and (boundp m) (symbol-value m)))
+                                        minor-mode-list)))
+                   #'cdr)
+                  :narrow '((?m . "Major")
+                            (?l . "Local Minor")
+                            (?g . "Global Major"))
                   :require-match t
-                  :category 'command
-                  :lookup #'intern)))
-
-;;;###autoload
-(defun consult-minor-command ()
-  "Run a command from one of the current active minor modes."
-  (interactive)
-  (call-interactively
-   (consult--read "M-X: "
-                  (mapcan #'consult--extract-commands
-                          (seq-filter (lambda (m)
-                                        (and (boundp m) (symbol-value m)))
-                                      minor-mode-list))
-                  :require-match t
-                  :category 'command
-                  :lookup #'intern)))
-
+                  :history 'consult--mode-command-history
+                  :lookup #'consult--lookup-candidate
+                  :category 'command)))
 
 (defun consult--yank-read ()
   "Open kill ring menu and return selected text."
--- a/consult.el
+++ b/consult.el
@@ -1101,8 +1101,7 @@ The arguments and expected return value are as specified for
                                         minor-mode-list)))
                    #'cdr)
                   :narrow '((?m . "Major")
-                            (?l . "Local Minor")
-                            (?g . "Global Major"))
+                            (?l . "Local Minor"))
                   :require-match t
                   :history 'consult--mode-command-history
                   :lookup #'consult--lookup-candidate


minad:
I didn't upload my GH version earlier, I just had it locally - so no worries. I wasn't decided if I should even push it. I will take a look and make some comments in the source.

consult.el
```diff
@@ -1044,6 +1044,55 @@ The arguments and expected return value are as specified for
...
+;;;###autoload
+(defun consult-minor-command ()
+  "Run a command from one of the current active minor modes."
+  (interactive)
+  (call-interactively
+   (consult--read "M-X: "
+                  (mapcan #'consult--extract-commands
+                          (seq-filter (lambda (m)
+                                        (and (boundp m) (symbol-value m)))
+                                      minor-mode-list))
```
minad, line 1090: Here my question is if it makes sense to distinguish local and global modes. We could use narrowing! We could make it one command `consult-mode-command` and then narrow with some keys for minor local/minor global/major. I would probably prefer that over having distinct commands.

consult.el
```diff
@@ -1044,6 +1044,55 @@ The arguments and expected return value are as specified for
         (funcall exit completion exit-status))
       t)))
 
+(defun consult--extract-commands (mode)
+  "Extract commands from MODE."
+  (let ((library-path (symbol-file mode))
+        (mode-rx (regexp-quote
+                  (substring (if (eq major-mode 'c-mode)
+                                 "cc-mode"
```
minad, line 1052: This is better than in the original smex version :)

consult.el
```diff
@@ -1044,6 +1044,55 @@ The arguments and expected return value are as specified for
...
+                               (symbol-name major-mode))
+                             0 -5))))
+    (mapcan
+     (lambda (feature)
+       (let ((path (car feature)))
+         (when (or (string= path library-path)
+                   (string-match-p mode-rx (file-name-nondirectory path)))
```
minad, line 1059: But the file name check is very ad-hoc. I also didn't have a better idea yet :(

Since smex and amx both had this maybe it is a good enough heuristic which has been tested a lot?

consult.el
```diff
@@ -1044,6 +1044,55 @@ The arguments and expected return value are as specified for
...
+                    (cdr feature))))))
+     load-history)))
+
+;;;###autoload
+(defun consult-major-command ()
+  "Run a command from the current major mode."
+  (interactive)
+  (call-interactively
+   (consult--read "M-X: "
+                  (consult--extract-commands major-mode)
```
minad, line 1076: Then the question is if you also want to include the local key bindings. This is what the smex version does, but maybe it doesn't add value?

consult.el
oantolin: A single command with narrowing does sound good.

consult.el
oantolin: Yeah, it is probably is battle-tested. The `(string= path library-path)` makes sense. I wonder what typical cases the `string-match-p` catches.

consult.el
oantolin: I didn't include them because @protesilaos didn't. I would expect all of the commands from key bindings to already be included, but I'm not sure.

consult.el
minad: Ok, let's do this then. I think it is a good idea - mostly people want all commands I gues. But if you want to explore you can narrow.

minad: We could also think about adding another narrowing possibility to only show commands with bindings, this goes in the direction of consult-binding...

consult.el
minad: I am sure you need this if larger modes are distributed over multiple files?

consult.el
minad: Ok, we can quickly check this, by copying over the snippet from my branch. But unfortunately it wouldn't work well together with the narrowing idea since from the keymap the provenance is unclear.

consult.el
oantolin: Yes, that's probably it.

minad:
@oantolin @protesilaos I tried to unify both commands, but now I have the feeling that it makes sense to filter out global minor-mode commands, since these add a lot of clutter. I think it is more useful to have only *local* commands, we could also call this `consult-local-command`. What do you think?

protesilaos:
> I tried to unify both commands, but now I have the feeling that it makes sense to filter out global minor-mode commands, since these add a lot of clutter. 

I think this is right.

> we could also call this `consult-local-command`

I do not have a particular preference on this.  How about making that plural, as in `consult-local-commands`?

minad:
> I do not have a particular preference on this. How about making that plural, as in consult-local-commands?

This is not consistent with the rest of the commands. At some point I made the decision to have them singular - bad or good but this is what we have now

minad:
I found the perfect solution for the global commands. We can hide them first and show them only if "g SPC" is pressed via the new narrowing system. I am also using this to unhide the ephemeral buffers :)

minad:
I pushed it here https://github.com/minad/consult/commit/468bc94087fa064e57c436c332d433ce4c3981fc. Please test.

minad:
@oantolin @protesilaos Do you have a good keybinding proposal for the README?

minad:
I added `C-c m` to the readme for now (https://github.com/minad/consult/commit/a90a28b05d89af3dad3d1ec0115258b23fce6626), I guess this makes sense since `C-c <letter>` are user-defined commands. Furthermore `C-c` are usually local in contrast to `C-x`.

protesilaos:
Thank you for your work on this!  Personally I will use `M-X`. Otherwise I would have chosen `C-c x` or `C-x x`.  Perhaps you want it to have some mnemonic value, in which case `C-x m` could be an option, though that is bound to `compose-mail` by default and I am using it already.

minad:
I also find `C-c m` mnemonic, and since `C-c` contains local commands, I like it better to have it there than under `C-x`. I also considered `C-x m` and other possibilities, but `C-x` is already so crowded. `M-X` is also an option, but I think such uppercase bindings are quite uncommon?

Would be great if you could test it and report if you find anything out of the order - I applied some larger scale refactorings to the narrowing system also in order to make this possible (#86), so things may be broken.

consult.el
minad: I didn't add the binding stuff.

consult.el
minad: I decided to not add the binding stuff. Keeping separate concepts separate.

oantolin:
I also would have suggested `M-X`, because I believe that's what `smex` suggested, but `C-c m` is fine.

protesilaos:
> I also find C-c m mnemonic, and since C-c contains local commands, I like it better to have it there than under C-x.

Correct!

> M-X is also an option, but I think such uppercase bindings are quite uncommon?

I know that they do not work in (some?) terminal emulators, but I am okay with that for my case.

> Would be great if you could test it and report if you find anything out of the order

I just tried it with `compose-mail`.  It seems to be listing global minor modes by default.  Will double-check.

> I applied some larger scale refactorings to the narrowing system also in order to make this possible (#86), so things may be broken.

I have not tried any narrowing yet.  Will need to read more about it and start testing things.

oantolin:
The refactoring to the narrowing system is great! I like that consult-minor-mode-menu no longer has prefixes, I'll add `(minor-mode . embark-symbol-map)` to the default value of `embark-keymap-alist`. I think the standard symbols actions are adequate for minor modes. The obvious actions missing would be to activate/deactivate/toggle the minor mode, but I find `eval-expression` adequate for this: by default it will enter `(my-minor-mode )` into the minibuffer and leave point right before the `)`.

minad:
For this command, maybe we should filter out the commands ending in `-mode`? Or add some more general filter functionality via regexps?

@protesilaos 

> I just tried it with compose-mail. It seems to be listing global minor modes by default. Will double-check.

Please do! For me it seems to work. But maybe some things are wrong, we are using a heuristic here after all.

oantolin:
> I have not tried any narrowing yet.

@protesilaos The narrowing system is a brilliant invention of @minad's. Some consult commands list several "types" of candidates together, each type is assigned a letter, you narrow to a particular type by typing its letter and then `SPC` in an empty minibuffer. You widen again by using `<backspace>` at the beginning of the line. This command can narrow to commands [g]lobal minor modes, [l]ocal minor modes or [m]ajor mode commands. When not narrowed, only the local and major commands are shown.

The recent refactoring @minad mentioned didn't change the user-facing functionality, but did change the implementation. Before this was implemented by modifying the candidates to include a prefix indicating which narrowing type they belong to. Now, in many cases, the narrowing can be done without mangling the candidates, making them work out of the box with Marginalia and with Embark. (Again, love the refactoring, @minad!)

minad:
> @protesilaos The narrowing system is a brilliant invention of @minad's. 

Haha, thanks! But it is not exactly my invention - I took the concept from the buffer switcher from the Selectrum wiki! I only made it work well, hopefully it really does now :)

oantolin:
So, one thing I don't like isn't your fault, @minad, it's something inherited from `smex`'s algorithm: any command defined in the same file as a minor mode is considered as belonging to that minor mode. For example, since I use `consult-preview-mode`, I get all the consult commands listed among the global minor mode commands.

Maybe this isn't really a problem. For one thing, for major modes it's arguably correct: major modes typically define at least some commands they don't bind, how else would you find them other than by looking at which commands are defined in the same file?

For another thing, we do have nice filtering through the completion system, so having too many candidates isn't really a serious problem.

oantolin:
> > @protesilaos The narrowing system is a brilliant invention of @minad's.
> 
> Haha, thanks! But it is not exactly my invention - I took the concept from the buffer switcher from the Selectrum wiki! I only made it work well, hopefully it really does now :)

Oh, the things I miss out on by not using Selectrum. I didn't even know this! Well, your implementation in consult is still brilliant. 😄

minad:
> For example, since I use consult-preview-mode, I get all the consult commands listed among the global minor mode commands.

Yes I know. This is what originally lead me to propose the removal/hiding of global mode commands. Besides that it is shameless consult self-promotion of its own commands. Maybe I should add consult-consult-command too :-P

> For another thing, we do have nice filtering through the completion system, so having too many candidates isn't really a serious problem.

I still added the regexp, see https://github.com/minad/consult/commit/fa7b5dd8ad8081e1b43cd1a51f781f58528722b1, since we already filter out the -- commands, so it does not hurt to have a customizable regexp.

oantolin:
Is there a way to find out what narrowings a command supports without peeking at the source code?

minad:
@oantolin Which-key :) It looks very nice actually, see here for a screenshot https://github.com/oantolin/embark/issues/50. But I think you guys are not using which-key?

oantolin:
Ooh, that is nice. Maybe I should be using which-key.

minad:
Btw, the usual which-key configuration of embark has been broken by your big refactoring - I should find a way to make it work again, since I also rely on which-key for the embark indicators. I wonder how anyone can live without which-key? Do you always hammer the C-h key?

oantolin:
Off-topic: it used to be that some `line` commands used the unicode prefix and some used actual line numbers in the candidates, and now they all use the unicode prefix, right?

minad:
@oantolin Yes, I changed that, I think consult-mark used the line numbers as part of the candidate before, but I changed this. Now it is uniform.

minad:
@oantolin Regarding the narrow indicators - I am open to a PR which adds alternative indicators to the prompt, it should be possible to add some alternative via a simple prompt overlay. I think @clemera mentioned that it is possible to have a multiline overlay before the prompt, did you try sth like this in embark too?

minad:
@oantolin A poor man's solution could be to simply do this: If :narrow is given to `consult--read`, simply apply a replacement to the prompt string, e.g., such that consult-buffer shows `Switch to [bfmv]:`, it is concise but not very pretty. This is what I considered before.

oantolin:
I think I'm happy binding this to a key:

```elisp
(defun show-narrowings ()
  (interactive)
  (message (mapconcat
            (lambda (x)
              (concat (propertize 
                       (single-key-description (car x))
                       'face 'consult-narrow-indicator)
                      " "
                      (cdr x)))
            consult--narrow-prefixes
            " ")))
```

I don't want the narrowings displayed all the time.

minad:
Sound's like a good idea. Maybe add a `consult-narrow-help-key` to allow binding this automatically. Or I define a `consult-narrow-keymap` and the users have to bind the commands themselves. I also don't have the narrow and widening key bound by default, since I have the no-keybinding policy in consult ;)

protesilaos:
> @protesilaos The narrowing system is a brilliant invention of @minad's. Some consult commands list several "types" of candidates together, each type is assigned a letter, you narrow to a particular type by typing its letter and then SPC in an empty minibuffer. You widen again by using <backspace> at the beginning of the line. This command can narrow to commands [g]lobal minor modes, [l]ocal minor modes or [m]ajor mode commands. When not narrowed, only the local and major commands are shown.

Sounds good!  Does not work on my end though.  Maybe there is something wrong with my config…  I have SPC to self-insert.

oantolin:
You know, the one thing I don't like about the narrowing system is actually pretty easy to fix. The system is great if you know at the beginning of the line that you intend to narrow. But I often realize after typing a little that I meant to narrow, so I kill the line, do the narrowing, and yank the line again. I think I'd prefer instead if the beginning of the line received no special treatment (so no fancy rebinding of `SPC`) and instead some command set the narrowing, something like:

```elisp
(defun consult-prompt-narrow (char)
  "Narrow to type CHAR, or widen if CHAR is RET."
  (interactive
   (list
    (read-char (concat (propertize "RET" 'face 'consult-narrow-indicator)
                       " widen "
                       (mapconcat
                        (lambda (x)
                          (concat (propertize 
                                   (single-key-description (car x))
                                   'face 'consult-narrow-indicator)
                                  " "
                                  (cdr x)))
                        consult--narrow-prefixes
                        " ")))))
  (message "CHAR %s" char)
  (consult--narrow-set (if (= char ?\C-m) nil char)))
```

That still needs to refresh the completion system, but you get the idea.

minad:
@oantolin This already exists if you set the consult-narrow-key and the consult-widen-key.

oantolin:
Oh, sorry! Does that also remove the special processing at the start of the line?

minad:
@oantolin No, the special processing is still there. Do you want an option to disable that?

minad:
I guess I should just use a keymap then the users can bind all the things themselves. If you don't bind SPC/DEL you don't get the special treatment, if you don't bind widen/narrow you won't get that. I will make a nicer design than what I am having right now.

oantolin:
> @oantolin No, the special processing is still there. Do you want an option to disable that?

No, not really. It doesn't get in the way. I was just asking.

> I guess I should just use a keymap then the users can bind all the things themselves. If you don't bind SPC/DEL you don't get the special treatment, if you don't bind widen/narrow you won't get that. I will make a nicer design than what I am having right now.

I actually the think the current design is OK. I was just ignorant about it.

minad:
> I actually the think the current design is OK. I was just ignorant about it.

Ok, but I will try nevertheless just to see if it could be better.

oantolin:
> Sounds good! Does not work on my end though. Maybe there is something wrong with my config… I have SPC to self-insert.

This is odd, @protesilaos, I think all you do is bind `SPC` to `nil` in `minibuffer-local-completion-map`. That's what I do as well, and the consult narrowing works for me.

minad:
@oantolin I added the simple help command https://github.com/minad/consult/commit/c106f3b23d2ef3aff7c827f045b4971f7b768364

minad:
@protesilaos Did you figure it out in the meantime, otherwise I would suggest we move that discussion to a new issue?

protesilaos:
> Did you figure it out in the meantime, otherwise I would suggest we move that discussion to a new issue?

It does not work in my `emacs -Q` test.  What I did:

1. `(load-file "/path/to/consult.el")`
2. `(require 'consult)`
3. `(define-key minibuffer-local-completion-map (kbd "SPC") nil)`
4. `M-x consult-mode-command SPC`.
5. Then whatever key I press, I get "no completions".  I assume this means that the key itself is not interpreted specially by consult so, per the default completion styles, nothing starts with that character and, hence, nothing is available to complete against (I guess?).

But maybe you have made some commit in the last few hours/minutes and I did not get that, so let me retry as soon as I can.

That granted, please move on and I will contact you when I have to.

minad:
@protesilaos Ok, can you please create an issue for this and I will look into it soon. Otherwise it might get lost.

oantolin:
@protesilaos, I don't think `M-x consult-mode-command SPC` is supposed to do anything. Try `M-x consult-mode-command 
m SPC` to narrow to major mode commands.

protesilaos:
> Try M-x consult-mode-command m SPC to narrow to major mode commands.

Ah yes, this works!  Sorry for the noise.

PR #91, opened 2020-12-25
Async support

minad:
I made a first toy implementation for async `consult--read`. It is pretty simple to use.

~~~ elisp
;;; -*- lexical-binding: t -*-

(let* ((rest)
       (proc (make-process :name "consult-async"
                           :command '("sh" "/home/user/elisp-snippets/test.sh")))
       (async (consult--async-make :cleanup (lambda () (ignore-errors (kill-process proc))))))
  (set-process-filter proc (lambda (_ out)
                             (let ((lines (split-string out "\n")))
                               (if (cdr lines)
                                   (progn
                                     (setcar lines (concat rest (car lines)))
                                     (setq rest (car (last lines)))
                                     (consult--async-push async (nbutlast lines)))
                                 (setq rest (concat rest (car lines)))))))
  (set-process-sentinel proc (lambda (_ event)
                               (when (and rest (string= event "finished\n"))
                                 (consult--async-push async (list rest) 'done))))
  (consult--read
   "Test: " async
   :sort nil))
~~~

There are todos regarding the UI refreshing, where help would be greatly appreciated...

* consult-selectrum--refresh is an ugly hack and should be written properly, furthermore check that scrolling works correctly
* consult--icomplete-refresh should be rewritten such that the scrolling works

See also https://github.com/raxod502/selectrum/issues/306

cc @oantolin @clemera


oantolin:
So did we settle on a meaning for the phrase "such that the scrolling works"?

minad:
@oantolin 

> So did we settle on a meaning for the phrase "such that the scrolling works"?

No, we did not. But what exists now for icomplete-vertical satisfies the definition of "not working". The selectrum refresher seems to work, but I need @clemera's and your input. Furthermore, the selectrum refresher should be written in a clean way.

minad:
@oantolin mentioned three possibilities for candidate selection after an update in https://github.com/raxod502/selectrum/issues/306

1. Select the first candidate matching the minibuffer input, this works automatically. This won't allow scrolling through multiple candidates which all match the input :-1:
2. Selecting the candidate with the index of the previously selected candidate in the old list. I would not like the candidate to magically change without the user pressing a key :-1:
3. Select the *same candidate* as before, works with scrolling, the candidate does not magically change :+1:

I would like option 3 if we can make it work!

minad:
I refactored the async support - async is a function now, this allows to build nice pipelines with candidate transformers.

~~~ elisp
(defun consult--async-example ()
  (thread-first (consult--async-sink)
    (consult--async-transform (lambda (x) (concat "<" x ">")))
    (consult--async-transform (lambda (x) (concat "#" x "#")))
    (consult--async-timer)
    (consult--async-indicator)
    (consult--async-process '("sh" "/home/user/elisp-snippets/test.sh"))))

(consult--read
 "Test: " (consult--async-example)
 :sort nil)
~~~

consult-selectrum.el
```diff
@@ -45,11 +45,7 @@
 (defun consult-selectrum--refresh ()
   "Refresh selectrum view."
   (when (eq completing-read-function #'selectrum-completing-read)
-    ;; TODO async refresh hack!
-    (let ((old-index selectrum--current-candidate-index))
-      (selectrum-exhibit)
-      (setq selectrum--current-candidate-index old-index)
-      (selectrum--minibuffer-post-command-hook))))
+    (selectrum-exhibit 'keep-selected)))
```
oantolin, line 48: I'll take a stab at writing the icomplete version of this. It will have to be slightly different because icomplete does not always have a candidate selected. Before you cycle among the candidates, none of the them is really selected. For example pressing `RET` in that circumstance does not enter the top candidate (unless it is the only one). But if you run `icomplete-forward-completions` and then immediately `icomplete-backward-completions` everything looks the same as if you hadn't cycled, the minibuffer input doesn't change, but now if you press `RET` you do get the top candidate.

Given this, I believe the correct behavior for icomplete when updating the candidate list is:

- if no candidate was selected yet, stick to no candidate being selected
- if a candidate was selected and it happens to still exist after the update, select it again
- if a candidate was selected but it disappears during the update, go to the no candidate selected state.

oantolin: Does that sound reasonable?

minad: This sounds great! Thank you so much for taking a stab at this! I am perfectly happy with having different behaviors for different backends, since this should match the behavior expected by the user of the particular completion system!
@clemera [already wrote the version for Selectrum](https://github.com/raxod502/selectrum/pull/307), which achieves option 3 from above! It worked well in my tests. I am not yet 100% happy with the patch, but I am sure we will see this being merged very soon!

minad:
@oantolin @clemera I am not sure if you already took a look at how consult-grep works, but right now I am first reading the regexp and then run grep. Then the candidates arrive asynchronously. Now the question if it would also be possible to combine the two steps, such that grep is automatically started on the go as you type. This could also be useful for commands like consult-search which queries a search engine like counsel-search. It is questionable if this is really useful and it is questionable if this is possible after all since when doing this we effectively work around the filtering used by completing-read. Would it make sense to replace the completion style in this case with something which simply matches everything? Would that work? Alternatively one could use a more sophisticated matching style, which separates two parts by semicolon, the part before the semicolon is used for the query and the part after the semicolon is passed to the real completion style, e.g. provided by orderless.

consult-selectrum.el
oantolin: Mmh. I'm wrong about the behavior of icomplete. It's not really that sometimes no candidate is selected. I would say the following is more accurate:

- The first candidate is always the "selected one", in the sense that `icomplete-force-complete-and-exit` selects that one (whether or not you have been cycling).
- The default way to exit icomplete is `icomplete-ret`, which does choose the first candidate if you have not entered any text into the minibuffer (more precisely if the part between the completion boundaries is empty, so for example if in find-file you navigate to some directory, the input is considered "empty" if you haven't type anything for the last path component yet), but will not choose the first candidate if you have entered some text into the minibuffer.

Given that corrected account I believe the correct behavior for refreshing icomplete is to search for the previous first candidate in the new candidate list and if it does appear, rotate the new candidate list to make that the first candidate, and if it does not appear, do not rotate the new candidate list.

oantolin: Oops, I ran into trouble @minad. I'm trying to test my icomplete refresh function and I wanted to use a very slowly updating async source, so I wrote this little script in a file called `slowsource` with I made executable and put on the `$PATH`:

```perl
#! /usr/bin/perl

for (1 .. 10000) {
    print int(rand(1000000)), "\n";
    sleep 1;
}
```

and this function to use it:

```elisp
(defun test-async ()
  (interactive)
  (consult--read
   "Number: "
   (thread-first (consult--async-sink)
     (consult--async-refresh)
     (consult--async-indicator)
     (consult--async-process '("slowsource")))))
```

But I think I must have misunderstood the API because I don't see any candidates appearing. What am I doing wrong?

minad: There is nothing wrong with your usage of the api. What is missing is the flush in your perl. In case there is another error, there is also the async stderr ephemeral buffer.

minad: For testing it may be easier to use timer based source? The async is just a function and you can push candidates from anywhere.

~~~
(funcall async '(list of candidates))
~~~

oantolin: Thank you! Classic trap, the buffering. Sorry to bother you for something silly.

minad: No worries! Thank you again for looking into this - I wouldn't be able to do this quickly without your help, since I am not familiar with the icomplete internals. Btw, how would this thing work in combination with embark-live-occur? If I am entering something, live-occur updates as does the standard emacs TAB completion. But can I also automatically refresh the live-occur somehow?

oantolin: For Embark live occur its easier since there is no notion of the selected candidate, so this suffices:

```elisp
(add-hook 'consult--completion-refresh-hook #'embark-occur--update-linked)
```

oantolin: OK, done! I made PR against the async branch.

minad: Thanks!

minad:
@oantolin @clemera The plan is to merge this into main soon. The async infrastructure is not very intrusive and works well with the rest. But I will wait until selectrum-exhibit acquires the keep candidate support. The commits will be squashed and split into two - one for the basic async infrastructure and one with the grep prototype. The grep prototype will be reworked after it hit the main branch, possibly after gathering feedback from users.

oantolin:
Mmh, isn't this "async-part; filter-part" a generalization of narrowing? Except that the narrowing selector, the "async-part" is longer than one key and is not limited to some list determined by the command. But it should behave similarly: the async-part determines which candidates the filter-part matches against, just like in narrowing.

minad:
> Mmh, isn't this "async-part; filter-part" a generalization of narrowing? Except that the narrowing selector, the "async-part" is longer than one key and is not limited to some list determined by the command. But it should behave similarly: the async-part determines which candidates the filter-part matches against, just like in narrowing.

Yes one can see the async part as a generalization of narrowing. However the narrowing solution is more low-tech, so maybe it makes sense to keep the narrowing part as is. The narrowing works via predicates which are always available on top of the way candidates are generated. But one could obviously implement narrowing also via async and not use predicates at all. Would you prefer that?

I like about the current approach that narrowing, preview and async are separate enhancements, which could be easily factored out. People may want to use narrowing but not async. I think this makes the consult--read implementation somehow robust and easy to modify/extend.

Regarding the particular issue with having this kind of single-step input for both filtering/async generation, I am not sure about the best way to implement this. One could do something like narrowing - extend the keymap (or use a transient map) and at first accumulate input which is used for async generation until a separator key is pressed. From then on the input is not intercepted such that the following input is used for completion filtering. But I am also looking for a solution which is as simple as possible while being reasonably general.

minad:
@oantolin It might also make sense to look at the experimental consult-websearch implementation - there I am **completely** disabling filtering by using the input as prefix (This is similar to how narrowing works). Having filtering effectively disabled is not really good and it would be nicer to have the "async; filter" separation though. But in the case of `consult-grep`, I think it would play a bigger role since you could start typing and searching right away.

oantolin:
I think having a clear "async; filter" separation would be very good. A key difference between these functions in consult and similar ones in counsel or helm, could be that the consult functions have the filter part matched by whatever completion styles you have configured. I haven't tried `consult-websearch` yet, but I have to say eliminating the filtering by prefixing with the input does not sound very appealing. I tried `consult-grep` and it seems pretty usable as is, maybe the only thing missing is to change the grep arguments without restarting the command.

Maybe the following is a sensible design?

Like `consult-grep`, all the async commands can first prompt for search terms, then do filtering on the results. There could be a command `consult-change-search-terms` or something, that the user can bind to a key if they want to change the search terms while filtering. For people that don't bind it, the commands are still useful.

oantolin:
Having a command to change the async-part seems less magical than giving the semi-colon special treatment and getting completing read to just filter on the stuff after the semicolon. The async part could be put into the minibuffer prompt, so people can see it while filtering.

minad:
> I think having a clear "async; filter" separation would be very good. A key difference between these functions in consult and similar ones in counsel or helm, could be that the consult functions have the filter part matched by whatever completion styles you have configured.

Agree.

>  I haven't tried consult-websearch yet, but I have to say eliminating the filtering by prefixing with the input does not sound very appealing.

Yes, it is ugly. But this is just a prove of concept since I wanted to test async generation of candidates triggered by keypresses. We should look for a better solution though.

> Like consult-grep, all the async commands can first prompt for search terms, then do filtering on the results. There could be a command consult-change-search-terms or something, that the user can bind to a key if they want to change the search terms while filtering. For people that don't bind it, the commands are still useful.

Okay, so you basically have a key to toggle between filter/async parts. Does not sound like a bad idea, pretty simple to implement, but I would prefer a more direct solution. If no alternative solution is possible, I can still go with what you are proposing - two-step input plus filter/async toggle.

minad:
> Having a command to change the async-part seems less magical than giving the semi-colon special treatment and getting completing read to just filter on the stuff after the semicolon. The async part could be put into the minibuffer prompt, so people can see it while filtering.

The semicolon idea is similar to narrowing by using letter+space. But narrowing offers also the variant with the consult-narrow-key, so this corresponds to the additional key. 

There is one other thing to consider though - if we have the two-step process, there is no way to immediately show the candidates while typing the initial regexp. This is only be possible if we unify the two steps into a single step, where part of the input is used for async and part of the input is used for filtering. That's the main reason why I am looking for something like this.

oantolin:
A key to toggle between aync/filter parts is a nice way to put it. I think that in the approach were you have a single line of input in the format "async; filter", the semicolon is basically a less convenient way to toggle between the parts: if you want to change something in the async part you have to move to the async part with some motion commands, it's more convenient to have a key to toggle between editing one and the other.

> There is one other thing to consider though - if we have the two-step process, there is no way to immediately show the candidates while typing the initial regexp.

I see. That is a disadvantage.

minad:
> 

A key to toggle between aync/filter parts is a nice way to put it. I think that in the approach were you have a single line of input in the format "async; filter", the semicolon is basically a less convenient way to toggle between the parts: if you want to change something in the async part you have to move to the async part with some motion commands, it's more convenient to have a key to toggle between editing one and the other.

Yes, I agree. It is just the same with the narrowing. But you should also consider that moving and editing is something people do all the time. It might not be the most efficient way to achieve something but it is very familiar and it can get you started right away - therefore I like it. Having a separate toggle key is certainly more efficient.

minad:
I just remembered one thing from the original consult-buffer discussion. For consult-buffer I originally had an implementation based on selectrum-read with narrowing and one based on completing-read without narrowing, back then. The key difference was that selectrum-read allowed to change the input string, which was then used for narrowing, and this is exactly what is needed here too. Basically intercepting the input string before it reaches the filtering framework. If there is a way to achieve this, we are done. This is even an orthogonal concept to the rest - a transformation function for the input string. As if I would advise minibuffer-contents...

oantolin:
> This is even an orthogonal concept to the rest - a transformation function for the input string

Huh, maybe that is the way to go! Although I think it is trickier than just advising minibuffer-contents, since I think that maybe completion uses `buffer-substring-no-properties` in the minibuffer. I don't recall exactly.

minad:
> Huh, maybe that is the way to go! Although I think it is trickier than just advising minibuffer-contents, since I think that maybe completion uses buffer-substring-no-properties in the minibuffer. I don't recall exactly.

I quickly looked into this - but it seems near impossible since there is no point where we can intercept as you say. The only possible interception is before the chars hit the minibuffer-contents, this means overwriting the keymap and this means it will be complicated. Still looking for a better solution...

minad:
Another weird idea - I could replace all characters in the minibuffer by whitespace (or any other irrelevant character) in a post-command-hook. Each character would be propertized with a display of the actual character. :see_no_evil:

minad:
@oantolin I thought a bit more about this async in a single step. I think there are only two viable options.

1. Tweak the completion-style locally, e.g. orderless has this style dispatcher. It works if I split at ; and throw away the first part. The first question is if it would be too intrusive to locally tweak the completion style. The second question is if it is possible to locally define derived completion styles, such that the local style is the current global style modified such that it removes the first part. By doing that we would not be dependent on the active completion style. Even if this is possible there is another complication since selectrum does not use completion styles directly, but has their own filtering framework. So unfortunately the selectrum functions would have to be overwritten too. Another thing to consider - it is not very pretty to introduce something like a style dispatchers/completion-style on the level of consult. This somehow violates the current separation of concerns where consult does not mess with the completion styles.
2. Give up on the idea and simply use the two-step process with some additional keybindings to toggle between filter/async-part, basically what you suggested. Downside is that no async candidate generation can be triggered when entering the async-part.

oantolin:
Well, as you might expect, I favor 2. But maybe there is a way to get live preview of the full result set while editing the async part?

minad:
I don't think this live preview is possible if I want to use the completion system for this. I could create my own window, sure - but that would neither look good nor would it be consistent with the completion system display. 

I am a bit divided now. I would like to keep things simple. This means not to do this, but then I feel sth is missing from the async.

There is some other thing I tried - basically filtering and rewriting the input string before it reaches the completion system. The async-part can then be shown as overlay, similar to narrowing. But then all the editing capabilities are lost. If you make a typo you have to restart. Alternatively I would have to special case all editing commands which would be crazy.

Rewriting the completion style seems to be the technologically cleanest solution. Why do you dislike it? I tried it with the orderless style dispatcher and it requires only five lines of code. But here the solution would not be quite as simple if I would only strip of the async-part and dispatch then to the original style. Maybe this is not even feasible. I don't know enough about completion styles tbh.

oantolin:
Maybe you're right about locally wrapping the completion style. I know how to do that, it wouldn't be hard. It bothers me a little that selectrum requires a separate implementation, but that often happens with selectrum anyway ---and ever since you put in `consult-selectrum.el` it's even "out of sight, out of mind" for me. 😄 

By the way, that reminds me: does jumping to the beginning or the end of the match in `consult-line` really work in Selectrum?

minad:
One can also take the point of view that editing async+preview is simply not completing read but sth else and therefore two steps are justified. But I think I would be once again willing to commit a type error in the name of a slick ui. This websearch is not particularly useful, but as a ui test it is kind of neat. 

For me one one no-go is that I would have to special case selectrum.

minad:
Yes beginning/end jump really works, but I overwrite both the selectrum refine function and the completion style with orderless.

Right now the special casing for selectrum is not bad. But adding this would be a major increase in the amount of needed special casing.

But there is also  https://github.com/raxod502/selectrum/issues/82

minad:
Take a look: https://github.com/minad/consult/blob/main/consult-selectrum.el

It really got better recently. Everything works the same like icomplete basically.

oantolin:
> Yes beginning/end jump really works, but I overwrite both the selectrum refine function and the completion style with orderless.

I wouldn't call that working. I mean, people who use, say, selectrum+prescient, but leave the completion styles at the default value, wouldn't say `consult-line` reliably jumps to the beginning of the match. Maybe there should be a `consult-filter-matches` defcustom with default value something like `(lambda (string candidates) (completion-all-completions string candidates nil (length string)))`, and the `consult-selectrum` package could set that variable to `selectrum-refine-candidates-function`.

minad:
@oantolin I think I am going to propose that selectrum deprecated the refine and highlight function and redirect instead to the completion styles functions.

Is it trivially possible to trivially wrap all-completions, try-completion such that they adhere to the selectrum api refine/highlight api? From what you wrote, it seems yes, but I did not check yet.

minad:
Like this?

~~~ elisp
(setq selectrum-refine-candidates-function
      (lambda (string candidates)
        (let ((result (completion-all-completions string candidates nil (length string))))
          (let ((last result))
            (while last
              (unless (listp (cdr last))
                (setcdr last nil))
              (setq last (cdr last))))
          result)))

(setq selectrum-highlight-candidates-function (lambda (_ candidates) candidates))
~~~

minad:
@oantolin I wrote a prototype which overwrites the completion-styles locally, such that the split is performed. See the functions `consult--async-split` in https://github.com/minad/consult/pull/91/commits/6dadf5ae42327f30c42704743dfd01a7cd9d8914. It is a bit more code than I had hoped, but still very simple. The good thing is that everything is encapsulated by this `consult--async-split` async function, so it is really non-intrusive from my perspective.

oantolin:
> Is it really that easy? If it is I don't see any justification for having custom selectrum-refine-candidates-function and selectrum-highlight-candidates-function. I must be missing something obvious. I wonder about the highlighting step, is highlighting applied to all candidates in that case? This is clearly not without an unnecessary overhead.

I think it is that easy. Probably the main reason for having the refine and highlight functions was that raxod simply didn't know about (even the existence, probably) of completion styles at the time Selectrum was written. A more current reason is probably that prescient would need to be rewritten, but maybe that's not too tough. It would be nice if prescient became a completion-style.

A slightly more idiomatic way to write the `selectrum-refine-candidates-function` is to use the built-in function `last`. The function is traditional in Lisp but oddly named: it does not give you the *element* of a list (and indeed does not assume the argument is a proper list), but the last cons in the list.

```elisp
(setq selectrum-refine-candidates-function
      (lambda (string candidates)
        (let* ((result (completion-all-completions string candidates nil (length string)))
               (last (last result)))
          (setcdr last nil)
          result)))
```

minad:
> A slightly more idiomatic way to write the selectrum-refine-candidates-function is to use the built-in function last. The function is traditional in Lisp but oddly named: it does not give you the element of a list (and indeed does not assume the argument is a proper list), but the last cons in the list.

I thought about using that, but then I wasn't sure if it would work with the weird list :)

gagbo:
Just chiming in (still trying to read and understand the whole conversation) to say that I'll be happy to test that with selectrum once it gets merged. I might try to understand a little more selectrum.el interface to see if I can help on the selectrum update function

minad:
@gagbo Thank you for your interest! You can try out this branch and test the functions consult-grep and consult-websearch. These commands are generating the candidate lists asynchronously. The question is if we want to have two steps for the filtering, e.g. in the case of consult-grep. There you first have to enter the regexp, and after that the grep search is performed. Then you can enter some input to filter the results.

oantolin:
> I wonder about the highlighting step, is highlighting applied to all candidates in that case? This is clearly not without an unnecessary overhead.

Completion styles don't know how many completions will be displayed by the completion UI so they have to highlight all completions just in case. :(

Does Selectrum only highlight the candidates displayed?

minad:
@oantolin 

> Completion styles don't know how many completions will be displayed by the completion UI so they have to highlight all completions just in case. :(
> Does Selectrum only highlight the candidates displayed?

Yes, I think it does. Therefore the Selectrum api is arguably better. But probably when the original Emacs API was designed, nobody thought yet about these ivy style completions.

minad:
@oantolin What do you think about the async split prototype now? See https://github.com/minad/consult/commit/1b441b9cbb2b9cde66c8fc0f7eeb2c76bcc0b210. Is this an acceptable approach from your perspective?

oantolin:
> Yes, I think it does. Therefore the Selectrum api is arguably better.

Agreed.

> But probably when the original Emacs API was designed, nobody thought yet about these ivy style completions.

Yes, I think that's the reason.

> What do you think about the async split prototype now? See 1b441b9. Is this an acceptable approach from your perspective?

Yes, it looks great! I'd write 59 as `?;`, but that's just because I haven't memorized ASCII. :P

oantolin:
So where does the part before the semicolon get used?

oantolin:
Also, will it be possible to ship the selectrum bits off to consult-selectrum?

minad:
> Yes, it looks great! I'd write 59 as ?;, but that's just because I haven't memorized ASCII. :P

Sure, but this messes with my highlighting. Not in your setup??

> So where does the part before the semicolon get used?

By the async function sitting below consult--async-split, in the websearch case by consult--async-json. In the grep case, I am not yet using this facility, but I will try this soon, I guess.

> Also, will it be possible to ship the selectrum bits off to consult-selectrum?

Sure. I just have to find a way to implement this nicely. Maybe simply provide two different consult--async-split functions, one for default and one for selectrum.

minad:
@oantolin I just tested this consult-grep refactoring. It is amazing! So much better than having two steps :partying_face:

oantolin:
I'll test as soon as I can.

minad:
@oantolin I moved out the selectrum specific system stuff to consult-selectrum from `consult--async-input-split`. Can you please take a look if `consult--async-input-split` is written in the "optimal form", regarding the default completion styles?

minad:
Merged into main

oantolin:
Sorry I couldn't get a chance to try them before merging, but I have tried them now and they work great. Since I have a pretty elaborate orderless setup, I really like the clear separation between what is passed as an argument to grep and what is used for filtering. And having both parts in the minibuffer input is really nice, you were right about that, probably better than the UI I had in mind.

You asked at some point for me to look over the `completion-styles` stuff. It looks great, very straightforward, with a single higher order wrapping function. When I last looked at it you were saving the orginal completion styles in a separate variable, I like the new version that just put the new style at the beginning and uses `(cdr completion-styles)` for the original styles better.

I have a couple of suggestions:

- Have these command report some special category. I'm not sure what to call the category, it is a fairly standard format, maybe `compiler-error` or `grep-result`. Or something more generic like `file-line`. I think I most want a specific category to be able to show these in Embark Occur list view by default, but I also want to write a `wgrep` exporter (see oantolin/embark#47).

- More documentation. Is the only way to know you use a comma to separate the grep arguments from the filtering part to read the source code? 😄

Great work on the whole async thing! And overriding the completion-styles was a very clever idea.

minad:
@oantolin Thank you for looking at it! I felt like it is ready, therefore I already merged it into main - also in order to avoid some synchronization with other kind of independent changes going on in main. And since the async part is relatively separate from the rest, it is also a rather low risk change, which is great! Having this configurable pipeline turned out to be a powerful design :)

> Sorry I couldn't get a chance to try them before merging, but I have tried them now and they work great. Since I have a pretty elaborate orderless setup, I really like the clear separation between what is passed as an argument to grep and what is used for filtering. And having both parts in the minibuffer input is really nice, you were right about that, probably better than the UI I had in mind.

Great if it works for you! Getting full access to the completion-style power and full access to the grep regexp power is the reason for the clear separation. And having them separated only by the comma still makes editing easy. What is maybe missing is the possibility to configure this separation character and the possibility to escape it, in case you want to use a comma inside the grep-regexp. But until now, I didn't have a need for this yet.

> You asked at some point for me to look over the completion-styles stuff. It looks great, very straightforward, with a single higher order wrapping function. When I last looked at it you were saving the orginal completion styles in a separate variable, I like the new version that just put the new style at the beginning and uses (cdr completion-styles) for the original styles better.

Thank you for cross-checking! I also like it better to minimize the amount of additional global state and it is reasonably concise like this!

> More documentation. Is the only way to know you use a comma to separate the grep arguments from the filtering part to read the source code? smile

I have this documented in the README - but it would probably make sense to expand this. For now I marked the commands as **EXPERIMENTAL** since I would like to gather some feedback and there could still be some bigger changes.

> Have these command report some special category. I'm not sure what to call the category, it is a fairly standard format, maybe compiler-error or grep-result. Or something more generic like file-line. I think I most want a specific category to be able to show these in Embark Occur list view by default, but I also want to write a wgrep exporter (see oantolin/embark#47).

I also thought about it, but I also expected somehow that you will remind me again of it! :) Having a wgrep exporter is an amazing idea. Both 'file-line and 'grep would make fine categories. I would go with the more special 'grep for now, we could still generalize it if the need comes up.

> Great work on the whole async thing! And overriding the completion-styles was a very clever idea.

Thanks! Also thank you very much for the discussion here and the feedback!

oantolin:
>  And having them separated only by the comma still makes editing easy.

Yes, I think I was underestimating how seemless jumping to a spot and editing feels to an Emacs user. Yes it takes more keystrokes than some command to toggle which part you edit, but you don't feel it since you do it all the time.

> What is maybe missing is the possibility to configure this separation character and the possibility to escape it, in case you want to use a comma inside the grep-regexp. But until now, I didn't have a need for this yet.

Having it configurable might be nice, but probably not necessary. If it were configurable I might change mine to `|` so it feel like I'm piping grep to orderless. :)

Escaping on the other hand feel a little more important. Somebody will ask for it at some point, I bet.
 
> I have this documented in the README - but it would probably make sense to expand this. For now I marked the commands as **EXPERIMENTAL** since I would like to gather some feedback and there could still be some bigger changes.

Oh, I hadn't noticed the README. I think the main documentation are really the docstrings: people installing from MELPA don't have the README! Of course they can go to GitHub and see it, the point is that the docstrings **are** installed.
 
> I also thought about it, but I also expected somehow that you will remind me again of it! :) Having a wgrep exporter is an amazing idea. Both 'file-line and 'grep would make fine categories. I would go with the more special 'grep for now, we could still generalize it if the need comes up.

`grep` is good for now. I'll get to work on the wgrep exporter. Step 1: install wgrep and figure out how to use it. (Somehow I never felt I needed it, even though I use `occur-edit-mode` *all the time*.)

minad:
> Oh, I hadn't noticed the README. I think the main documentation are really the docstrings: people installing from MELPA don't have the README! Of course they can go to GitHub and see it, the point is that the docstrings are installed.

Good point! I made #99 to track this. The docstrings as of now are not really good.

> grep is good for now. I'll get to work on the wgrep exporter. Step 1: install wgrep and figure out how to use it. (Somehow I never felt I needed it, even though I use occur-edit-mode all the time.)

:+1: Looking forward to it! Then I will also install wgrep :)

oantolin:
The wgrep export was supereasy! It's actually just grep export, and then you can use wgrep on the resulting grep mode buffer in the normal way.

oantolin:
> Piping is a neat analogy. But if you don't feel strongly about it, [...]

Oh, I definitely do not feel strongly about it. The comma is fine.

travitch:
Thanks - this async grep implementation was exactly what I was looking for! I had a horrible hacky synchronous grep that worked but was kind of miserable to use.

gagbo:
I think I’m just missing a way to pass rg command line arguments, but I guess that’s just something to handle in the processing of the `async-part`.

For example, I might want to pass a flag to search with/without case-sensitivity, in hidden files, or only searching in files matching a certain type/extension.

I’ll try to see how I can make it work when it comes up again in my daily usage

minad:
@travitch Good to hear! Your completing-read xref implementation was something we did also consider for inclusion - but I think the asynchronous grep is a better solution.

@gagbo I think we only have to do minimal adjustments to make this possible, I will create a separate issue to continue the discussion there.

travitch:
@minad I agree that this is much better for general searching. For reference, the only reason that I was using xref at all was because it is the suggested interface to the dumb-jump package (https://github.com/jacktasia/dumb-jump/). That package basically just uses grep/ripgrep, except it has dozens of language-specific rules for finding the definition sites of identifiers. I've found that to be immensely more powerful than grep for some use cases (though I still use plain grep for other things). I don't think that including anything xref-specific in consult is critical. I might give it a shot (building on your example) because the preview support might be useful in actually figuring out which candidate dumb-jump reported is the definition you want to jump to.

minad:
@travitch I looked into xref and didn't find it overwhelming. A lot of clos boilerplate but not very useful for my purposes here. Furthermore it seemed synchronous only or did I miss something there? 

As of now consult reports xref-location completion categories which is helpful for some integrations, e.g. Embark. But I am not using anything from the xref infrastructure. In Emacs 28 there will be a show xref function similar to your completing-read xref implementation, so I probably won't add this here.

I am not using dumb-jump - I found https://github.com/jacktasia/dumb-jump/pull/334 which adds an asynchronous backend to dumb-jump, but I am not sure how this is works. Maybe it would be interesting to implement a consult-dumb-jump which asynchronously performs the dumb jump search. But maybe this is not possible or would require heavy restructuring in dumb jump?

PR #92, opened 2020-12-27
consult--icomplete-refresh: keep current candidate selected

oantolin:


consult.el
```diff
@@ -1910,10 +1910,25 @@ OPEN is the function to open new files."
...
 (declare-function icomplete-exhibit "icomplete")
 (defun consult--icomplete-refresh ()
-  "Refresh icomplete view."
-  ;; TODO ensure that scrolling works with async
+  "Refresh icomplete view, keep current candidate selected if possible."
   (when icomplete-mode
-    (setq completion-all-sorted-completions nil)
+    (let ((top (car completion-all-sorted-completions)))
+      (completion--flush-all-sorted-completions)
+      (unless (null top)
```
oantolin, line 1917: This would be better as `when top` of course. 😆

Issue #26, opened 2020-12-02
consult-line should jump to the position of the match on the line


minad:
related #7

oantolin:
I don't know a good generic way to figure out where a completion style found a match. Since completion-styles can basically do whatever they want, and are, for example, under no obligation to actually match anything, this is probably impossible to do in general.

minad:
I had hoped for you chiming in. I have no idea how these matching styles work, so if you could describe it a bit or point me to some resources it would be good. But for now this is very low priority, maybe let's have this discussion at some other point.

There is also the idea of having something like `consult-match` which should behave like `counsel-isearch` and starts generating the candidates base on the matches etc. But this depends on me figuring out how all the dynamic candidates work and it probably also depends on the matching style as in this issue.

veronikazaglotova:
More than that, it also has to move to the end of the position on the line. It's the default behavior for isearch (which I use currenly for moving in my file. Nothing but isearch does that) and swiper (which I didn't use, but I watched the 10 minute example video and it goes to the end of the candidate indeed), and I believe that it's the behavior that most people would expect. I really wish I could use avy or ctrlf or consult-line, but to me they're unusable when they work like that.
Isearch also goes to the beginning of the candidate if you press C-r (I find it very useful for keyboard macros), but that's probably not possible with swiper and consult-line.
For more argumentation, see abo-abo/avy#316

minad:
@veronikazaglotova Right now consult-line jumps to the beginning of the line. You want to have it jump to the end of the line or the end of the match? But as of now it looks unlikely we can implement the feature in a generic way, since it depends on the completion style being used and these completion-styles can do whatever they want as @oantolin said. Therefore we will probably keep the behavior of jumping to the beginning of the line.

veronikazaglotova:
@minad I want neither. I want it to go **after** the match. If I have the following:
```
(use-pakcage something
```
and want to fix the typo, I want to search for "use-pak", then cursor moves to letter "c", I type C-t, and it fixes the typo.

Here's how it looks like when using isearch: https://streamable.com/rsjldc

minad:
@veronikazaglotova As I said, it is technically not possible as of now. But for such local jumps I rather recommend avy. I use consult-line differently, e.g., if I want to edit something in the context of multiple occurrences of some string, not for local jumps.

veronikazaglotova:
>it is technically not possible as of now.
[...]

PR #93, opened 2020-12-27
Allow customizing consult-line to go to start or end of match

oantolin:
This adds a variable `consult-goto-start-of-match` controlling whether `consult-line` leaves you at the beginning or at the end of the match, as discussed in issue #26.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -98,6 +98,12 @@ You may want to add a function which pulses the current line, e.g.,
 `xref-pulse-momentarily'."
   :type 'hook)
 
+(defcustom consult-line-point-placement 'start-of-match
+  "Where to leave point after `consult-line' jumps to a match."
+  :type '(choice (const :tag "Start of the line" start-of-line)
+                 (const :tag "Start of the match" start-of-match)
+                 (const :tag "End of the match" end-of-match)))
+
 (defcustom consult-line-numbers-widen t
   "Show absolute line numbers when narrowing is active."
   :type 'boolean)
@@ -912,7 +918,8 @@ INPUT is the input string entered by the user.
 CANDIDATES is the line candidates alist.
 CAND is the currently selected candidate."
   (when-let (pos (cdr (assoc cand candidates)))
-    (if (string-blank-p input)
+    (if (or (string-blank-p input)
+            (eq consult-line-point-placement 'start-of-line))
         pos
       ;; Strip unique line number prefix
       (while (and (> (length cand) 0)
@@ -929,14 +936,17 @@ CAND is the currently selected candidate."
               (setq end (- end step)))
             (setq step (/ step 2))))
         ;; Find match start position, remove characters from line beginning until matching fails
-        (let ((step 16))
-          (while (> step 0)
-            (while (and (< (+ start step) end)
-                        (completion-all-completions input (list (substring cand (+ start step) end)) nil 0))
-              (setq start (+ start step)))
-            (setq step (/ step 2))))
+        (when (eq consult-line-point-placement 'start-of-match)
+          (let ((step 16))
+            (while (> step 0)
+              (while (and (< (+ start step) end)
+                          (completion-all-completions input (list (substring cand (+ start step) end)) nil 0))
+                (setq start (+ start step)))
+              (setq step (/ step 2)))))
         ;; Marker can be dead
-        (ignore-errors (+ pos start))))))
+        (ignore-errors (+ pos (pcase consult-line-point-placement
+                                ('start-of-match start)
+                                ('end-of-match end))))))))
 
 ;;;###autoload
 (defun consult-line (&optional initial)


minad:
Is there a customization option for isearch btw? I haven't found one. I found this post, https://endlessparentheses.com/leave-the-cursor-at-start-of-match-after-isearch.html, but this is not good since the cursor only goes to the beginning when leaving the search.

oantolin:
I believe there is no boolean variable for isearch. There is an isearch-mode-end-hook, though. I set that hook to the following function:

```elisp
(defun isearch-exit-at-start ()
  "Exit search at the beginning of the current match."
  (unless (or isearch-mode-end-hook-quit
              (bound-and-true-p isearch-suspended)
              (not isearch-forward)
              (not isearch-other-end))
    (goto-char isearch-other-end)))
```

I also bind this command to `<S-return>` in `isearch-mode-map`:

```elisp
(defun isearch-exit-at-end ()
  "Exit search at the end of the current match."
  (interactive)
  (let ((isearch-other-end (point)))
    (isearch-exit))
  (unless isearch-forward (goto-char isearch-other-end)))
```

oantolin:
Oh, sorry, I hadn't followed the link. It must be where I got the idea in the first place!

oantolin:
I use a slightly different condition though, I wonder why I use that and not Sylvain Rousseau's condition... I can't remember.

minad:
@oantolin Ok, I think there is no way to implement this isearch thing differently, since during the search the cursor must be at the end of the match in order to be able to step to the next match. The isearch-mode-hook seems to be the best one can do.

Regarding this issue - Now that we are introducing a configuration option here, I consider over-engineering it - I would like to offer three options: 'start, 'start-match, 'end-match. If setting 'start, the line matching will be disabled completely.

minad:
> Regarding this issue - Now that we are introducing a configuration option here, I consider over-engineering it - I would like to offer three options: 'start, 'start-match, 'end-match. If setting 'start, the line matching will be disabled completely.

I quickly write the patch for this.

oantolin:
I'd suggest being more explicit in the names of the options: `start-of-line`, `start-of-match`, `end-of-match`. Maybe toss in `end-of-line` for symmetry?

oantolin:
Also the variable should be renamed with more options. Maybe `consult-line-point-placement`?

minad:
> I'd suggest being more explicit in the names of the options: start-of-line, start-of-match, end-of-match. Maybe toss in end-of-line for symmetry?

Yes, also thought about end-of-line. I would like it for symmetry, but there are two reasons why I omitted it: 1. It is useless, 2. the other commands all jump to the beginning of the line, e.g. consult-imenu. I don't want to adjust them such that they also offer the end-of-line setting. So basically consult-line is only augmented by start-of-match/end-of-match and otherwise consult-line can be made to behave like consult-imenu and friends.

oantolin:
Ah, OK, that makes sense, no `end-of-line`, then.

oantolin:
I know you offered to do it, but if you'd like I can make the changes.

oantolin:
What should the default be, though? Maybe `start-of-line` to match `consult-imenu` and the others?

minad:
Ok, please go forward with it! start-of-match should be the default, since it is better. Still having the start-of-line option is sufficient if people want to make the behavior of the commands consistent.

oantolin:
So I have heard of commit squashing but never done it before? Do you want me to make these a single commit?

minad:
Yes squashing is a generally a good idea for single features!

consult.el
```diff
@@ -927,22 +928,26 @@ CAND is the currently selected candidate."
       (let ((start 0)
```
minad, line 928: The start-of-line check should be performed right in the beginning, just return pos directly.

consult.el
```diff
@@ -927,22 +928,26 @@ CAND is the currently selected candidate."
...
             (end (length cand)))
         ;; Find match end position, remove characters from line end until matching fails
-        (let ((step 16))
-          (while (> step 0)
-            (while (and (> (- end step) 0)
-                        (completion-all-completions input (list (substring cand 0 (- end step))) nil 0))
-              (setq end (- end step)))
-            (setq step (/ step 2))))
+        (unless (eq consult-line-point-placement 'start-of-line)
```
minad, line 931: The end computation must be done in every case, since end is used later.

oantolin: I currently do the end computation for `start-of-match` and for `end-of-match`. That should be enough, no? The third case, `start-of-line` doesn't need it.

consult.el
oantolin: Sure, I guess it could be done that way. I'll change it.

consult.el
minad: Right, I misread. Sorry.

oantolin: Now that the start-of-line case is handled earlier, I removed this check, but I still think it was correct before. 😉

oantolin:
OK, I made the changes you suggested and figured out how to squash into a single commit. (I think.)

minad:
Thanks, looks good. I slightly adjusted the names such that they match the standard Emacs terminology.

Issue #103, opened 2020-12-30
Preliminary versions of consult-{find, fd, locate}


inigoserna:
I've improved and somehow generalized the code and added the requested `consult-find` function (but with an issue I'll comment later).

``` emacs-lisp
(defcustom consult-find-min-input 3
  "Minimum number of letters which must be entered, before find/fd/locate... is called."
  :type 'integer)


(defun consult--find-async (cmd)
  "Async function for `consult-find' and similar functions.

  CMD is the locate argument list."
  (thread-first (consult--async-sink)
    (consult--async-refresh-timer)
    (consult--async-process cmd)
    (consult--async-throttle)
    (consult--async-split)))


(defvar consult--find-cmd '("find"))

(defun consult-find (&optional arg)
  "Search for REGEXP with find.
  If ARG then use HOME as start path, else use `default-directory`"
  (interactive)
  (let ((initialpath (expand-file-name (if arg "~" default-directory))))
    (consult--read
     "Find: "
     (consult--find-async (lambda (input)
                            (when (>= (length input) consult-fd-min-input)
                              (append consult--find-cmd (list initialpath "-iname" input)))))
     :sort nil
     :require-match t
     :category 'file
     :history 'file-name-history)))


(defvar consult--fd-cmd '("fd" "-c" "never" "-H"))

(defun consult-fd (&optional arg)
  "Search for REGEXP with fd.
  If ARG then use HOME as start path, else use `default-directory`"
  (interactive)
  (let ((initialpath (expand-file-name (if arg "~" default-directory))))
    (consult--read
     "FD: "
     (consult--find-async (lambda (input)
                            (when (>= (length input) consult-fd-min-input)
                              (append consult--fd-cmd (list input initialpath)))))
     :sort nil
     :require-match t
     :category 'file
     :history 'file-name-history)))


(defvar consult--locate-cmd '("locate" "-iebr"))

(defun consult-locate ()
  "Search for REGEXP with locate."
  (interactive)
  (consult--read
   "Locate: "
   (consult--find-async (lambda (input)
                          (when (>= (length input) consult-locate-min-input)
                            (append consult--locate-cmd (list input)))))
   :sort nil
   :require-match t
   :category 'file
   :history 'file-name-history))
```

`consult-fd` and `consult-locate` are similar to the ones I showed before.
I have an issue with `consult-find` that I'll comment later.

[...]

PR #106, opened 2020-12-30
Added consult-find, consult-fd, consult-locate

inigoserna:
Added consult-find, consult-fd, consult-locate as discussed in https://github.com/minad/consult/issues/103.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -39,7 +39,7 @@
 ;; This package took inspiration from Counsel by Oleh Krehel. Some of the
 ;; commands found in this package originated in the Selectrum wiki. See the
 ;; README for a full list of contributors.
-v2
+
 ;;; Code:
 
 (eval-when-compile
--- a/consult.el
+++ b/consult.el
@@ -2149,5 +2149,87 @@ if not available use `default-directory'."
 
 (add-hook 'consult--completion-refresh-hook #'consult--icomplete-refresh)
 
+
+;;;; find, fd, locate
+(defun consult--guess-initial-dir (&optional askdir)
+  "Returns root directory for use in `consult' searching functions.
+If ASKDIR is not `nil' prompt the user for the inital directory,
+else use simple logic: first try `project', then `projectile',
+and defaults to `default-directory'."
+  (expand-file-name
+   (if askdir
+       (read-directory-name "Initial directory: ")
+     (cond
+      ((and (fboundp 'project-current) (project-current))
+       (project-root (project-current)))
+      ((and (fboundp 'projectile-project-root) (projectile-project-root))
+       (projectile-project-root))
+      (default-directory default-directory)
+      (t "~")))))
+
+
+(defun consult--find-async (cmd)
+  "Async function for `consult-find' and similar functions.
+CMD is the locate argument list."
+  (thread-first (consult--async-sink)
+    (consult--async-refresh-timer)
+    (consult--async-process cmd)
+    (consult--async-throttle)
+    (consult--async-split)))
+
+
+(defvar consult--find-cmd '("find"))
+(defvar consult--fd-cmd '("fd" "-c" "never" "-H" "-p"))
+(defvar consult--locate-cmd '("locate" "-iebr"))
+
+
+;;;###autoload
+(defun consult-find (&optional arg)
+  "Search for REGEXP with find.
+If ASKUSER is not `nil', prompt the user, else use simple logic."
+  (interactive "P")
+  (find-file
+   (let ((initialpath (consult--guess-initial-dir askuser)))
+     (consult--read
+      (concat "Find in " (propertize initialpath 'face 'consult-file) ": ")
+      (consult--find-async (lambda (input)
+                             (append consult--find-cmd (list initialpath "-ipath" input))))
+      :sort nil
+      :require-match t
+      :initial consult-async-default-split
+      :category 'file
+      :history 'file-name-history))))
+
+;;;###autoload
+(defun consult-fd (&optional askuser)
+  "Search for REGEXP with fd.
+If ASKUSER is not `nil', prompt the user, else use simple logic."
+  (interactive "P")
+  (find-file
+   (let ((initialpath (consult--guess-initial-dir askuser)))
+     (consult--read
+      (concat "FD in " (propertize initialpath 'face 'consult-file) ": ")
+      (consult--find-async (lambda (input)
+                             (append consult--fd-cmd (list input initialpath))))
+      :sort nil
+      :require-match t
+      :category 'file
+      :history 'file-name-history))))
+
+;;;###autoload
+(defun consult-locate ()
+  "Search for REGEXP with locate."
+  (interactive)
+  (find-file
+   (consult--read
+    "Locate: "
+    (consult--find-async (lambda (input)
+                           (append consult--locate-cmd (list input))))
+    :sort nil
+    :require-match t
+    :category 'file
+    :history 'file-name-history)))
+
+
 (provide 'consult)
 ;;; consult.el ends here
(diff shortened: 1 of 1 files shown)


consult.el
```diff
@@ -2149,5 +2149,87 @@ if not available use `default-directory'."
 
 (add-hook 'consult--completion-refresh-hook #'consult--icomplete-refresh)
 
+
+;;;; find, fd, locate
+(defun consult--guess-initial-dir (&optional askdir)
```
minad, line 2154: Instead of this function you should `(funcall consult-grep-directory-function)`. We should rename `consult-grep-directory-function` to something more generic though.

consult.el
```diff
@@ -2149,5 +2149,87 @@ if not available use `default-directory'."
...
+  "Returns root directory for use in `consult' searching functions.
+If ASKDIR is not `nil' prompt the user for the inital directory,
+else use simple logic: first try `project', then `projectile',
+and defaults to `default-directory'."
+  (expand-file-name
+   (if askdir
+       (read-directory-name "Initial directory: ")
```
minad, line 2161: And the askdir functionality should go into `consult-grep-directory-default`.

consult.el
```diff
@@ -2149,5 +2149,87 @@ if not available use `default-directory'."
...
+
+;;;###autoload
+(defun consult-find (&optional arg)
+  "Search for REGEXP with find.
+If ASKUSER is not `nil', prompt the user, else use simple logic."
+  (interactive "P")
+  (find-file
+   (let ((initialpath (consult--guess-initial-dir askuser)))
+     (consult--read
+      (concat "Find in " (propertize initialpath 'face 'consult-file) ": ")
```
minad, line 2194: Maybe use an extra face for this - consult-directory-indicator?

consult.el
```diff
@@ -2149,5 +2149,87 @@ if not available use `default-directory'."
...
+  (find-file
+   (let ((initialpath (consult--guess-initial-dir askuser)))
+     (consult--read
+      (concat "FD in " (propertize initialpath 'face 'consult-file) ": ")
+      (consult--find-async (lambda (input)
+                             (append consult--fd-cmd (list input initialpath))))
+      :sort nil
+      :require-match t
+      :category 'file
+      :history 'file-name-history))))
```
minad, line 2217: I guess it would make sense to extract (find-file (consult--read ...)) into a separate function `consult--find`. This function could then be used by `consult-find`, `consult-fd`, `consult-locate`.

minad:
@inigoserna I refactored consult-grep in preparation of this PR, see https://github.com/minad/consult/commit/fef1fc8522281d17d79c4605e83e4c8ee6aea471. We can use `consult--with-directory` for `consult-fd` etc.

consult.el
inigoserna: IMO it provides useful information for the search that it's important to be highlighted.
But I don't think it deserves a new face.

consult.el
inigoserna: Uhmmm... I hadn't noticed this function before.
Yes, I think we should unify `consult-grep-directory-default` and `consult--guess-initial-dir` under a generic name. Please choose one.
Doesn't my implementation be enough?

consult.el
inigoserna: Shouldn't be better if the function mentioned previously include the asking part?

consult.el
minad: I had something only for grep before, but I made it more generic in https://github.com/minad/consult/commit/fef1fc8522281d17d79c4605e83e4c8ee6aea471.

consult.el
minad: We could also refactor this consult-directory-default and consult--with-directory and move all the logic into consult-directory-default if you think that's better.

consult.el
inigoserna: I've tried it, but it was not possible, as each function needs different arguments and `input` is passed in different order in the lambda.

``` emacs-lisp
(append consult--find-cmd (list initialpath "-ipath" input))
(append consult--fd-cmd (list input initialpath))
(append consult--locate-cmd (list input))
```

minad: Make it a higher order function?

inigoserna: Sorry, this is above my skills. I'm not a programmer.
Anyway, wouldn't it be too convoluted for just these functions?

inigoserna: Hey! I think what you mean!

This seems to work...

``` emacs-lisp
(defun consult--find (prompt func)
  "Search for REGEXP with locate."
  (find-file
   (consult--read
    prompt
    (consult--find-async func)
    :sort nil
    :require-match t
    :category 'file
    :history 'file-name-history)))

(defun consult-locate ()
  "Search for REGEXP with locate."
  (interactive)
  (consult--find "Locate: "
                 (lambda (input) (append consult--locate-cmd (list input)))))
```

minad:
@inigoserna I merged it with a few changes. Please test it. We should still discuss the default settings, I am not 100% happy with those. Maybe it makes sense to have different default initial inputs.

* `consult-grep`: `/`
* `consult-fd`: `#`
* `consult-find`: `#*/*`

Maybe it would really make sense to wrap the find string by asterisks?

inigoserna:
> @inigoserna I merged it with a few changes. Please test it. We should still discuss the default settings, I am not 100% happy with those. 

It works in the tests I've run.

> Maybe it makes sense to have different default initial inputs.
>     * `consult-grep`: `/`
>     * `consult-fd`: `#`
>     * `consult-find`: `#*/*`

I don't like it, sorry. 

> Maybe it would really make sense to wrap the find string by asterisks?

No. The deviation from running from CLI could be confusing.


Some comments:
- You have removed the logic to detect the root of a project/projectile when running from a subdir. IMO it is very useful.
- I miss the beautiful colourized path in the prompt ;-)
- Also, I think we should differentiate the prompt between find and fd commands. The user could be testing the results from one or from the other and should know which one is running.

minad:
> You have removed the logic to detect the root of a project/projectile when running from a subdir. IMO it is very useful.

No, I haven't :) there is consult-project-root-function.

> Also, I think we should differentiate the prompt between find and fd commands. The user could be testing the results from one or from the other and should know which one is running

Ok, we should do that! I originally thought fd and find are more similar.

inigoserna:
> > You have removed the logic to detect the root of a project/projectile when running from a subdir. IMO it is very useful.
> 
> No, I haven't :) there is consult-project-root-function.

But this function is undefined by default, isn't it?
In my opinion this functionality should be present without customization.

What do you think about defining commands as a string - it is easier to understand and customize -, so the code only need a `split-string`:

``` emacs-lisp
(defvar consult--locate-cmd "locate --ignore-case --existing --regexp")
;; so we do this before using the command:  (split-string consult--locate-cmd) -> ("locate" "--ignore-case" "--existing" "--regexp")
```

minad:
> But this function is undefined by default, isn't it?
In my opinion this functionality should be present without customization.

No, this ist mit possible unfortunately without introducing dependencies to other packages like projectile.

> What do you think about defining commands as a string - it is easier to understand and customize -, so the code only need a split-string:

This is not a good idea since it will introduce shell quotation problems.

inigoserna:
> > But this function is undefined by default, isn't it?
> > In my opinion this functionality should be present without customization.
> 
> No, this ist mit possible unfortunately without introducing dependencies to other packages like projectile.

Oh, I understand.


> > What do you think about defining commands as a string - it is easier to understand and customize -, so the code only need a split-string:
> 
> This is not a good idea since it will introduce shell quotation problems.

It should not. The commands definition does not need to container quotes at all, only command name and usual arguments.
In case an user with a special need to add f.e. a path with spaces for a path exclusion he could use single quotes anyway.
Am I missing something?

minad:
We should not change this.

* It is potentially less safe because of quoting. Even big it is no problem now it does not mean it will stay like this. The split-string you are proposing is really not a good solution.
* The make-process api expects a list of arguments
* There is no real advantage in having a string. Marginally more convenient at best.

PR #111, opened 2021-01-02
Allow some defcustoms to have nil as a valid value

### LINKED ISSUE #110: Some defcustoms have type mismatches
`consult-project-root-function` is defined with `:type 'function` which does not allow the default value of `nil`.  This results in a `(mismatch)` in the Custom buffer.

Similarly for `consult-view-open-function` and `consult-view-list-function`, and also `consult-{narrow,widen}-key` which are `:type 'vector`.

It looks like these can be fixed using `choice`, e.g.

```elisp
(defcustom consult-project-root-function nil
  "Function which returns project root, used by `consult-buffer' and `consult-grep'."
  :type '(choice function (const nil)))
```

aspiers:
Some defcustoms are defaulting to `nil` without this being an allowed value.  So add `nil` as a type choice.

Fixes #110.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -64,24 +64,24 @@
 
 (defcustom consult-narrow-key nil
   "Prefix key for narrowing during completion."
-  :type 'vector)
+  :type '(choice vector (const nil)))
 
 (defcustom consult-widen-key nil
   "Key used for widening during completion.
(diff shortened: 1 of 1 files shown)


minad:
Thanks!

Issue #104, opened 2020-12-30
Add a changelog


minad:
At some point we need a changelog in order to make it easier to follow the project.

Issue #99, opened 2020-12-29
Improve command docstrings


minad:
We should make more use of the Emacs self documentation capabilities. This means improving the docstrings of every command. Right now most commands have rather short descriptions, missing some details. For example `consult-grep` as mentioned in #91 by @oantolin.

minad:
I made some improvements. There could still be more, but the situation is a little bit better now. We also have the info manual.

PR #112, opened 2021-01-02
Explain how to integrate consult-*grep with projectile

aspiers:
This saves people from having to search for the relevant projectile function themselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,9 +111,11 @@ Most provided commands follow the naming scheme `consult-<thing>`.
     using an advanced completion style like orderless. `consult-grep` supports
     preview. If `consult-project-root-function` is configured and the function
     returns non-nil, `consult-grep` searches the current project directory.
-    Otherwise the `default-directory` is searched. If `consult-grep` is invoked
-    with prefix argument, you can specify the directory manually, i.e.,
-    `C-u M-g r`.
+    Otherwise the `default-directory` is searched.  For example, if you use
+    [projectile](https://github.com/bbatsov/projectile/), set
+    `consult-project-root-function` to `projectile-project-root`. If
+    `consult-grep` is invoked with prefix argument, you can specify the
+    directory manually, i.e., `C-u M-g r`.
   * `consult-find`, `consult-fdfind`, `consult-locate`: Find file by matching
     the path against a regexp. Like `consult-grep` either the project root or
     the current directory is used as root directory for the search. The input


aspiers:
BTW in https://github.com/aspiers/emacs/commit/3133f77058f9714fd0a938f4a02101b08fd90c41 you can see a nice example of how to use my [`with-packages`](https://github.com/aspiers/emacs/blob/master/.emacs.d/lib/with-packages.el) macro to configure this in a reliable way only when both `consult` and `projectile` are loaded.  At some point I'll get around to publishing that as a proper MELPA package.

minad:
I always wondered about this - what the best way is for such integration configuration. Could you add your macros to use-package instead? Or propose an alternative there? I would prefer that over an extra package.

aspiers:
I already tried, and so far failed: https://github.com/jwiegley/use-package/issues/71

minad:
> I already tried, and so far failed: jwiegley/use-package#71

Okay, that's a lengthy discussion already.

> This saves people from having to search for the relevant projectile function themselves.

I already mention projectile in the package configuration section, see https://github.com/minad/consult#package-configuration. I prefer not to overload the description of the commands. Note that the project-root-function is also used by consult-buffer! Maybe the configuration options should just be described better?

If you have suggestions on how to improve the readme I would greatly appreciate it. See also #99. Documentation is generally a bit lacking as of now.

minad:
There's another thing, I had forgotten. I thought about creating an info file from the readme if that is possible. This way the manual could be installed in Emacs and at the same time be shown here, which might make it more useful and accessible.

aspiers:
[**@minad**](https://github.com/minad) commented on [January 2, 2021 2:06 PM](https://github.com/minad/consult/pull/112#issuecomment-753477728 "2021-01-02T14:06:03Z - Replied by Github Reply Comments"):
> > This saves people from having to search for the relevant projectile function themselves.
> 
> I already mention projectile in the package configuration section, see [minad/consult#package-configuration](https://github.com/minad/consult#package-configuration). I prefer not to overload the description of the commands.

Ah nice, I missed that!  Probably I had copied, pasted, and modified that large `use-package` chunk into my own emacs init files before it got updated to include that.  How about adding a link to that section instead of duplicating the info, with text something like this?

    For example, if you use projectile, see the example use-package declaration below.

> Note that the project-root-function is also used by consult-buffer! Maybe it would be better to improve the configuration option descriptions instead?

Yes I like that idea!

> If you have suggestions on how to improve the readme I would greatly appreciate it. See also [#99](https://github.com/minad/consult/issues/99). Documentation is generally a bit lacking as of now.

Sure, will do if I think of anything :-)

minad:
> Ah nice, I missed that! Probably I had copied, pasted, and modified that large use-package chunk into my own emacs init files before it got updated to include that. How about adding a link to that section instead of duplicating the info, with text something like this?

See also #104 ;)

No, I think it would be better to generally the structure of the readme, such that it is obvious to the reader where the crucial information is. Right now there is no other way to read everything and to occasionally check the proposed example configuration to see if something has been added.

aspiers:
[**@minad**](https://github.com/minad) commented on [January 2, 2021 2:08 PM](https://github.com/minad/consult/pull/112#issuecomment-753477933 "2021-01-02T14:08:10Z - Replied by Github Reply Comments"):
> There's another thing, I had forgotten. I thought about creating an info file from the readme if that is possible. This way the manual could be installed in Emacs and at the same time be shown here, which might make it more useful and accessible.

`pandoc` can convert from Markdown to texinfo, and then `makeinfo` from texinfo to info.  I did a quick test and it roughly worked.  I wonder if having the source as `README.org` might yield better results as it's a more flexible format, but I'm not sure.

minad:
Yes, I would then go to org. There is an org to info command afaik.

PR #119, opened 2021-01-03
make git grep/grep/rg commands and regexps customizable

omar-polo:
Hello,

I saw the announcement on r/emacs about consul-grep and I wanted to give it a shot since it seems awesome.

I'm not able to use neither consult-git-grep, consult-grep nor consult-ripgrep.  For consul-grep the answer is easy: I'm on OpenBSD and grep here doesn't accept all the arguments that gnu grep does.  In most cases it has the single letter option (i.e. `-n` instead of `--line-number`, `-E` instead of `--extended-regexp` etc), but there are some flags that aren't supported (in particular `-Z/--null`).

consul-git-grep fails because it uses `/usr/bin/grep` under the hood? probably.  The most puzzling is why ripgrep doesn't work.

(Sorry for not being too specific, "doesn't work" doesn't really mean anything, but I don't know what it's not working.  When I type something at the prompt the various consult-git-grep/grep/ripgrep a red exclamation mark appears right after "Grep" in the prompt and that's all).

This is a first step though, I'm proposing to make the various `-command` part of the public API and thus customizables, as well as the regexps, so folks like me can tune 'em for their specific `grep` implementations.

Cheers!

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -2170,23 +2170,25 @@ used. See also `consult-imenu'."
   (interactive)
   (consult--imenu (consult--project-imenu-items)))
 
-(defconst consult--grep-regexp "\\([^\0\n]+\\)\0\\([^:\0]+\\)[:\0]"
-  "Regexp used to match file and line of grep output.")
+(defcustom consult-grep-regexp "\\([^\0\n]+\\)\0\\([^:\0]+\\)[:\0]"
+  "Regexp used to match file and line of grep output."
+  :type 'regexp)
 
-(defconst consult--grep-match-regexp "\e\\[[0-9;]+m\\(.*?\\)\e\\[[0-9;]*m"
-  "Regexp used to find matches in grep output.")
+(defcustom consult-grep-match-regexp "\e\\[[0-9;]+m\\(.*?\\)\e\\[[0-9;]*m"
+  "Regexp used to find matches in grep output."
+  :type 'regexp)
 
 (defun consult--grep-matches (lines)
   "Find grep match for REGEXP in LINES."
   (let ((candidates))
     (save-match-data
       (dolist (str lines)
-        (when (string-match consult--grep-regexp str)
+        (when (string-match consult-grep-regexp str)
           (let* ((file (expand-file-name (consult--strip-ansi-escape (match-string 1 str))))
                  (line (string-to-number (consult--strip-ansi-escape (match-string 2 str))))
                  (str (substring str (match-end 0)))
                  (loc (consult--format-location (file-relative-name file) line)))
-            (while (string-match consult--grep-match-regexp str)
+            (while (string-match consult-grep-match-regexp str)
               (setq str (concat (substring str 0 (match-beginning 0))
                                 (propertize (substring (match-string 1 str)) 'face 'consult-preview-match)
                                 (substring str (match-end 0)))))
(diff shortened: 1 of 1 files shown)


minad:
Yes, this is planned but not yet done since I am still experimenting. Therefore the variables are not public yet. If we do this the find command variables should also go public.

I think consult-grep works with bsd greps, at least on mac, I've heard. But it seems not to work with the openbsd variant.

The exclamation mark means that the command failed. There is also the ` *consult-async-stderr*` buffer to check.

Alternatively we could adjust some options to be more friendly, e.g using the short variants by default if that works on more platforms.

Not using --null is probably unacceptable since this will decrease robustness on the most used platforms. But I sympathize at least somewhat with broader support, I also use openbsd from time to time. 

The built-in emacs grep regexp supports null and non-null. And it is a mess in grep.el. This is a route I don't want to go.

omar-polo:
(sorry for turning this PR into an issue)

> I think consult-grep works with bsd greps, at least on mac, I've heard. But it seems not to work with the openbsd variant.

I just checked the freebsd manual and it provides a `--null` long argument, mac probably uses the same code.  And I see why it's useful, I was just trying to cope with the fact that it isn't supported here. 

> The exclamation mark means that the command failed. There is also the *consult-async-stderr* buffer to check.

I'm sorry, but I don't seem to have a `*consult-async-stderr*` buffer or any buffer with "consult" in the name (or at least Ibuffer and consult-buffer doesn't show it).

> The built-in emacs grep regexp supports null and non-null. And it is a mess in grep.el. This is a route I don't want to go.

I'm not asking for that :)

I'll probably give a shot at adding a `--null` flag to OpenBSD' grep and hope it gets accepted.

Thanks for looking into this!

P.S. consult-ripgrep works, when I opened the PR it wasn't because I changed by accident `consult--ripgrep-command`, PEBKAC as always.

minad:
The missing buffer has a space in front. It is ephemeral. You can select it if you enter space as first character in switch-to-buffer or consult-buffer.

minad:
> I'll probably give a shot at adding a --null flag to OpenBSD' grep and hope it gets accepted.

Oh that's a long shot :)

As I said for me it would generally be okay to use the short options if that makes it easier for openbsd. Just let me know which changes are needed!

omar-polo:
> The missing buffer has a space in front. It is ephemeral. You can select it if you enter space as first character in switch-to-buffer or consult-buffer.

Arg, stupid me, I should have thought of that.

> Oh that's a long shot :)

Not really, I have a patched grep with a --null that works with consult \o/

(consult was just the drop that that made glass overflow, there were other cases in the past were I used something that assumed a --null flag)

> As I said for me it would generally be okay to use the short options if that makes it easier for openbsd. Just let me know which changes are needed!

With the patch to add `--null`, I can use consult-grep with these flags:

```elisp
(setq consult--grep-command '("grep" "--null" "--line-buffered" "-E" "-n" "-I" "-R" "." "-e"))
```

i.e. without `---color`, without `--exclude-dir`, with `-E` instead of `--extended-regexp`, `-n` instead of `--line-number` and `-R` instead of `-r`.

Anyway, at least by looking at the manpages, both grep on [freebsd](https://manpage.me/?q=grep) and [netbsd](https://manpage.me/index.cgi?apropos=0&q=grep&sektion=0&manpath=NetBSD+7.1&arch=default&format=html) support the flags you're using (other than obviously GNU grep and probably grep on mac), so I don't really see the point to change them only for OpenBSD.

I'll upstream my patch for grep and wait until you feel comfortable to make `consult-*--command` and friends public.  This PR was just to avoid changing a "private" variable, because I didn't knew they were still experimental.

Thanks!

minad:
Regarding color - this is somewhat optional - but by not having colors, you won't see the highlight of the matches.

If I were you I would probably just install gnu grep...

omar-polo:
oh yeah, I forgot to mention that I can install the package ggrep to have gnu grep, that's an option too.

anyway, thanks for making consult, and the grep interaction in particular, it's amazing :)

minad:
Thank you! It is good to know about ggrep, so if the issue with openbsd grep comes up again, we can point that out!

PR #140, opened 2021-01-07
consult-bookmark: Fix creating bookmarks and add more handlers

clemera:
Thank you very much for implementing it here are my suggestions for more narrowing prefixes. Also creation of bookmarks when providing a non existing name did not work.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -215,7 +215,14 @@ You may want to add a function which pulses the current line, e.g.,
   :type 'integer)
 
 (defcustom consult-bookmark-narrow
-  '((bookmark-default-handler ?f "File"))
+  '((bookmark-default-handler ?f "File")
+    (help-bookmark-jump ?h "Help")
+    (Info-bookmark-jump ?i "Info")
+    (image-bookmark-jump ?p "Picture")
+    (doc-view-bookmark-jump ?d "Docview")
+    (Man-bookmark-jump ?m "Man")
+    (woman-bookmark-jump ?w "Woman")
+    (gnus-summary-bookmark-jump ?g "Gnus"))
   "Bookmark narrowing list.
 
 Each element of the list must have the form '(handler char name)."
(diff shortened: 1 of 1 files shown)


minad:
@clemera Thanks! Are these all standard bookmarks? Will some of those work as fast preview? Just file/position markers or are they all more sophisticated such that we cannot easily preview them. In order to perform preview one could also just run the bookmark jump, but then this might have bad side effects and performance issues. For this reason I only perform the preview when I know well what is going to happen or when I control it.

clemera:
Yes these are are from Emacs core. I don't think preview works when there is a handler, I did not test it (I generally don't use automatic preview) but there is a comment that only bookmarks without a handler are previewed.

clemera:
Yes these are are from Emacs core. I haven't looked into if they could be make work with preview (I generally don't use automatic preview). I would be interested in a manual option so I can press a key and preview the candidate.

minad:
Manual preview is just an Embark action I guess? Do you use no preview at all? For example there is also the yank preview.

clemera:
I use the regular yank-pop usually, could I add an embark action which uses the consult preview?

minad:
No preview is not an action. I meant in order to manually preview you could just execute the default action. This would work with consult-buffer. But then you don't get the auto closing behavior of temporary preview buffers as in consult-grep for example. Maybe it really makes sense to add a consult specific action which triggers the preview.

clemera:
Actually having auto preview for consult yank only would be nice, I haven't really setup consult for my purposes, yet.

minad:
@clemera Okay. Right now it is possible to enable preview globally and then disable it per action. Probably it would be nice to define the default preview mode globally (off, auto, manual) and per command. The manual mode would then require a keybinding in the minibuffer map.

clemera:
Does it need to be mode? Maybe it could just be a variable?

minad:
See #143, Maybe continue the discussion there. Yes it has originally been a mode since I did some setup. This is gone now but I have been hesitant to make it a variable in order to avoid breakage. When or if #143 comes we could break it and we necessarily have. Then it should also be a variable.

clemera:
Okay nice, thanks!

Issue #108, opened 2021-01-01
1. make-process default-directory needs trailing slash on mac, 2. pcase-let broken on emacs 28


jixiuf:
With `consult-ripgrep-project-root` ,consult-ripgrep still search in current directory instead of project root.
```
(defun consult-ripgrep-project-root()
  (let ((consult-project-root-function  #'(lambda() (project-root (project-current t)))) )
    (call-interactively 'consult-ripgrep)))

```

minad:
@jixiuf I cannot confirm this. Maybe project-root returns nil, then it will just use the current directory. Please ensure that you use the newest version and then test something like the following:

~~~ elisp
(let ((consult-project-root-function (lambda () "/etc")))
  (call-interactively #'consult-ripgrep))
~~~

jixiuf:
> @jixiuf I cannot confirm this. Maybe project-root returns nil, then it will just use the current directory. Please ensure that you use the newest version and then test something like the following:
> 
> ```emacs-lisp
> (let ((consult-project-root-function (lambda () "/etc")))
>   (call-interactively #'consult-ripgrep))
> ```

I tried that , still does not work, the prompt changed to `Ripgrep in /etc` but ,it still search in current directory,
And My Emacs version GNU Emacs 28.0.50 (build 1, x86_64-apple-darwin19.6.0, NS appkit-1894.60 Version 10.15.7 (Build 19H114))

minad:
Please try `/etc/`! See https://emacs.stackexchange.com/questions/5735/temporarily-set-the-default-directory-for-start-process

minad:
This seems to make a difference on mac. I will push a fix for this.

jixiuf:
this works for emacs 27.1 on macOS, but doesn't work for emacs master.

minad:
@jixiuf can you debug this? This is probably a mac specific upstream bug. Not the fault of consult.

jixiuf:
this can be reproduced with
```
(pcase-let ((`(,default-directory) '( "/tmp/")))
    (call-interactively 'find-file)) ; doesn't change default-directory to "/tmp/"
```


```
(pcase-let ((`(,default-directory) '( "/tmp/")))
    (message default-directory) ;with this line the default-directory change to "/tmp/", don't not why
    (call-interactively 'find-file))
```

minad:
@jixiuf seems like a pcase optimization bug. Can you report this upstream via emacs-report-bug?

jixiuf:
[...]

PR #142, opened 2021-01-07
Make sure the grep commands are executed in the right directory

plexus:
The grep should happen in the context of the directory returned by
`consult--directory-prompt`, i.e. the current project root if
`consult-project-root-function` is set.

This means two things, making sure `default-directory` is set correctly when the
command executes, and making sure `expand-file-name` is called with the right
directory so the relative paths are converted to absolute paths correctly.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -1035,7 +1035,8 @@ the comma is passed to ASYNC, the second part is used for filtering."
 
 ASYNC is the async function which receives the candidates.
 CMD is the command argument list."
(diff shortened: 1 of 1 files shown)


plexus:
Looking at existing issues it might be I was actually running into https://github.com/minad/consult/issues/108

minad:
Yes, this is probably #108. I refuse to fix this until I have confirmation that the Emacs 28 regression is actually intensional. We need confirmation by Stephan Monnier that this is the case. Otherwise upstream should be fixed.

PR #147, opened 2021-01-07
add consult-man

omar-polo:
Hello,

This adds the ability to search for a manual page using `man -k`.

Note: this was tested only on OpenBSD, which uses `mandoc`.  Mandoc is used also by various linux distribution, void linux for instance comes to mind, but not all.  Unfortunately I cannot test this on other systems and I don't have knowledge on difference between `man`/`apropos` implementations.

It's different (and more powerful) than the `M-x man` command because it allows to search on the content of the manpages (and use the apropos expression syntax) instead of only searching on the page name.  To be clear: with this I can type `Va=optind` and see all the manual pages where the variable `optind` is mentioned, something that's not possible with `M-x man`.

Another thing: it would be nice if we could split the user input on spaces: this will allows us to make better use of the expression syntax.  For instance, I can

```
$ apropos =.cf =.cnf =.conf
```

to list all the manpages whose description matches ".cf", "cnf" or ".conf".  This pr calls man with arg `("man" "-k" ".cf .cnf .conf")`, we should call it with `("man" "-k" ".cf" ".cnf" ".conf")`.

(a note about the name: I initially called it `consult-apropos` because I intented to use apropos(1), but then I saw that there is already an `consult-apropos`, for emacs' apropos, so I changed the name to `consult-man` and changed `consult-man-command` to "man -k" to be more consistent.)

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -201,6 +201,11 @@ You may want to add a function which pulses the current line, e.g.,
   "Command line arguments for locate."
   :type '(repeat string))
 
+(defcustom consult-man-command
+  '("man" "-k")
+  "Command line arguments for man apropos."
+  :type '(repeat string))
+
 (defcustom consult-preview-key 'any
   "Preview trigger, can be nil, 'any or a key."
   :type '(choice (const nil) (const any) vector))
@@ -311,6 +316,7 @@ Each element of the list must have the form '(handler char name)."
 (defvar consult--error-history nil)
 (defvar consult--grep-history nil)
 (defvar consult--find-history nil)
+(defvar consult--man-history nil)
 (defvar consult--line-history nil)
 (defvar consult--apropos-history nil)
 (defvar consult--theme-history nil)
@@ -2417,6 +2423,33 @@ CMD is the find argument list."
     :category 'file
     :history '(:input consult--find-history))))
 
+(defun consult--man-async (cmd)
+  "Async function for `consult--man'.
+
+CMD is the man argument list."
+  (thread-first (consult--async-sink)
+    (consult--async-refresh-timer)
+    (consult--async-process (consult--command-args cmd))
+    (consult--async-throttle)
+    (consult--async-split)))
+
+(defun consult--man (prompt cmd initial)
+  "Run man CMD with INITIAL input.
+
+PROMPT is the prompt.
+CMD is the man argument list."
+  (let ((match (consult--read
+                prompt
+                (consult--man-async cmd)
+                :sort nil
+                :require-match t
+                :initial (concat consult-async-default-split initial)
+                :add-history (concat consult-async-default-split (thing-at-point 'symbol))
+                ;; XXX: category?
+                :history '(:input consult--man-history))))
+    (string-match "\\([^[:blank:]]+\\) - " match)
+    (man (match-string 1 match))))
+
 ;;;###autoload
 (defun consult-find (&optional dir initial)
   "Search for regexp with find in DIR with INITIAL input."
@@ -2430,6 +2463,12 @@ CMD is the find argument list."
   "Search for regexp with locate with INITIAL input."
   (interactive)
   (consult--find "Locate: " consult-locate-command initial))
+
+;;;##autoload
+(defun consult-man (&optional initial)
+  "Search for regexp with apropos with INITIAL input."
+  (interactive)
+  (consult--man "Man: " consult-man-command initial))
 
 ;;;; default completion-system support
 


omar-polo:
Oh, I forgot to say, the regexp is to match on the page name.  For instance, `man -k printf` returns 

```
printf(1) - formatted output
aa_printf(3) - print text to AA-lib output buffers.
agentx, agentx_log_fatal, [...] agentx_wantwrite, agentx_write(3) - manage an interface to an agentx master
[...]
kcgi_buf_printf(3) - put variable content into a kcgi buffer
khtml_printf(3) - put variable content for kcgihtml
khttp_printf(3) - put variable content for kcgi
[...]
```

(`[...]` is because some lines are way, way, way too long).

The regexp extract the page name with the number, so we can call man on the right page (sometimes there are pages with the same name but on different sections: see printf(1) and printf(3) for instance)

Also, I'm not sure what the category should be.

minad:
It is interesting to have the ability to search through the descriptions too, in contrast to builtin `man` and `woman`. It seems that `man -k` matches on the description too.

* I would like to discuss the alternative, since I am hesitant to only add commands which allow matching on annotations without other benefits over the built-ins. We could instead annotate `man` and `woman` commands with Marginalia, then you could search through descriptions via an Embark buffer and `flush-lines`/`keep-lines`. The completing-read matching would still only happen on the command name.

* Is it possible to obtain the man apropos information via other means, e.g., via some code in woman.el? Instead of using an external program?

* Instead of calling the `man` command on the result it may make sense to offer an option to either to use man or woman?

omar-polo:
I would love to have annotation with marginalia for manual pages (and since you gave me this idea I'll try to see if I can do something in that regard), but this pr is still more powerful than what we could achieve with marginalia+embark.

For instance, have you ever wondered which manual pages on your system mentions the file `resolv.conf`.  Type `M-x consult-man RET Pa=resolv.conf` and you'll find a whole bunch of manual pages, from dig(1) to intro(8).  And it's interactive, which is awesome 👍:)

> Is it possible to obtain the man apropos information via other means, e.g., via some code in woman.el? Instead of using an external program?

How is this different than using `locate`, `find`, `grep` or `ripgrep`?  Also, if a system doesn't have `man(1)`, the users in that system probably won't need `consult-man` either.

> Instead of calling the man command on the result it may make sense to offer an option to either to use man or woman?

I don't understand this, I'm already using `man` (the emacs interactive command from man.el) to display the page.  I admit I never used `woman`, and I'll take a look at that (I just tried to read the manpage for `printf` using woman, but it opened the raw `/usr/share/man/man1/printf.1` and called the buffer `*WoMan 1 printf*`).

I do agree though that this is the weirdest `consult-*` command so far, since it uses its own syntax (the apropos one) for the query.

Counter-idea: provide a `consult-live-command` (or something like that) to interactively run a command with input from minibuffer and do stuff with the selected line?  This would allow the users to implement `consult-man` on top of this abstraction, and avoid adding further dependecies (such as man(1)) to consult.  I'll probably play with this idea

minad:
> I would love to have annotation with marginalia for manual pages (and since you gave me this idea I'll try to see if I can do something in that regard), but this pr is still more powerful than what we could achieve with marginalia+embark.

I agree that this PR is more powerful. But the Marginalia annotations would be less controversial for now and we could make a more well informed decision here afterwards. For Marginalia you already need some way to extract the apropos description!

> ...man RET Pa=resolv.conf 

What is this Pa? I tried `man -k Pa=resolv.conf` but it does not match anything. (See below, this is openbsd specific)

> How is this different than using locate, find, grep or ripgrep? Also, if a system doesn't have man(1), the users in that system probably won't need consult-man either.

It is not different, maybe only that `consult-man` is less general purpose than grep and find commands. So I am more hesitant here. I turned down a request for `consult-linux-app` for example. But I don't say no to `consult-man`, I just want to know if the functionality already exists (maybe partially) in the built-in man.el and woman.el packages.

> I don't understand this, I'm already using man (the emacs interactive command from man.el) to display the page. I admit I never used woman, and I'll take a look at that (I just tried to read the manpage for printf using woman, but it opened the raw /usr/share/man/man1/printf.1 and called the buffer *WoMan 1 printf*).

I don't know why there is man and woman, but there is some reason. From what I know, woman renders the manual using pure elisp without the external program. If we have both possibilities built into Emacs we should offer the possibility to select one or the other. However this may not make sense if you necessarily have to invoke `man -k`. Therefore I was also wondering if the man apropos functionality is somehow already present in woman.el.

> I do agree though that this is the weirdest consult-* command so far, since it uses its own syntax (the apropos one) for the query.

Do you have a reference for this? This Pa thing seems to be an openbsd or bsd thing, I only found it in the bsd man page of apropos.

> Counter-idea: provide a consult-live-command (or something like that) to interactively run a command with input from minibuffer and do stuff with the selected line? This would allow the users to implement consult-man on top of this abstraction, and avoid adding further dependecies (such as man(1)) to consult. I'll probably play with this idea

This is something I considered myself. It would be nice to have something general. The point with grep and find is that they are somehow natural to Emacs since we are working all the time with files and there is also built-in grep. But since there is also built-in man and woman, my argument is kind of weak. But find and grep is still more important than man I would argue.

omar-polo:
> What is this Pa? I tried man -k Pa=resolv.conf but it does not match anything. (See below, this is openbsd specific)

Ugh, my bad, I should have checked more.  It's not openbsd specific, but [mandoc](https://mandoc.bsd.lv/) specific.  mandoc originates from OpenBSD, but also present by default on some linux distros, and since I was using Void before moving to OpenBSD, I just gave for it for granted.

> It is not different, maybe only that consult-man is less general purpose than grep and find commands. So I am more hesitant here. I turned down a request for consult-linux-app for example. But I don't say no to consult-man, I just want to know if the functionality already exists (maybe partially) in the built-in man.el and woman.el packages.

fair enough

> I don't know why there is man and woman, but there is some reason. From what I know, woman renders the manual using pure elisp without the external program. If we have both possibilities built into Emacs we should offer the possibility to select one or the other. However this may not make sense if you necessarily have to invoke man -k. Therefore I was also wondering if the man apropos functionality is somehow already present in woman.el.

I still haven't read woman.el, but I doubt this is possible.  mandoc leverages on mandoc.db, a database of manpages, while if it's true that woman is an elisp implementation of roff/troff/man it won't be as fast (and probably as accurate,  different systems uses different formats for the man pages).  At that point, one is better off grepping for "^\\.Pa resolv.conf" in `/usr/share/man` and call woman afterwards.

> This is something I considered myself. It would be nice to have something general. The point with grep and find is that they are somehow natural to Emacs since we are working all the time with files and there is also built-in grep. But since there is also built-in man and woman, my argument is kind of weak. But find and grep is still more important than man I would argue.

Maybe we can find a way to generalize the way `consult-grep`, `consult-find` etc are defined, and made that public.  There are various places where I think I could benefit from something like that.

Also, probably this isn't even needed, since `consult--read` seems already flexible enough, but I'm not familiar enough with the internal API to see if they can misbehave if used against random commands.

minad:
> I still haven't read woman.el, but I doubt this is possible. mandoc leverages on mandoc.db, a database of manpages, while if it's true that woman is an elisp implementation of roff/troff/man it won't be as fast (and probably as accurate, different systems uses different formats for the man pages). At that point, one is better off grepping for "^\.Pa resolv.conf" in /usr/share/man and call woman afterwards.

This makes a lot of sense. There is some database we are querying using an external program. It is okay to have something like this in consult. But I would like to first investigate a bit the generic idea and maybe try to get annotators for marginalia.

> Maybe we can find a way to generalize the way consult-grep, consult-find etc are defined, and made that public. There are various places where I think I could benefit from something like that.

Agree! We could still have the wrappers. But their definition is already not that complicated, so most generalization already happened in those async pipeline steps. Maybe this could be made a bit more concise. But I like the functional style of that.

> Also, probably this isn't even needed, since consult--read seems already flexible enough, but I'm not familiar enough with the internal API to see if they can misbehave if used against random commands.

I don't think there are many bad things possible, except the usual things if you are allowed to invoke arbitrary commands - so basically anything can happen ;) Furthermore the commands should be interruptible. It wouldn't be a good idea to use them one commands with side effects. So I guess we should be careful with the generic solution. But it is worth an exploration...

minad:
@omar-polo I decided that I would like to pull this, since I think it is useful, and general enough. Furthermore it is justified to have this here since it is a clear improvement over the status quo:

* allows to search through the man database/descriptions (but dependent on the power of the backend)
* the built-in man command also supports apropos: (man "-k ..."), but non-interactive
* woman.el does not support apropos, there is a TODO in the woman.el file, see https://github.com/emacs-mirror/emacs/blob/29c7f8c915c3889dfd5b25878aa0692f826cd38f/lisp/woman.el#L327

Furthermore I am happy to have usability/accessibility utilities!

minad:
See here: https://github.com/minad/consult/commit/51c05c874dcb767e07a6a1bf4522e06599bbca45

PR #159, opened 2021-01-13
Add consult-flush-lines

oantolin:
I renamed the consult-keep-lines function to consult--line-subset and added parameters for the prompt and the match function. The new consult-keep-lines call that function with the match function given by the match hook, and consult-keep-lines calls it with a function derived from that which puts the results of the match function in a hashtable, and returns the lines not in the hashtable.


minad:
I pulled it with some minor improvements. Please take a look!

PR #171, opened 2021-01-18
Isearch history

astoff:
Do you think this is of broad enough interest? It's meant to be bound in `isearch-mode-map`.

I'm not sure if there's an useful completion category to report in this `consult--read`.

### DIFF
--- a/README.org
+++ b/README.org
@@ -253,6 +253,12 @@ variables and functions with their descriptions.
  - =consult-history=: Insert a string from the current buffer history.
    This command can be invoked from the minibuffer. In that case the
    history stored in the =minibuffer-history-variable= is used.
+ - =consult-isearch=: During an Isearch session, this command picks a
+   search string from history. Outside of Isearch, it reads a string
+   from the minibuffer and starts searching; this is useful, for
+   instance, as an Embark action. (Note that Isearch does not normally
+   read from the minibuffer. What you see at the bottom of the screen
+   is the echo area!)
 
 ** Major and minor modes
  :properties:
@@ -492,6 +498,14 @@ the box with the default completion, Icomplete and Selectrum.
           ("M-y" . consult-yank-pop)
           ("<help> a" . consult-apropos))
 
+   ;; Bind `consult-isearch' if desired.  The example below overloads
+   ;; it with `consult-history', so that you can access the search
+   ;; history with the same key, but only after starting Isearch.  Of
+   ;; course you can bind the command to a dedicated key in the global
+   ;; map instead.
+   :bind (:map isearch-mode-map
+          ("C-c h" . consult-isearch))
+
    ;; The :init configuration is always executed (Not lazy!)
    :init
 
@@ -545,6 +559,11 @@ the box with the default completion, Icomplete and Selectrum.
  (use-package consult-flycheck
    :bind (:map flycheck-command-map
                ("!" . consult-flycheck)))
+
+ ;; Optionally integrate `consult-isearch' with Embark.
+ (use-package embark
+   :bind (:map embark-general-map
+          ("C-s" . consult-isearch)))
  #+end_src
 
 ** Customizable variables
--- a/consult.el
+++ b/consult.el
@@ -2431,6 +2431,19 @@ In order to select from a specific HISTORY, pass the history variable as argumen
 
 ;;;;; Command: consult-isearch
 
+(defun consult--isearch-category (cand)
+  "The Isearch mode of history element CAND, encoded as a character."
+  (let* ((props (plist-member (text-properties-at 0 cand)
+                              'isearch-regexp-function))
+         (fun (cadr props)))
+    (cond ((null props) ?r)
+          ((null fun) ?l)
+          ((eq fun t) ?w)
+          ((eq fun 'word-search-regexp) ?w)
+          ((eq fun 'isearch-symbol-regexp) ?_)
+          ((eq fun 'char-fold-to-regexp) ?')
+          (t ?c))))
+
 ;;;###autoload
 (defun consult-isearch ()
   "Read a search string with completion from history.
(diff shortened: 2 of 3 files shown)


minad:
I will try it. I wonder if we should make this part of consult-history if that is possible. If an isearch is active use the isearch history. consult-history is similar to an Embark style context sensitive command.

astoff:
That would be nice, but, strangely, Isearch is implemented as a minor mode that hijacks everything you type instead of a command that reads from the minibuffer. No idea what the rationale was for this choice.

minad:
Yes, it would require special casing which is quite different from the existing action of inserting. I wonder if that would be preferred over an extra command. I like the context sensitive style, but I am not sure if the special casing is a good idea. I will play a bit around to see.

astoff:
Binding `consult-history` in global-map and `consult-isearch-history` to the same key combination in `isearch-mode-map` makes it context-sensitive. It's not incredibly beautiful, but it's the logical consequence of Isearch's design.

minad:
Right that's probably the cleaner option.

minad:
Alternatively rename the command to consult-isearch and make it available when no isearch session is active. Then the history will be shown for selection and a new session is started.

astoff:
Right, why not? Seems a bit academic, but this allows, for instance, to start a search as an Embark action.

astoff:
Ah, the possibility of calling the command as an Embark action means that it shouldn't fail if the history is empty, right? Also, should the function have any arguments? Currently, it hardcodes the initial search direction (as forward) and uses the default search style (i.e., you can't choose between literal vs. regexp vs. word vs. symbol, except by binding a dynamic var). As an Embark action, it would be marginally advantageous to be able to start a symbol search.  But I think none of that justifies making the function more complicated. What do you think?

minad:
Yes, it should not fail. Regarding the other questions I am undecided.

astoff:
Ok, I'm happy with the command now. I can add something to the documentation if you like (suggesting to bind it to `C-c h` in `isearch-mode-map`, and mentioning Embark as well).

PS: There are some loooong lines in your code!

minad:
> Ok, I'm happy with the command now. I can add something to the documentation if you like (suggesting to bind it to C-c h in isearch-mode-map, and mentioning Embark as well).

Yes, please add some documentation to README.org! I think it makes sense to bind it to another key since it is not only a history command, but you can also start it when no isearch is active.

> PS: There are some loooong lines in your code!

I added some newlines ;)

README.org
```diff
@@ -545,6 +559,11 @@ the box with the default completion, Icomplete and Selectrum.
  (use-package consult-flycheck
    :bind (:map flycheck-command-map
                ("!" . consult-flycheck)))
+
+ ;; Optionally integrate `consult-isearch' with Embark.
```
astoff, line 563: @oantolin Do you endorse this binding in `embark-general-map`?

minad:
@astoff Sounds reasonable. Nevertheless, I will probably revise the proposed bindings in the README after this PR. Right now most commands are bound to `M-g`. I will probably move the search-related to commands to `M-s`. Then I will also propose a `consult-isearch` binding under `M-s` additionally to the binding in the isearch map.

minad:
I just tried the command. There is a flicker when starting `consult-isearch` from outside a running isearch session. Can this be avoided by reading the string first and then starting the search with the given input?

astoff:
> Can this be avoided by reading the string first and then starting the search with the given input?

Perhaps, but when isearch is on the `consult--read` must be inside the `with-suspended-isearch` macro, so I'm afraid this would require separate `consult--read`s for each case.

Apparently the flicker happens because the `(setq isearch-new-string ...)` form takes about 200 ms (with `consult-isearch` called via Embark so there's no interactivity). The stuff before it (activating `isearch-mode` and entering `with-suspended-isearch`) is pretty fast. Maybe the minibuffer is awaiting on some timer set up by the previous isearch stuff? I tried binding some variables like `isearch-lazy-highlight` around the call to `isearch-mode`, but nothing seems to make much of a difference.

Isearch is so weird, I'm really curious why it was designed like that.

minad:
A separate consult--read would be fine. Maybe factor it out as a consult--isearch-read function?

minad:
> Isearch is so weird, I'm really curious why it was designed like that.

I guess there is a reason for https://github.com/raxod502/ctrlf and https://github.com/thierryvolpiatto/isearch-light after all ;)

astoff:
I pushed a change in that direction. We should test more if we keep it. But it extremely ugly and unreadable. I think I still prefer the small visual glitch.

I knew about Ctrlf. It looks promising, but it bothers me that it's a minor mode. There will always be conflicts if you bind a key like `C-s` at a keymap with high precedence. The other one might also be nice, once they test it in the default theme and realize their colors are crazy :-)

minad:
> I pushed a change in that direction. We should test more if we decide to keep it. But it's extremely ugly and unreadable. I still prefer the small visual glitch.

Yes, definitely. I prefer to have no glitches.

consult.el
```diff
@@ -2429,6 +2429,38 @@ In order to select from a specific HISTORY, pass the history variable as argumen
...
+     (cons str (mapconcat 'isearch-text-char-description str ""))))
+
+;;;###autoload
+(defun consult-isearch ()
+  "Read a search string with completion from history.
+
+This replaces the current search string if Isearch is active, and
+starts a new Isearch session otherwise."
+  (interactive)
+  (let* ((hist (if (if isearch-mode isearch-regexp
```
minad, line 2445: The hist variable should go away I suppose?

minad:
I have one more improvement proposal: Can we merge the two histories, regexp and literal, and then depending on the selection switch the search mode? Each candidate should be marked with a prefix "Regexp" or "Literal" and then one could even add the Consult :narrow option. When entering a candidate which is not in the list, either use the current search mode or if no search is active use the literal search?

consult.el
astoff: Oops...

astoff:
I wonder a bit if this isn't conflating two things (looking up history and cycling the search style) that are easier to do separately. Maybe it would end up just feeling natural, but it might also end up being confusing, I'm not sure. Anyway, there would be some details to sort out. How do we sort the entries from the two histories without losing the relative order within each category, which is important? Also, if we switch from regexp to normal search, how do we choose the right style (literal, word, symbol, etc)?

minad:
> I wonder a bit if this isn't conflating two things (looking up history and cycling the search style) that are easier to do separately. 

Yes, it is. But it allows you to quickly scan through the recent search strings. Maybe you have a better idea. It is not that uncommon to have multiple different candidates smashed together in a consult command, e.g. as in consult-buffer.

> How do we sort the entries from the two histories without losing the relative order within each category, which is important?

I have not thought about this in detail. Maybe I would just concatenate the histories in their original order?

> Also, if we switch from regexp to normal search, how do we choose the right style (literal, word, symbol, etc)?

Normal search shares the history search-ring, right? But the style is not captured in the history? So we cannot do anything better than choosing the default?

astoff:
All right, I think I'm convinced :smiley:. I'll work on this, but not until the weekend.

astoff:
Does this last commit solve the flickering for you? Setting `inhibit-redisplay` looks a bit unsound, but it seems to be just the right thing to do.

minad:
I will check. Looks at least better than before :) But independeny, I would still like to have some possibility to search through the different histories regexp/normal simultaneously.

astoff:
Sure, this is just about the visual aspect.

astoff:
All right, there it is.  Now that I see the narrowing in action, I quite like it!

I suppose you would need to include a variant of `consult--isearch-category` in marginalia so that search mode annotations are displayed.  If you want to make it so that narrowing only works if marginarlia-mode is on, to avoid repeated code, I wouldn't object.  I already added a completion `:category`.

Now, about a possible consult-line/iserach integration: you will notice from the code that adding stuff into isearch's history can be a bit of problem, because the history includes metadata such as the relevant Isearch mode.

minad:
Oh wow, isearch stores the search function after all. Thank you for going through with this. I think this will be cool.

> I suppose you would need to include a variant of consult--isearch-category in marginalia so that search mode annotations are displayed. If you want to make it so that narrowing only works if marginarlia-mode is on, to avoid repeated code, I wouldn't object. I already added a completion :category.

Regarding annotations, we could potentially move them to Marginalia, but you could also just show your own annotations by prefixing the candidates with hidden strings and a 'display property, see for example `consult--kmacro-candidates`.

 I would only move the annotators if we win something by reusing Marginalia code. This is for example the case for consult-buffer, which can reuse the file and buffer annotators!

> Now, about a possible consult-line/iserach integration: you will notice from the code that adding stuff into isearch's history can be a bit of problem, because the history includes metadata such as the relevant Isearch mode.

Yes, this should be improved a bit more I guess. Can you use a :lookup function to return the string in the desired form? If you use :require-match, we can always get the original search type. Without :require-match we can just use the current search type or the search default? But I understand that this is a problem, since you may want to restart a search from the history and edit it slightly.

minad:
I pulled `consult-isearch` with a few modifications. Please take a look and give it a test!

astoff:
Looks good, even better indeed ;-). Just one comment: I didn't add a user-error if the history is empty because this would break the Embark use-case in the uncommon situation where this is the very first search the user does. (Also a quick question for my own sake: why did you prefer `let*` at the spot where `inhibit-redisplay` is set?)

PR #173, opened 2021-01-20
consult-yasnippet

mohkale:
This command works as a parallel to [ivy-yasnippet](https://github.com/mkcms/ivy-yasnippet) and allows you to
expand yasnippet snippets through a completing-read interface. It
supports previewing the current snippet expansion and overwriting the
marked region with a new snippet completion.

## Embark
This package also provides the `consult-yasnippet-visit-snippet-file`
package. Which can be used as an alternative action for
`consult-yasnippet` through for eg. embark:

```lisp
(require 'embark)

(embark-define-keymap embark-yasnippet-completion-actions
                      "Embark actions for `consult-yasnippet' and derivatives"
                      ;; NOTE: Binding differs from `ivy-yasnippet' which uses "v".
                      ("d" consult-yasnippet-visit-snippet-file))

(add-to-list 'embark-keymap-alist '(yasnippet . embark-yasnippet-completion-actions))
```

## marginalia
The default completion interface simply shows the name associated with a
snippet. You can annotate it with the prefix/key needed to expand to it
using marginalia.

```lisp
(require 'marginalia)

(defun marginalia-annotate-yasnippet (cand)
  ;; NOTE: Maybe there's a less indirect way to do this :/.
  (when-let* ((cand (marginalia--full-candidate cand))
              (template (alist-get cand consult-yasnippet--snippets nil nil #'string-equal))
              (key (yas--template-key template)))
    (concat " " (propertize (concat "[" key "]")
                            ;; TODO: custom face
                            'face 'font-lock-type-face))))

(push '(yasnippet . marginalia-annotate-yasnippet) marginalia-annotators-light)
(push '(yasnippet . marginalia-annotate-yasnippet) marginalia-annotators-heavy)
```


mohkale:
@minad

Please let me know if there's anything else you'd like me to add/modify.

minad:
Thanks, Mohsin! I will give you a review in the source. It is a good idea to provide Embark action commands here. But instead of Marginalia it is better to annotate the candidates directly using the 'display property.

~~~ elisp
(propertize "prefix" 'display "prefix-annotation")
~~~

README.org
```diff
@@ -236,6 +236,9 @@ variables and functions with their descriptions.
   components.
 - =consult-flymake=: Jump to Flymake diagnostic, like
   =consult-flycheck=.
+- =consult-yasnippet=: Expand [[https://github.com/joaotavora/yasnippet][yasnippet]] snippets in the current buffer.
+  This command is available from the additional =consult-yasnippet.el=
+  package.
```
minad, line 241: This should probably go to the editing section?

consult-yasnippet.el
```diff
@@ -0,0 +1,158 @@
...
+
+;;; Code:
+
+(require 'consult)
+(require 'yasnippet)
+
+(defgroup consult-yasnippet nil
+  "Consulting `completing-read' for yasnippet snippets."
+  :group 'consult
+  :prefix "consult-yasnippet-")
```
minad, line 43: The group is not needed. It would only be necessary if you define faces or custom variables.

consult-yasnippet.el
```diff
@@ -0,0 +1,158 @@
...
+   (let ((buffer-undo-list t))                                                  ; Prevent querying user (and showing previews) from updating the undo-history
+     (unwind-protect
+         (consult--read
+          "Choose a snippet: "
+          consult-yasnippet--snippets
+          :lookup 'consult--lookup-cdr
+          :require-match t
+          :preview #'consult-yasnippet--preview
+          :category 'yasnippet)
+       (consult-yasnippet--preview nil t)))))                                   ; Restore contents of region from before preview (while still ignoring undo history).
```
minad, line 143: The :preview function has a bit of an unexpected feature - even if previews are disabled, the preview function is called with the restore=t flag in the end. Therefore you don't need the `(consult-yasnippet--preview nil t)` call.

consult-yasnippet.el
```diff
@@ -0,0 +1,158 @@
...
+
+(defvar consult-yasnippet--buffer nil
+  "The buffer in which `consult-yasnippet' was begun")
+
+(defvar consult-yasnippet--region nil
+  "The position (a cons of the start and end `point's) of where `consult-yasnippet' was begun")
+
+(defvar consult-yasnippet--region-contents ""
+  "The original contents of `consult-yasnippet--region'.")
+
```
minad, line 56: You define multiple global variables here. These should be lexically bound variables in a closure returned by the `consult-yasnippet--preview` function.

consult-yasnippet.el
```diff
@@ -0,0 +1,158 @@
...
+  (goto-char (car consult-yasnippet--region))
+  (when (not (string-equal "" consult-yasnippet--region-contents))
+    (push-mark (point))
+    (push-mark (cdr consult-yasnippet--region) nil t))
+
+  (yas-expand-snippet (yas--template-content template)
+                      nil nil
+                      (yas--template-expand-env template)))
+
+(defun consult-yasnippet--preview (template _)
```
minad, line 69: This function should return a lambda, which captures the preview state.

consult-yasnippet.el
```diff
@@ -0,0 +1,158 @@
...
+             (mapcar (lambda (template)
+                       (cons (yas--template-name template) template))
+                     (yas--all-templates (yas--get-snippet-tables))))
+            (consult-yasnippet--buffer (current-buffer))
+            (consult-yasnippet--region (if (region-active-p)
+                                           (cons (region-beginning) (region-end))
+                                         (cons (point) (point))))
+            (consult-yasnippet--region-contents (buffer-substring (car consult-yasnippet--region)
+                                                                  (cdr consult-yasnippet--region))))
+       ,@body)))
```
minad, line 122: Most of this should probably go to the preview function then when it captures the state.

minad:
The main improvements I would like to see is the change to lexically captured state, since this fits better with how things are done in the other Consult functions. Probably that is different from how Ivy/Counsel is handling its state.

See `consult--preview-position` https://github.com/minad/consult/blob/6d45873bd0095b1017b098d1c645cfe115fe2116/consult.el#L848 or `consult--yank-read`https://github.com/minad/consult/blob/6d45873bd0095b1017b098d1c645cfe115fe2116/consult.el#L2167

consult-yasnippet.el
mohkale: huh weird, while I was writing this I ran into issues with previews not being removed when you cancel completing-read. Guess that was because of something else. I'll remove this :-).

consult-yasnippet.el
mohkale: Removed.

README.org
mohkale: Moved.

consult-yasnippet.el
mohkale: > These should be lexically bound variables in a closure returned by the consult-yasnippet--preview function.

That was the route I initiially went down but for some reason I couldn't access them in the closure. Let me try again.

consult-yasnippet.el
mohkale: @minad I still seem to have an issue with previews not being cleaned up after removing this. I'll push what I've got in a few minutes please double check whether there'es anything wrong with it.

consult-yasnippet.el
mohkale: Done

consult-yasnippet.el
mohkale: Done.

consult-yasnippet.el
minad: Regarding the arguments of the preview function, I should probably make this more clear:

* candidate=non-nil, restore=nil -> preview
* candidate=non-nil, restore=t -> restore and candidate has been selected
* candidate=nil, restore=t -> restore, abort, no candidate has been selected

Does this help?

mohkale: @minad 

That second case doesn't seem to be happening for me. Is `consult` passing the candidate in that case as well? That would explain the issue I'm having (`consult-yasnippet--preview` is then setting up the preview for that candidate which is then ignored by `consult-yasnippet` where it's re-entered). It's easier to demonstrate than explain.

Could you try:
1. Pulling this branch.
2. Requiring `consult-yasnippet`
3. Evaluate the following.
4. Try exanding a snippet.

```lisp
(defun consult-yasnippet--read-snippet ()
  (barf-if-buffer-read-only)

  (let* ((buffer-undo-list t))
    (consult--read
     "Choose a snippet: "
     (consult-yasnippet--candidates)
     :lookup 'consult--lookup-cdr
     :require-match t
     :preview (consult-yasnippet--preview)
     :category 'yasnippet)))
```

consult-yasnippet.el
mohkale: Done.

consult-yasnippet.el
minad: I just checked. I added this to `consult-theme` and it seems to work as I described. Does this not work for you?

~~~ elisp
       :preview (lambda (cand restore)
                  (message "%S %S" cand restore)
~~~

I see the following messages:

~~~
some-theme nil
some-other-theme nil
....
selected-theme t OR nil t
~~~

consult-yasnippet.el
```diff
@@ -0,0 +1,143 @@
...
+         (previewer (consult-yasnippet--preview)))
+    (unwind-protect
+        (consult--read
+         "Choose a snippet: "
+         (consult-yasnippet--candidates)
+         :lookup 'consult--lookup-cdr
+         :require-match t
+         :preview previewer
+         :category 'yasnippet)
+      (funcall previewer nil t))))
```
minad, line 126: Here you pass restore=t.

consult-yasnippet.el
```diff
@@ -0,0 +1,143 @@
...
+previews that're already active.
+
+When TEMPLATE is not given, this function essentially just resets
+the state of the current buffer to before any snippets were previewed."
+  (let* ((buf (current-buffer))
+         (region (if (region-active-p)
+                     (cons (region-beginning) (region-end))
+                   (cons (point) (point))))
+         (region-contents (buffer-substring (car region) (cdr region))))
+    (lambda (template _restore)
```
minad, line 66: And you ignore _restore? Why?

mohkale: Well you always have to restore... at least that's how `ivy-yasnippet` was setup. I've only passed `t` to it for consistency (in case sometime down the line we actually use that parameter).

consult-yasnippet.el
mohkale: It works, but it also explains why we need to restore on exit.

```lisp
(consult--read
 "Foo: "
 '("a" "b" "c" "d")
 :preview (lambda (cand restore)
            (message "%S %S" cand restore)))
```

If you eval this your message buffer will look like:

```lisp
"b" nil
"a" nil
"c" nil
"d" nil
"c" nil
"a" nil
"a" t
"a"
```

At the end you pass the candidate and restore, but seeing as we restore before we preview a candidate we end up restoring, then previiweing, then when `consult-yasnippet` is called it expands the snippet in the middle of the preview.

The way things are implemented now we need to have that `unwind-protect` preview restore block.

mohkale: I think I understand now, I'll fix this :-).

mohkale: Fixed.

minad: > At the end you pass the candidate and restore, but seeing as we restore before we preview a candidate we end up restoring, then previiweing, then when consult-yasnippet is called it expands the snippet in the middle of the preview. The way things are implemented now we need to have that unwind-protect preview restore block.

I don't really understand what you mean. If restore=t is set and a candidate is passed, the operation should be committed. Otherwise it should just be reverted. But I think preview should be rewritten to use overlays, so you will end up with separate preview and final expansion code in the end. See `consult--yank-read` for the overlay preview code.

mohkale:
@minad I've been meaning to say but I think `consult--read` needs better documentation messages. I find some of it hard to grok on the face of it. For example `:lookup` maps a candidate as it was in the original candidate list before passing to `:preview` or returning from `consult--read`. The docstring at the moment doesn't really express this. Similarly `:preview` doesn't describe the form or implied contract of the preview function. The behaviour you've described above would've been really nice to have available in the docs.

consult-yasnippet.el
mohkale: Yeah I misunderstood the point of `restore`. I thought the order was something like:

```
1. restore
2. preview
3. restore
4. preview
5. ...
```

but it's more akin to you keep previewing until you restore. So you either restore or you preview, but I was doing both on each preview. That's a misunderstanding on my part, sorry. I've set it up to skip previewing when restore is true. Next should be overlays.

minad:
Well, it is private API. So that's what you got for now :)

I agree that documentation should generally be improved. But time is limited and the project is still young.

mohkale:
@minad that's fair. Guess this whole package is so cool I'm kind of really eager to understand how it all works :+1:.

minad:
I should also add - the internals are still in flux. The behavior of the :preview restore has been changed very recently when I added consult-keep-lines and consult-focus-lines. Before that :preview was only called with activated previews and not always during restore. But since the setup of the closure is executed always it kind of makes more sense to always call the destructor/restoration too. And it made it easier to implement consult-keep-lines and consult-focus-lines :)

mohkale:
@minad 

I've never really used overlays before, but from the looks of it I'll have to apply an overlay to the region with the `display` property set to an expansion of the snippet. I'm kinda not sure how to get that expansion though :? ivy-yasnippet just expands the entire snippet and then performs substitutions. I could expand it in a temp buffer but I'm not sure whether all the associated font-lock rules would still be in affect. Is this what you're suggesting?

mohkale:
@minad 

So I'm trying out overlays... but yeah, if I expand in a temp-buffer I lose any syntax highlighting in the snippet expansion.

```lisp
(defun consult-yasnippet--foo (template)
  (with-temp-buffer
    (yas-minor-mode +1)
    (unwind-protect
        (yas-expand-snippet (yas--template-content template)
                            nil nil
                            (yas--template-expand-env template))
      (unwind-protect
          (mapc #'yas--commit-snippet
                (yas-active-snippets (point-min) (point-max)))))
    (buffer-substring (point-min) (point-max))))
```

Screenshot_20210120_184235

minad:
> I've never really used overlays before, but from the looks of it I'll have to apply an overlay to the region with the display property set to an expansion of the snippet. I'm kinda not sure how to get that expansion though :? ivy-yasnippet just expands the entire snippet and then performs substitutions. I could expand it in a temp buffer but I'm not sure whether all the associated font-lock rules would still be in affect. Is this what you're suggesting?

Yes, maybe using a separate buffer could be an option. If you use a separate buffer you must enable the correct mode there. But let's think a bit about it. Maybe it is better to perform the expansion in-place during preview

mohkale:
> Maybe it is better to perform the expansion in-place during preview

What do you mean? Save point, expand snippet, delete expansion, show overlay.

minad:
> What do you mean? Save point, expand snippet, delete expansion, show overlay with contents of expansion.

Maybe that would be an option. I am not sure about the best way, but I am sure we will figure it out.

mohkale:
@minad 

Okay, the approach I described above does work :-). 

```lisp
(defun consult-yasnippet--foo (template orig-offset)
  "Expand and return contents of TEMPLATE.
ORIG-OFFSET is the distance from "
  (let ((pos (point)))
    (unwind-protect
        (yas-expand-snippet (yas--template-content template)
                            nil nil
                            (yas--template-expand-env template))
      (unwind-protect
          (mapc #'yas--commit-snippet
                (yas-active-snippets (point-min) (point-max)))))
    (let ((end (- (point-max) orig-offset)))
      (prog1 (buffer-substring pos end)
        (delete-region pos end)))))

(defun consult-yasnippet--preview ()
  (let* ((buf (current-buffer))
         (pos (point))
         (region (if (region-active-p)
                     (cons (region-beginning) (region-end))
                   (cons (point) (progn (insert " ") (point)))))
         ;; (region-contents (buffer-substring (car region) (cdr region)))
         (overlays)
         (invisible)
         )
    (lambda (template restore)
      (with-current-buffer buf
        (consult--invisible-restore invisible)
        (mapc #'delete-overlay overlays)
        (cond
         (restore
          (delete-region (car region) (cdr region))
          )
         (template
          (setq invisible (consult--invisible-show))
          (let ((yas-verbosity 0)
                (inhibit-ad-only t)
                (yas-prompt-functions '(yas-no-prompt))
                (orig-offset (- (point-max) (cdr region))))
            (setq overlays (list (consult--overlay
                                  (car region) (cdr region)
                                  'display (consult-yasnippet--foo template orig-offset)
                                  ))))
          )
         )
        ))))
```


But I'm having issues with geting the overlay to appear. It sounds like overlays need to be applied over at least one character and for some reason when I apply it over multiple character the overlay gets truncated. Honestly I'm not sure when I'll have the time to properly understand how overlays work so hopefully I can come back to this later. For now the writing into the buffer approach works pretty well for me.

minad:
Okay, I think I am also fine with modifying the buffer. However you should ensure that the buffer modification status and everything is kept intact. I just reworked consult-keep-lines, see https://github.com/minad/consult/commit/ef08e036e173ed5609c71f7b8395535c068d3ac1. There is the function with-silent-modifications which may be useful for that.

consult-yasnippet.el
```diff
@@ -0,0 +1,140 @@
...
+  (let* ((buffer-undo-list t))                                                  ; Prevent querying user (and showing previews) from updating the undo-history
+    (consult--read
+     "Choose a snippet: "
+     (consult-yasnippet--candidates)
+     :lookup 'consult--lookup-cdr
+     :require-match t
+     :preview (consult-yasnippet--preview)
+     :category 'yasnippet)))
+
+(defun consult-yasnippet-visit-snippet-file (snippet)
```
minad, line 125: Needs docstring

consult-yasnippet.el
```diff
@@ -0,0 +1,140 @@
...
+  (interactive (list (consult-yasnippet--read-snippet)))
+  (yas--visit-snippet-file-1 snippet))
+
+(defun consult-yasnippet (template)
```
minad, line 129: Needs docstring.

consult-yasnippet.el
```diff
@@ -0,0 +1,140 @@
...
+                        (yas-active-snippets (point-min) (point-max)))
+                (setcdr region (- (point-max) orig-offset))))
+            (redisplay)))))))
+
+(defun consult-yasnippet--candidates (&rest body)
+  "Retrieve the list of available snippets in the current buffer."
+  (unless (bound-and-true-p yas-minor-mode)
+    (error "`yas-minor-mode' not enabled in current buffer"))
+
+  (cl-loop for template in (yas--all-templates (yas--get-snippet-tables))
```
minad, line 96: Please rewrite this with mapcar. I am currently not using cl-loop in consult. It is only a stylistic thing.

consult-yasnippet.el
```diff
@@ -0,0 +1,140 @@
...
+          (setcdr region (point))
+
+          (when (and template (not restore))
+            (unwind-protect
+                (consult-yasnippet--expand-template template region region-contents)
+              (unwind-protect
+                  (mapc #'yas--commit-snippet
+                        (yas-active-snippets (point-min) (point-max)))
+                (setcdr region (- (point-max) orig-offset))))
+            (redisplay)))))))
```
minad, line 89: Why is this redisplay needed?

consult-yasnippet.el
```diff
@@ -0,0 +1,140 @@
...
+          ;; We always undo any snippet previews before maybe setting up
+          ;; some new previews.
+          (delete-region (car region) (cdr region))
+          (goto-char (car region))
+          (setcar region (point))
+          (insert region-contents)
+          (setcdr region (point))
+
+          (when (and template (not restore))
+            (unwind-protect
```
minad, line 83: I don't understand this nested unwind-protect.

consult-yasnippet.el
```diff
@@ -0,0 +1,140 @@
...
+  (let* ((buf (current-buffer))
+         (region (if (region-active-p)
+                     (cons (region-beginning) (region-end))
+                   (cons (point) (point))))
+         (region-contents (buffer-substring (car region) (cdr region))))
+    (lambda (template restore)
+      (with-current-buffer buf
+        (let ((yas-verbosity 0)
+              (inhibit-redisplay t)
+              (inhibit-read-only t)
```
minad, line 70: Why inhibit-read-only? We already barf in the beginning? Why inhibit-redisplay?

consult-yasnippet.el
```diff
@@ -0,0 +1,140 @@
...
+
+This starts a `completing-read' session with all the snippets in the current
+snippet table with support for previewing the snippet to be expanded and
+replacing the active region with the snippet expansion.
+
+This function doesn't actually expand the snippet, it only reads and then
+returns a snippet template from the user."
+  (barf-if-buffer-read-only)
+
+  (let* ((buffer-undo-list t))                                                  ; Prevent querying user (and showing previews) from updating the undo-history
```
minad, line 116: Since you disable undo here, it won't work for the final expansion of the snippet. It should be possible to undo this.

minad:
Okay, I added a few more comments. Most important is that undo is still working with preview. Please try to address these. Then I will give it another test and maybe merge it as is or apply some minor adjustments myself.

consult-yasnippet.el
mohkale: Done.

consult-yasnippet.el
mohkale: Done

consult-yasnippet.el
mohkale: No idea :yum:. It was there in ivy-yasnippet so I copied it verbatim.

minad: Yes sure, but this needs a bit of investigation. I will only pull code which is well-understood and explained. I must understand it at least, otherwise I cannot maintain it.

consult-yasnippet.el
mohkale: Again, just [copied and pasted](https://github.com/mkcms/ivy-yasnippet/blob/83402d91b4eba5307f71884a72df8e11cc6a994e/ivy-yasnippet.el#L140-L145) from ivy-yasnippet.

consult-yasnippet.el
mohkale: Taken from [ivy-yasnippet](https://github.com/mkcms/ivy-yasnippet/blob/83402d91b4eba5307f71884a72df8e11cc6a994e/ivy-yasnippet.el#L135-L136).

consult-yasnippet.el
mohkale: I'm not having any issues using evil-undo and undo-tree. And I don't see why there would be any issue. The final expansion doesn't happen in `consult-yasnippet--read-template`, this command just presents snippets as completion candidates and returns the chosen snippet. The final expansion happens at `consult-yasnippet`.

consult-yasnippet.el
minad: Yes, but this does not help me understanding it.

consult-yasnippet.el
minad: Ah okay, let me check this. I may have missed that.

consult-yasnippet.el
mohkale: I try to avoid using mapcar whenever I end up using a lambda with it because of performance concerns (I seem to recall reading somewhere that lambda evaluations due to closures slow down evaluation). But if its a style thing then I hope you don't mind I've changed it to a dolist,push,reverse operation.

consult-yasnippet.el
minad: You are right. This should be good then!

consult-yasnippet.el
mohkale: That's understandable. I kind of assumed a lot of the stuff and logic in there is to deal with edge cases so I didn't think too hard about the ramifications. If you'd like to understand all the nitty gritty stuff, maybe you should ask @mkcms or @conao3, I'm sure they can be of more help.

consult-yasnippet.el
minad: Yes it slows down, but it is marginal in this case. You can safely use a mapcar here.

mohkale: Is what I've currently submitted still within the style guidelines or would you prefer I update it to use `mapcar`?

minad: I don't really like it since it is just a clouded mapcar. I also have such code in consult.el but only where performance matters. However these things are minor nitpicks, I can also change it myself one way or the other if I feel like it. More important for me is to understand what is going on and this has not yet been addressed sufficiently. I will add some more comments to the current version to mark the places which I find hard to follow. I understand that you copied it from ivy-yasnippet, but you could also ask the original author if they are interested in a port to consult

consult-yasnippet.el
```diff
@@ -0,0 +1,153 @@
...
+(require 'yasnippet)
+
+(defun consult-yasnippet--expand-template (template region region-contents)
+  "Expand the yasnippet template TEMPLATE at point."
+  (deactivate-mark)
+  (goto-char (car region))
+  (when (not (string-equal "" region-contents))
+    (push-mark (point))
+    (push-mark (cdr region) nil t))
+
```
minad, line 47: Needs a comment/explanation: Why are marks pushed?

consult-yasnippet.el
minad: Yes, maybe @mkcms or @conao3 are interested in helping with a port of ivy-yasnippet to consult.

consult-yasnippet.el
mohkale: I think it's related to how `yas-expand-snippet` works. If the region is active it seems to overwrite that region with the snippet expansion. When region-contents isn't empty, there was a region active at the start of `consult-yasnippet` so here we restore that region so `yas-expand-snippet` replaces any previous expansions (or the original `region-contents`) with the current snippet template.

I can try adding a comment to that affect.

consult-yasnippet.el
```diff
@@ -0,0 +1,156 @@
+;;; consult-yasnippet.el --- Provides the command `consult-yasnippet' -*- lexical-binding: t; -*-
+
+;; Author: Mohsin Kaleem, Consult and Selectrum contributors
```
minad, line 3: Here the original authors should be acknowledged.

consult-yasnippet.el
```diff
@@ -0,0 +1,156 @@
...
+  "Retrieve the list of available snippets in the current buffer."
+  (unless (bound-and-true-p yas-minor-mode)
+    (error "`yas-minor-mode' not enabled in current buffer"))
+
+  (let ((annotation)
+        (cands))
+    (dolist (template (yas--all-templates (yas--get-snippet-tables)))
+      (setq annotation (concat " ["
+                               (propertize (yas--template-key template)
+                                           'face 'font-lock-type-face)
```
minad, line 104: consult-key face?

consult-yasnippet.el
mohkale: Oooh, I wasn't sure what to put there. Thanks for letting me now. :smiley:.

consult-yasnippet.el
mohkale: Roger.

consult-yasnippet.el
minad: I think it is fair to write `<name of original authors>, Mohsim Kaleem, Daniel Mendler`. Then we should also add a comment that it is derived from ivy-yasnippet and a link to ivy-yasnippet in the commentary section. Maybe you already have that. I haven't looked.

consult-yasnippet.el
mohkale: All right, switched to `mapcar` to save you the hassle of changing it sometime down the line.

I'll open an issue on `ivy-yasnippet` and ask whether they'd be willing to help out :-).

consult-yasnippet.el
mohkale: I don't really care if my names in their but I agree we should say some of it was derived from `counsel-yasnippet`. Looks like the venerable alphapapa advises putting a [credits](https://github.com/alphapapa/emacs-package-dev-handbook#template) in the package docstring to reference helpful contributors.

I've added that now. How does this sound?

```lisp
;; This package would not have been possible without the ivy-yasnippet[1]
;; package which showed me how to access and preview yasnippet templates
;; through a `completing-read' interface.
;;
;;  [1] https://github.com/mkcms/ivy-yasnippet
```

minad: I think it is better to write something like "This package has been derived from ... by ..." and some link. It is not as if we just looked up some tricks from there. It is clearly a derivative.

mohkale: Plain and simple?

```lisp
;; This package has been derived from the ivy-yasnippet[1] package.
;;
;;  [1] https://github.com/mkcms/ivy-yasnippet
```

minad: Yes something like this but the author names should also go there. It is not okay if we add our names but omit the original authors ;) But look, these things are not crucial. I will fix this afterwards if necessary. I don't want to over discuss things.

mohkale: I added it at the top just before your name like you sugggested before... would you like me to add it in both places?

consult-yasnippet.el
conao3: I am the maintainer of ivy-yasnippet, but I am the maintainer of the current state and will not change the code unless there is a problem.
In that respect, I don't know the intention of this redisplay either. You can try removing it and see if it causes any problems in production.

(Actually, I became a fan of ivy-yasnippet as soon as I found the package, and I ran for maintainer in the absence of a maintainer, but there was a fatal glitch and I am not using it now.
 The bug is that if the snippet contains an Elisp that asks for input, when previewing the snippet in the buffer, it will override ivy-yasnippet and ask the user for input.
 If it didn't have this glitch it would be one reason for me to move from ivy to consult)

minad: @conao3 Thank you for chiming in here. Can you point us to a snippet which has this glitch? 

@mohkale It seems we have to deconstruct the code ourselves before it is ready for Consult. The presence of a known problematic glitch is reason enough to not merge it in the current form.

mohkale: @minad 

I don't really think there's a way around that on our end :/. Maybe we can ask the yasnippet maintainer to expose a function to expand a snippet, without any elisp being evaluated. Otherwise I guess we can leave this in limbo until we figure out a good way to handle it.

Thanks for all the time you've invested in this PR :+1:.

conao3: Any snippet has `yas-completing-read`. Below snippet is a sample snippet I have.
```
# name: begin_src
# key: begin_src
# --
#+begin_src ${1: $$(yas-completing-read "Prompt: "
(append
 '("C++")
 (mapcar
  (lambda (elm)
    (substring (symbol-name elm) 3))
  (cl-remove-if-not
   (lambda (elm)
     (string-prefix-p "ob-" (symbol-name elm)))
   features))))}
  $0
#+end_src
```

For detail reprostep, please see https://github.com/mkcms/ivy-yasnippet/issues/10 .

minad: This does not sound like a deep problem to me. We can just cl-letf this yas read function locally returning some placeholder and go on.

minad: Is there only yas-completing-read or also yas-read-number etc? Which functions would require special handling? Alternatively we could just write our own preview expander assuming it is not too complicated.

mohkale: My concern is that if you can execute arbitrary lisp then nothing stops someone from running `completing-read` or any other blocking commands. Perhaps that's out of the concern for this particular problem. I'll try `cl-letf` and returning empty strings for everything.

Incidentally I can only see 3 yas commands with complete in them:
1. `yas-completing-read`
2. `yas-completing-prompt`
3. `yas/completing-prompt`

mohkale: Alright I've wrapped `yas-expand-snippet` in a `cl-letf` block such that `yas-completing-read` always returns an empty string. I'm now no longer getting any `completing-read` popups when previewing the snippet with `consult-yasnippet`.

minad: Okay, so that could be an option. But it wouldn't be 100% safe. It would be better to not execute elisp at all. How complicated is the expansion code? Maybe propose upstream to yasnippet a function which returns a snippet as previewable string. After it has been added there we can proceed here.

minad: I looked into yasnippet. Unfortunately it seems not to get much maintenance these days. For us, I see the following potential ways going forward, sorted by required effort:

1. Give up on the ivy-yasnippet/consult-yasnippet idea if it cannot be implemented reliably
2. Accept the unsafe solution (but this is not something I am eager to do honestly)
3. Write our own expander code here, which would work better in our scenario
5. Look for an alternative yasnippet-compatible/like package and support that instead (I think the alternatives are tempo, skeleton and yasnippet, maybe more?)
6. Try to contribute to upstream yasnippet, maybe they are even looking for a maintainer

minad: Is something like this good enough?

~~~ elisp
(defun my-expander ()
  (interactive)
  (message "%S"
           (mapcar (lambda (x)
                     (replace-regexp-in-string
                      "\\$[0-9]+" ""
                      (replace-regexp-in-string
                       "\\${[0-9]+:\\([^}]*\\)}" "\\1"
                       (yas--template-content (cdr x)))))                   
                   (car (yas--templates-for-key-at-point)))))
~~~

And then just show the obtained templates as an overlay as I originally suggested.

mohkale: @minad

I don't really have an issue with 2, I've been using ivy-yasnippet for years (exactly 2 :yum:) and I've never run into the issue that @conao3 mentioned. Although of course seing as the brunt of the maintanance work will be on you I understand your hesitation to do so.

If you'd like to go the 1 route that's fine. I can just add what I've got to the wiki or somewhere else and users will have to maintain it themselves. This is probably the best route for anything that doesn't directly conform to consults philosophy.

I find 3 interesting, but honestly not sure of the complexity or ramifications of it. If you can come up with a single function that takes a snippet template then returns a string then that's great :smile:, otherwise I'm not sure I'll be able to do so. In regards to using overlays there's still the issue with syntax highlighting rules when expanding snippets in memory or a temp-buffer. End of the day we need some way to apply the current buffers font-lock rules to it, the easiest of which would be to just insert it. (Maybe there's a function that can fontify a string, I'm not sure, I haven't found it yet).

As for 4, just a short while back yasnippet got a new patch allowing regexp keys (I believe) so it is still actively developed. I'm sure if you submitted a PR @joatavora would consider it :smile:.

minad: > I find 3 interesting, but honestly not sure of the complexity or ramifications of it. If you can come up with a single function that takes a snippet template then returns a string then that's great smile, otherwise I'm not sure I'll be able to do so. In regards to using overlays there's still the issue with syntax highlighting rules when expanding snippets in memory or a temp-buffer. End of the day we need some way to apply the current buffers font-lock rules to it, the easiest of which would be to just insert it. (Maybe there's a function that can fontify a string, I'm not sure, I haven't found it yet).

I agree. Overlays are a bad idea because of the fontification. I forgot this again. 

But I still think option 3 is the way to go. The way Consult is written as of now is such that previews are always "benign", they do nothing unexpected and must be solid under all circumstances. I think we should just use this simple string substitution function from above or something slightly more sophisticated. Maybe we can reuse some parser infrastructure from yasnippet? 

Would you like to continue to work on that or do you expect me to do this? I am willing to maintain this package as part of Consult. But I am not very interested in working out all the details. I already spent a bit of time on this PR, so it is obvious that I would like to have it here.

minad: The function I pasted above is a good starting point. But I think in order make this reliable we need something more sophisticated obviously.

~~~ elisp
(defun my-expander ()
  (interactive)
  (message "%S"
           (mapcar (lambda (x)
                     (replace-regexp-in-string
                      "\\$[0-9]+" ""
                      (replace-regexp-in-string
                       "\\${[0-9]+:\\([^}]*\\)}" "\\1"
                       (yas--template-content (cdr x)))))                   
                   (car (yas--templates-for-key-at-point)))))
~~~

mohkale: @minad

If you'd like to get started on this approach please do so. Otherwise I'm not sure when but I'll probably come back to this PR and try my hand at it when I feel the need for it. Atm the current implementation works for my needs so I'm not really in a rush to improve it IYGWIM. Regardless thank you for all the help you've given with this PR, I've learnt quite a lot :smile:.

minad: Okay, then I guess I close this PR for now. Maybe you or someone else will pick it up again in the future.

PR #175, opened 2021-01-22
First draft: consult-hippie

minad:
@protesilaos This is a first draft for a `consult-hippie` command taken from here https://gist.github.com/JohnLunzer/7c6d72a14c76c0a3057535e4f6148ef8. I believe it originated in the Emacs wiki (TODO: Check license). I am not fully convinced by this command, since it does not seem like a significant improvement over `dabbrev-completion`. But I am interested in feedback!


protesilaos:
> I am not fully convinced by this command, since it does not seem like a significant improvement over dabbrev-completion. But I am interested in feedback!

Thanks!  I will pull in the changes and start testing.

protesilaos:
Very quick first comparison between `dabbrev-completion` and `consult-hippie`:

```
Dan<dabbrev-completion>

Daniel's
Daniel
```

And:

```
Dan<consult-hippie>

Daniel Mendler's Consult is a welcome addition to the ecosystem of
Daniel
Dance
Dancing
Dangerous
Daniel's
```

I will need more time testing this with some real world situations.  The only observation right now is whether a list could be broken into sublists.  So this:

```
Daniel Mendler's Consult is a welcome addition to the ecosystem of
```

Becomes something like:

```
Daniel Mendler's Consult is a welcome addition to the ecosystem of
Daniel Mendler's Consult
Daniel Mendler's
```

Current hippie config:

```elisp
(setq hippie-expand-try-functions-list
        '(try-expand-list
          try-expand-line
          try-expand-dabbrev-all-buffers
          try-expand-dabbrev
          try-expand-dabbrev-from-kill
          try-expand-all-abbrevs
          try-complete-lisp-symbol-partially
          try-complete-lisp-symbol
          try-complete-file-name-partially
          try-complete-file-name))
```

minad:
Sure, I could manually split the candidates in `consult-hippie`. However it would be more principled if you write a special `try-*` function which produces such candidates.

About the examples you showed, I am still unsure about the use case case. For example if you type "Dan" and want to complete to "Dance", this is not helpful at all, since you need three key presses of `<down>` to arrive at the desired candidate. It is more useful with longer candidates obviously. For example when completing function names or long words, e.g., after typing "consult--". Maybe very short candidates or candidates which are only few characters longer than the original input should not even be shown.

protesilaos:
> About the examples you showed, I am still unsure about the use case case. For example if you type "Dan" and want to complete to "Dance", this is not helpful at all, since you need three key presses of <down> to arrive at the desired candidate. It is more useful with longer candidates obviously. 

You are right!  Those are not offering an obvious benefit.---and is why I would need more time to test this in real scenaria.  Back when I was using hippie-expand more regularly, what I liked the most was completing longer stretches of text, like:

```elisp
`(con<consult-hippie>

(consult-view ((,class :inherit bold :foreground ,fg-special-warm)))
(consult-preview-line ((,class :inherit modus-theme-special-mild)))
(consult-preview-error ((,class :inherit modus-theme-intense-red)))
(consult-preview-cursor ((,class :inherit modus-theme-intense-blue)))
...
```

The reason I stopped using it was because it did not offer completion, so it was more miss than hit.

By the way, there is a bug currently.  If the text starts with a parenthesis, then that gets added again.  For instance:

```elisp
`(con<consult-hippie>
`((consult-view ((,class :inherit bold :foreground ,fg-special-warm)))
```

minad:
> Back when I was using hippie-expand more regularly, what I liked the most was completing longer stretches of text.

This is exactly where I see the largest benefit of dabbrev and hippie. 

> By the way, there is a bug currently. If the text starts with a parenthesis, then that gets added again. For instance:

Okay, this whole thing certainly needs a good cross check. Basically I only took the verbatim version from the gist above with some minor improvements to make it faster (inhibit modification etc). Could it be that the parentheses are introduced by paredit/smartparens or some other similar package?

protesilaos:
> Yes, this is exactly where I see the largest benefit of dabbrev and hippie.

Let's see what the tests reveal.  By the way, have you notified the people on the wishlist thread?  Maybe someone else would like to give it a try.

> Could it be that the parentheses are introduced by paredit/smartparens or some other similar package?

I don't use those or anything similar that auto-inserts characters.

minad:
> Let's see what the tests reveal. By the way, have you notified the people on the wishlist thread? Maybe someone else would like to give it a try.

Let's also summon @clemera, @hmelman and @oantolin here, since they took part in the hippie/dabbrev discussion in #6.

hmelman:
I'll play with this.  FWIW, when I used `hippie-expand` I tended to use it for symbol names, not for multi-word strings.  Even in this case it's useful to automagically get the case correct or to deal with more difficult to type punctuation. Also I believe `hippie` enabled the use of text in other (visible) windows/buffers before `dabbrev` got that capability.  So it was very convenient to look up the definition or documentation of something in another emacs window and then use hippie to complete that in the current buffer.  This was fantastic before there was smart completion-at-point or LSP enabled "IntelliSense" in non-elisp buffers.

Now I use it in `markdown-mode` and have switched to `dabbrev` because it completes words better when some instances have markup at their beginning.  I haven't been able to make `hippie` work as well in that case and it's this that led me to find out that the `hippie` functions have their own `dabbrev` implementation and don't call through to `dabbrev.el`.

manuel-uberti:
Just for reference, Helm offers [helm-dabbrev.el](https://github.com/emacs-helm/helm/blob/master/helm-dabbrev.el) which covered all my use cases for completion most of the times. For the few times it didn't get it right, Company helped.

minad:
@manuel-uberti What makes this helm-dabbrev special? What does it do differently than `dabbrev-completion` for example. It seems to bring its complete own dabbrev implementation. Something like this is mostly out of the question for Consult. I would want to reuse either dabbrev or hippie.

manuel-uberti:
As I said: just for reference. :)

`dabbrev` suits me well enough these days.

protesilaos:
> hmelman: Now I use it in markdown-mode and have switched to dabbrev because it completes words better when some instances have markup at their beginning. I haven't been able to make hippie work as well in that case and it's this that led me to find out that the hippie functions have their own dabbrev implementation and don't call through to dabbrev.el.

> manuel-uberti: dabbrev suits me well enough these days.

I should add that I also am a full time dabbrev user (no company-mode or any pop-ups).  What I would want from a dabbrev-completion-like command is the ability to complete longer stretches of text more efficiently.  For example, to complete a common pattern in my buffer, say, "Protesilaos Stavrou", with `dabbrev-expand` I must do this:

```
Pro <dabbrev-expand> SPC <dabbrev-expand>
```

My vision for `consult-hippie` (or similar) was that it would basically eliminate the SPC and treat common occurrences of words/symbols as a "phrase" construct.  Basically a greedy variant of dabbrev.  Though I understand this may well be outside the scope of Consult, in which case `consult-hippie` may not be of much value over `dabbrev-completion`.

minad:
Comment by @protesilaos in https://gitlab.com/protesilaos/modus-themes/-/issues/155

I guess we should move the consult-hippie discussion there when you feel like it.  Perhaps hippie-expand is not the right tool for the job and maybe we need a new (or existing?) tool that provides a superset for dabbrev or something like dabbrev but with wider reach.  While dabbrev matches words/symbols, this one should be able to match longer patterns on lines.  So when the user types "Eur" it shows candidates like:
European Council
European Central Bank
European Coal and Steel Community
European Border and Coast Guard Agency
European body for the enhancement of judicial co-operation
...
Performance penalties could be mitigated by minimum characters and those kind of things you already know.

minad:
I also asked myself if it would be appropriate to create a new tool, hippie done right. I have at least these criticisms regarding hippie:

* Hippie does not reuse dabbrev, I assumed it does but it reinvents its own version.
* Hippie mutates the buffer directly and does not offer an accessible API for completion.
* Since hippie seems to allow arbitrary expansions, it is too uncontrolled and not reusable.

The idea would be that the package should offer pure functions returning lists of candidates such that it is a better fit for completion. I think this would also require less code since you would only have a single point where the expansion is applied in the target buffer. If someone is looking for a new small Emacs project, this could be it - if it does not already exist :)
The expansions you propose seem useful, but ideally I would like something which is configurable like hippie, but more flexible such that arbitrary backends can be plugged and only useful expansions are generated. Maybe this is something which should be implemented as an autocomplete/company backend?

oantolin:
It does sound like a company backend would be appropriate. Is there a way to not have company complete automatically, but only when I run a command. I love completion, but don't particularly enjoy **auto**completion.

minad:
In my configuration company shows this popup and then I have to complete explicitly? Seems to be pretty much the default configuration.

minad:
What I dislike about company is that only one backend is active at a time. I don't understand why this has to be the case.

protesilaos:
I think those are fair criticisms and a company backend would be interesting to see.  I also remember company would perform manual operations by default, though it could be configured to auto-complete when only a single candidate was available.

Can company somehow work from the minibuffer instead of its typical overaly/child-frame?

protesilaos:
hippo is a neat name, by the way.

oantolin:
> In my configuration company shows this popup and then I have to complete explicitly? Seems to be pretty much the default configuration.

Yes, but the popup pops up automatically as you type. I want it to only appear when summoned. Do you know if I can configure it to do that?

oantolin:
Forget company! Can this hippo be implemented as a completion at point function, so it participates in the normal `C-M-i`?

oantolin:
I think I made a mistake: hippo is you name for the multiple backend thing. I claim that is just `completion-at-point` and `completion-at-point-functions` is the list of backends.

And what I should have said instead of hippo, is that @protesilaos's "complete longer streches of text" could probably be provided as a function to add to `completion-at-point-functions`.

minad:
> Yes, but the popup pops up automatically as you type. I want it to only appear when summoned. Do you know if I can configure it to do that?

No idea if there is an option. I guess there is. I want it to popup automatically. For me there a no big downsides. It does not introduce window bumpiness like the minibuffer.

> Forget company! Can this hippo be implemented as a completion at point function, so it participates in the normal C-M-i?

Sure that would be good too. Actually the hippo should allow it to be used for both company and completion at point. This is something I like about dabbrev-expand/hippie-expand, that they need their own key. These things should be unified.

oantolin:
> I want it to popup automatically. For me there a no big downsides. It does not introduce window bumpiness like the minibuffer.

True but it does introduce a different kind of bumpiness that's even worse: it happens right at the spot I'm looking at!

clemera:
I have `(setq company-idle-delay nil)` in my config, this will stop the auto popup

protesilaos:
Is not company implementing a "capf" backend for completion-at-point-functions?  Shouldn't those be independent of the UI?  Hence something that could also work with `C-M-i`.

oantolin:
Yes, that's right, @protesilaos.

So to recap, I believe hippo already exists: it is `capf`! The backends are functions added to `completion-at-point-functions`, and a company backend that uses them also already exists: `company-capf`. A non-company UI to use them also exists: `completion-at-point`.

The only thing that does not already exist, as far as I know, is a completion at point function that completes longer phrases, as @protesilaos wants.

minad:
Right :+1: we only need better completion at point functions!

minad:
Just for clarification - these completion-at-point-functions are all tried in order and combined or is only the first one used which returns non-nil?

oantolin:
I think only the first one returning non-nil is used: `completion-at-point` runs `completion-at-point-functions` using `run-hook-wrapped`, which stops at the first non-nil result. The wrapper it uses is a little complicated, but basically checks that the function returns a completion table with at least one match to the text at point.

minad:
Yes, I just looked there too and I came to the same conclusion as you. I would like the hippo completion to be able to combine multiple backends, since completion-at-point does not support it by itself.

oantolin:
I wonder if `company-capf` uses the first non-nil return value or combines all of them. Probably the first, to be compatible with `completion-at-point`.

oantolin:
> I have `(setq company-idle-delay nil)` in my config, this will stop the auto popup

With this setting when do you get the pop up, @clemera?

clemera:
@oantolin You need to call `company-complete`

oantolin:
A small first step towards the phrase completer would be to reimplement `dabbrev` as a completion at point function.

Then, upon passing to phrases, one thing that would have to decided is how to decide where the phrase ends. That seems tricky to get right. Have you thought about that, @protesilaos? From your "European" examples, it seems that stopping the phrase before 2 (or 3?) consecutive lower case words would catch some of the intended phrases.

protesilaos:
> oantolin: Then, upon passing to phrases, one thing that would have to decided is how to decide where the phrase ends. That seems tricky to get right. Have you thought about that, @protesilaos? From your "European" examples, it seems that stopping the phrase before 2 (or 3?) consecutive lower case words would catch some of the intended phrases.

Yes, this is tricky and I cannot claim to have any concrete action plan.  Perhaps the simplest way is to look forward up to N consecutive targets at a time?  Say three or four words max.  A natural stop would be punctuation (perhaps major-mode specific?).  Another could be line breaks (are those reliable?).

I understand this may not prove realistic.  No worries though.

minad:
Closing this, since `consult-hippie` is probably not going to happen. It would be nice if someone comes up with an improved completion-at-point-based expander, with dabbrev or other backends.

I have a list of potential small Iceberg-related projects/experiments, which may get attention at some point. Some are just (failed?) experiments.

* better dabbrev/expand capf (this issue)
* simpler completion at point (https://github.com/minad/corfu)
* uchronia, input history, repeating commands (https://github.com/minad/uchronia)
* fuzzy filtering outside emacs (https://github.com/minad/consult-fuzzy)
* isearch-mb (not mine, by @astoff, https://github.com/astoff/isearch-mb, see #209)

astoff:
I wasn't aware of this thread, so I'll add a couple of comments.

> Is there a way to not have company complete automatically, but only when I run a command? I love completion, but don't particularly enjoy autocompletion.

When I manually trigger a completion, I want to immediately start _filtering_ candidates, as in Selectrum completion in region. Company doesn't work that way, presumably because it's designed for autocompletion, so it has to be less intrusive. To me, the only problem with Selectrum completion in region to me is the visual presentation: I want the completions to somehow appear near the completion subject. After checking out some alternatives, I still think Ivy has the best approach to completion in region, visually (unfortunately it seems a bit buggy sometimes). I guess that's what corfu is/was supposed to be. I wonder if that kind of UI is within Selectrum's scope, @clemera.

> Just for clarification - these completion-at-point-functions are all tried in order and combined or is only the first one used which returns non-nil?

Company has the notion of grouped backends, see at the end of the docstring to `company-backends`.

It should be possible to write a capf backend that mixes the results of a list of backends. But maybe this feature should be built into capf itself, as with company.

minad:
> Company has the notion of grouped backends, see at the end of the docstring to company-backends.

Yes, I am aware. But I think for capf this does not apply. I think company backends are actually not that useful. The company maintainers are moving more and more towards capf it seems. LSP also uses capf exclusively. But still if capf is used, it could be useful to combine multiple capf backends. I don't think this is possible as of now.

> It should be possible to write a capf backend that mixes the results of a list of backends. But maybe this feature should be built into capf itself, as with company.

Agree, it should either be builtin or there should be some mixer backend.

> I want the completions to somehow appear near the completion subject. After checking out some alternatives, I still think Ivy has the best approach to completion in region

Do you mean posframe? Selectrum does not support that (yet?), but there is emacs-mini-frame which comes close. But afaik this is also buggy. For me the best visual presentation is company for capf. It is unobtrusive and fast because of overlays. The only problem is that it is hard to configure well and one has to jump through hoops if one wants it to behave similarly to completion. See this comment (https://github.com/company-mode/company-mode/issues/1061#issuecomment-771635807) for the configuration I am using now.

astoff:
> Do you mean posframe?

Ivy supports several display methods, but what I'm referring to (Ivy's default completion in region replacement) is based on overlays, quite similar to corfu in fact.

With your `,` command, which I would rather bind to `SPC`, and some of your tweaks, company probably also works as I would want it to. Still, I think that my expectations for completion in region are pretty sensible and shouldn't require that much configuration. I like the corfu idea.

minad:
> With your , command, which I would rather bind to SPC, and some of your tweaks, company probably also works as I would want it to.

Yes, I originally tried SPC but this does not work well in combination with shell completions. It is a bit more compatible with my muscle memory this way. I use TAB cycling for completion and only if there are many candidates I use the completion system style.

protesilaos:
> Closing this, since consult-hippie is probably not going to happen. I consider hippie/dabbrev to be dead-ends in the light of capf.

I think this is the right course of action.  Thanks again for your efforts!

> simpler completion at point (https://github.com/minad/corfu)

Any short description of how it differs from company in terms of scope?  I have drawn some inferences from this discussion, but need to be sure.

Also, Corfu is a nice place!

minad:
"Completion Overlay in Region FUnction" is a small hack I wrote as an experiment, it should not be used for anything. I only experimented with this because I got annoyed when I reconfigured company recently. I think if this would be written properly it would just be almost like company, but only focused on capf and would behave a bit more like a familiar completion system. My problem with company is that it is hard to configure the way I want it and it seems to be overly complex. Prot, I think I've read in your configuration that you don't get too much value out of Company, I guess that is my problem too. But I do like the unobtrusive overlay interface in contrast to opening either a full completion buffer or a vertical minibuffer ui like Selectrum. The way I've configured company is such that it cycles for a small number of candidates and for a larger number opens the selection overlay.

protesilaos:
> I think if this would be written properly it would be almost like company, but only focused on capf and would behave a bit more like a familiar completion system.

Ah yes, that makes sense.  A capf that combines multiple sources, as mentioned in this thread, would be the way to go and was my initial hope for a hippie-style command.

> Prot, I think I've read in your configuration that you don't get too much value out of Company, I guess that is my problem too.

I tried Company for several weeks and never felt that it was helping me type faster.  For single words I found that it was almost always quicker for me to type them out than insert 2-3 characters, wait for the pop-up, check the list, pick an item.  Besides, dabbrev usually does the right thing in such a case.

To make Company more effective for my case, I had to push its delays close to zero (I normally add higher delays), at which point it became a distraction.  Usually I type without pausing, but sometimes I pause for a split second even though I do not intend to use Company.  So having a constant pop-up near the point adds too much visual noise.  This is particularly annoying when I write some long form text (i.e. quite often).

For me checking the minibuffer for in-buffer completion is not a problem because I am already going into the mental mode of being more deliberate in order to pick a candidate.  And my natural tendency is to check where I always do.

Long story short: any pop-up facility should be optional.

> But I do like the unobtrusive overlay interface in contrast to opening either a full completion buffer or a vertical minibuffer ui like Selectrum.

I think this is true for most users, so I understand why my expectation for an option is not easy to satisfy.

dgutov:
> What I dislike about company is that only one backend is active at a time. I don't understand why this has to be the case.

Because when I have a "smart" backend configured and it misbehaves for some reason, or it actually runs out of completions for a given prefix, an automatic failover to `company-dabbrev-code` is not what I expect it to do. "No completions" is a valid answer.

But if one wants to browser what the rest of backends offer, they can `M-x company-other-backend`, once or several times.

It's true that `company-capf` doesn't offer the same courtesy to c-a-p-f elements (not yet, at least). Someday, when c-a-p-f API is powerful enough (and hopefully nicer as well), we can drop the `company-backends` stuff and simplify things.

minad:
@dgutov Makes sense! I hope company is moving more into the direction of capf. I saw that you are proposing in company PRs that people should rather use capf instead of company specific backends.

dgutov:
Often, yes. Because the core maintainers like that direction, and using a frontend-agnostic API is just good sense.

It is missing a number of features, though. So I can't just deprecate `company-backends` in favor of it yet.

minad:
If case, someone of you is interested in this - I hacked a bit on [Corfu](/minad/corfu), which can be seen as a Vertico counterpart for completion in region. Icomplete conflates completion in region and minibuffer completion, which I believe is not the best idea. A different UI may be preferred for each use cases and it simplifies things in the code as well. (Icomplete has a lot of small helper function which all have to check if it is completion in region or minibuffer completion). But both Vertico and Corfu follow mostly the design of Icomplete.

protesilaos:
Thanks! I just scanned its README:

> Since this package is part of GNU ELPA, contributions require copyright assignment to the FSF.

Has this been done?

minad:
> Has this been done?

No, not yet. I intend to submit it there soon. It needs a bit more testing as of now. It would also be good to see if it works for other people. But I guess most people are satisfied with Company. For me this is another experiment in what can be done with less.

It is also a test bed to see how all this completion-in-region stuff works internally. You may have seen the long discussion regarding completing-read on the mailing list and there I also discussed with @dgutov about asynchronous completion, also in the context of in-buffer completion. Now I understand a little bit better how that works. In-buffer completion is a lot messier than minibuffer completion, in particular since the state is not so nicely encapsulated as in the case of the minibuffer. Since the minibuffer opens a recursive editing session, every buffer-local variable there is encapsulated automatically.

I am sure you are aware that Icomplete also supports in-buffer completion. I am not very fond of that technically since Icomplete conflates completion-in-region and minibuffer-completion. And then I prefer the Company UI for completion in region.

There exists a variety of methods now:

* Use default completion for completion-in-region with `*Completions*` buffer
* Transfer completion in region to the minibuffer via `consult-completion-in-region`, `selectrum-completion-in-region` or something else
* Completion in the buffer with a popup via Corfu
* Use Icomplete in-buffer completion with an overlay  similar to Corfu)
* Completion in the buffer with a popup via Company (this is not strictly the `completion-in-region-function`, it uses its own entrypoint or time-based trigger)
* And more?

But now that we have all these methods I am still not sure what the best solution  Probably it is a combination.

1. When in-buffer completing in the minibuffer, then use default `*Completion*`
2. When in-buffer completing in a normal buffer, use Company or Corfu
3. Maybe there are also cases where one wants to transfer to the minibuffer instead of using `*Completion*`?

I am currently using 1+2, by enabling `vertico-mode` in `prog-mode/eshell-mode`, using default completion for the rest.

protesilaos:
>     Has this been done?
>
> No, not yet. I intend to submit it there soon. It needs a bit more
> testing as of now. It would also be good to see if it works for other
> people. But I guess most people are satisfied with Company. For me this
> is another experiment in what can be done with less.

Thank you for the explanation!  I do follow the discussions on
emacs-devel, trying to learn whatever I can in the process.

Just to note that if/when Corfu is added to GNU ELPA I can help you with
syncing the git repos.

-- 
Protesilaos Stavrou
protesilaos.com

minad:
> Just to note that if/when Corfu is added to GNU ELPA I can help you with
syncing the git repos.

Thank you for the offer and for your work and help with ELPA! I will come back to you when it lands there. In the meantime, you may want to give the package a quick try - see if it does what one would expect from it in your setup? But I don't want to steal your time, I know you prefer a different kind of UI :)

jaor:
> No, not yet. I intend to submit it there soon. It needs a bit more
> testing as of now. It would also be good to see if it works for other
> people. But I guess most people are satisfied with Company. For me this
> is another experiment in what can be done with less.

@minad, i for one am already trying corfu, talking as a long-time
company user.  for me, it's doing for company what vertico does for
selectrum, i.e., giving me 90% of the functionality i care about with
10% of the code (so that it's not only easy but really fun providing the
10% bits i might miss... or discovering i don't need them, or making
them work /exactly/ the way i wanted).

i particularly like the way it integrates with orderless out of the box,
company was giving me headaches in that area, and, it being big and
complex, i didn't bother to investigate proper fixes.

protesilaos:
On 2021-04-15, 13:02 -0700, Daniel Mendler ***@***.***> wrote:

> In the meantime, you may want to give the package a quick try - see if
> it does what one would expect from it in your setup? But I don't want
> to steal your time, I know you prefer a different kind of UI :)

Yes, I do prefer another UI, though that does not prevent me from trying
things.  And maybe I will change my mind---it happens all the time.

At any rate, I already added a comment on its issue tracker.  Though I
have been awake since ~4 so I must go to bed soon.  I will give it a try
some time tomorrow and update you accordingly.

-- 
Protesilaos Stavrou
protesilaos.com

minad:
Have a good night then! I am looking forward to your comments.

astoff:
A quick visual impression about Corfu: the popup width changes a lot as you type, because it's always as narrow as possible to fit the visible candidates. Maybe it would look nicer if there was a minimum width.

Another comment that is more related to the completion at point API than Corfu itself: Is it possible to support the situation where a completion backend (i.e., a `completion-at-point-functions` element) already returns a filtered list of candidates, and you want to disable any further filtering? Typing more should call the completion backend again for a fresh candidate list. The use case for this would be special fuzzy completions which are not based on just the candidate text. For example, in TeX, typing `\cite{partsoftheauthorortitle<TAB>` could complete to `\cite{mycitation}` (an example is shown in the GIF [here](https://github.com/astoff/digestif)).

minad:
> Maybe it would look nicer if there was a minimum width.

Good idea!

> already returns a filtered list of candidates, and you want to disable any further filtering? Typing more should call the completion backend again for a fresh candidate list. The use case for this would be special fuzzy completions which are not based on just the candidate text.

This should somehow work already, but it could be broken due to me handling the `:exit-function` incorrectly. If you press TAB, select an candidate via up/down and RET it should complete to the selected candidate and then call the `:exit-function` for expansion. And even when you enter more filtering strings these should be replaced by the candidate.

astoff:
> but it could be broken due to me handling the :exit-function incorrectly.

One test case for you can be Eglot expanding a yasnippet when selecting a candidate. You would have to find a language server that provides snippets (as Digestif, for example) to check this. But I can already report to you that this just works. :tada:

The fuzzy completions, on the other hand, do not...

minad:
> One test case for you can be Eglot expanding a yasnippet when selecting a candidate. You would have to find a language server that provides snippets (as Digestif, for example) to check this. But I can already report to you that this just works. tada

Great, I am glad that this works. You are the first one to test this. I didn't test with lsp+elglot up to now :smile: And yes, yasnippet was on my list to try regarding the expansion.

> The fuzzy completions, on the other hand, do not...

What do you mean exactly by that? When I type something like "prefix f u z y" and then up/down and RET, the whole input is replaced by the candidate you selected. And then expansion should also work.

Above you wrote about "Typing more should call the completion backend again for a fresh candidate list.". I am doing that by calling `completion-all-completions` again, which should call the backend and then pass through the filtering as configured.

Do you mean that the `completion-at-point-function` should be called again, starting the whole process all over on each keypress? One could probably do something like this but it would be a significant deviation from how things are now.

minad:
> The use case for this would be special fuzzy completions which are not based on just the candidate text. For example, in TeX, typing \cite{partsoftheauthorortitle<TAB> could complete to \cite{mycitation} 

Okay, after rereading again what you wrote here, this is what you mean. I got confused by the term "fuzzy completion".  Having candidates returned which are not matching the original input is incompatible with how completion styles work. I have to think how this is supposed to work. Company-CAPF probably does not pass the returned candidates through the completion styles at all and simply shows everything directly.

dgutov:
> Company-CAPF probably does not pass the returned candidates through the completion styles at all and simply shows everything directly.

Perhaps I misunderstood you, but `company-capf` goes through `completion-all-completions`, so it uses the completion-styles mechanism. Which also makes fuzzy-matching-on-the-server difficult.

minad:
@dgutov I speculated and it was wrong. Company goes through `completion-all-completions` which is the right thing to do. 

However it seems one is often required to use completion styles which pass a prefix to the backend, e.g., as you say when filtering on the server. In this case one is required to use the `basic` completion style.

I was wondering if one could support Orderless completion with Company or Corfu by compiling the first word to a prefix and then filter only using the remaining words on the side of Emacs inside the completion style. I made an issue to keep this in mind (https://github.com/minad/corfu/issues/5). cc @oantolin

dgutov:
`completion-category-defaults` and `completion-category-overrides` (user-defined) decide which completion styles are used, based on the _category_ of the completion table.

Whether completion is "in region" or not, doesn't enter the equation, but I suppose you could override one of these variables with a let-binding in your `completion-in-region-function`.

minad:
> Whether completion is "in region" or not, doesn't enter the equation, but I suppose you could override one of these variables with a let-binding in your completion-in-region-function.

Yes, this is what I was suggesting.

astoff:
> However it seems one is often required to use completion styles which pass a prefix to the backend, e.g., as you say when filtering on the server.

I'm not sure I understand what you meant here, but let me clarify a bit how completion requests work in LSP. The editor only passes the file, line and column to the server. It's _not_ up to the editor to decide what is constitutes the region being completed. Then the server simply returns a list of candidates.

> The fuzzy completions, on the other hand, do not [work with Corfu]...

I take that back, actually! When I hit `C-M-i` after `\cite{partofthetitle` I do see the relevant suggestions. If I keep typing, then Emacs tries to refine the list and that immediately results in 0 candidates, since no BibTeX label starts with `partofthetitle`. But this is already much better than nothing.

minad:
> I take that back, actually! When I hit C-M-i after \cite{partofthetitle I do see the relevant suggestions. If I keep typing, then Emacs tries to refine the list and that immediately results in 0 candidates, since no BibTeX label starts with partofthetitle. But this is already much better than nothing.

I still don't understand fully what you describe now. When you enter `\cite{partofthetitle` the server returns the relevant completions, which do not match the prefix. But then you shouldn't see the candidates since they already don't match?

astoff:
Yes, there's indeed something funny going on. From just observing the UI and without further debugging, it seems that I see all candidates returned by the backend only as long as the completion style doesn't match any of the candidates. So:

* If the completion prefix is `foo` and the candidates are `afoo`, `fooa` and `bar` (the latter being a fuzzy match, e.g. it's a bibitem whose authors contain `bar`), then the UI shows only `afoo` and `fooa`. It would probably remove `fooa` as well if I wasn't using orderless.
* If the completion prefix is `foo` and the candidates are `x`, `y` and `z`, then:
    * The `*Completions*` buffer UI claims there are no matches
    * Corfu shows all the 3 candidates.

minad:
Okay, maybe there is something wrong with which string is used for the filtering. Does this also happen with Company? Both Corfu and Company use `completion-all-completions` to retrieve the completions. 

> Corfu shows all the 3 candidates.

In particular this is should not happen.

> If the completion prefix is foo and the candidates are afoo, fooa and bar (the latter being a fuzzy match, e.g. it's a bibitem whose authors contain bar), then the UI shows only afoo and fooa. It would probably remove fooa as well if I wasn't using orderless.

So this is a big problem with completion styles. I wonder how Company handles this problem, it should be affected as much as Corfu by the filtering. We probably need something to distinguish backend-filter-string and completion-sty-filter-string. You are using my Consult async commands, `consult-ripgrep` etc? I mean something like this, `#grep-regexp#completion-sty-filter`. The backend handles only the first part, the frontend handles the second part. In the context of LSP you wrote that there is not even the notion of a filter string passed to the backend, so one would actually only have `trigger-prefix#filter-string`, and `trigger-prefix#` is then ignored by the completion styles. The user has to enter this like `trigger-prefix TAB #filter-string`.

astoff:
Since [isearch-mb](https://github.com/astoff/isearch-mb) was mentioned here at some point, let me say that I made some changes and I think it works quite well now.

I should work out of the box with any isearch extension that starts a search (e.g., `consult-isearch`), and it just needs an advice to integrate with commands that run during the search or end a search. For instance:
```
(advice-add 'consult-line :around 'isearch-mb--after-exit)
(define-key isearch-mb-minibuffer-map (kbd "M-s l) 'consult-line)
```

There's some flicker in the match counts, but this is solved in Emacs 28.

minad:
@astoff Do you plan to push some of the changes upstream. I saw a mail on emacs devel.

astoff:
Good question.  I think this is good enough for a (M)ELPA package now (I'll test it some more, anyone else is welcome to try as well).  I'm skeptical the functionality could make it into isearch.el (if that's what you meant by upstream), because the Emacs people are extremely conservative about Isearch (when I tried to fix Isearch's manifestation of [this issue](https://github.com/astoff/isearch-mb/issues/1), they wanted to introduce a defcustom to enable the correct behavior!).  But I guess I'll ask.

minad:
@astoff It does not seem that conservative to me. From what I see there have been adjustments recently. Attaching the sorting properties to the strings for example, which was introduced with 27. But introducing an entirely different interaction mode is certainly more problematic. I don't know what the best way forward is. If its up to me I would probably make a radical change and throw out the existing isearch echo area mode. But I am not an Emacs maintainer and priest responsible for the preservation of all the Emacs idiosyncrasies.

astoff:
In my comment, I referred exclusively to Isearch key sequences. Apparently they have to be strictly backwards compatible. Not even making `isearch-del-char` backtrack a bit, as you suggested in that issue, is acceptable :shrug:

minad:
Since dabbrev has been discussed here and @protesilaos talked a few times about dabbrev, I just made sure that Corfu works with dabbrev. See https://github.com/minad/corfu/commit/fe8674a9ef4fb557ec2a5a8b4f1de2acdc960c31. Turns out, dabbrev does not provide a capf backend, but it still makes use of `completion-in-region` and the `completion-in-region-function` and therefore automatically works with Corfu :tada:

protesilaos:
> Since dabbrev has been discussed here and @protesilaos talked a few
> times about dabbrev, I just made sure that Corfu works with dabbrev. See
> ***@***.*** Turns out, dabbrev does not provide a capf backend,
> but it still makes use of completion-in-region and the
> completion-in-region-function and therefore automatically works with
> Corfu 🎉

Just tried it.  It works!

I guess that to call it with just TAB instead of dabbrev-completion we
would need it to have a capf backend?  Basically my ideal functionality
would merge as many candidates as possible from all relevant backends.

By the way, as I am writing this I realised that Corfu can work with
'message-mode-hook' when either BBDB or EBDB is in use: it helps
complete contacts in the To/Cc/Bcc fields.

-- 
Protesilaos Stavrou
https://protesilaos.com

minad:
> I guess that to call it with just TAB instead of dabbrev-completion we
would need it to have a capf backend?  Basically my ideal functionality
would merge as many candidates as possible from all relevant backends.

Yes, indeed.

> By the way, as I am writing this I realised that Corfu can work with
'message-mode-hook' when either BBDB or EBDB is in use: it helps
complete contacts in the To/Cc/Bcc fields.

This is probably the command `mail-abbrev-complete-alias`, which uses `completion-in-region`. Then there is another etags command. You can grep the code bases for `(completion-in-region ` to see all uses. The most important use case is certainly in the `completion-at-point` command, which invokes the capfs.

minad:
But I must say, actually I find it quite handy to have dabbrev always available on a separate key with the same UI. It may still make sense to implement a dabbrev capf. Since dabbrev.el already contains a completion table used by `dabbrev-completion`, with a slight refactoring one could turn the functionality into a capf.

oantolin:
Weird, I can't get `dabbrev-completion` to use Corfu. If I turn `corfu-mode` off, `dabbrev-completion` does use `consult-completion-in-region`, but with `corfu-mode` on I get the "Scanning for dabbrevs...done" message in the echo area but nothing happens, the Corfu UI does not appear.

minad:
@oantolin Maybe you use an outdated version since I fixed this just recently?

oantolin:
Probably, I'm using the one from GNU ELPA, not the main branch.

oantolin:
Yep, that was it. Current main works just fine, sorry for the noise.

oantolin:
@protesilaos Did you mean `mail-abbrev-complete-alias` as @minad thought? You mentioned BBDB, so I thought you might mean `bbdb-complete-mail`. If that's what you meant, I'd love for that to use Corfu, it always seems to popup a `*Completions*` buffer for me.

minad:
`bbdb-complete-mail` does not use `completion-in-region` it seems, it seems to do its own thing.

oantolin:
Yep, it saddens me. And `mail-abbrev-complete-alias` isn't very useful for me because I mostly don't use mail aliases. I think I only have one for each class I teach that expands to the list of all students in that class. For individual people I don't tend to use them.

protesilaos:
> @protesilaos Did you mean mail-abbrev-complete-alias as @minad thought?
> You mentioned BBDB, so I thought you might mean bbdb-complete-mail. If
> that's what you meant, I'd love for that to use Corfu, it always seems
> to popup a *Completions* buffer for me.

I am not sure about mail-abbrev-complete-alias because I did not check
it yet (I am currently writing an essay).  My guess is that it probably
is bbdb-complete-mail.  I use EBDB, which should be practically the
same.  Here is the relevant part:

    ebdb-complete-mail is a variable defined in ‘ebdb.el’.

    Its value is ‘capf’
    Original value was t

      You can customize this variable.

    If non-nil composition MUAs will complete EBDB contacts.
    Completion takes place within mail headers that specify one or
    more message recipients.  A value of ‘capf’ will add an EBDB
    collection to ‘completion-at-point-functions’.  Any other non-nil
    value will override "TAB" to call ‘ebdb-complete-mail’.

And when I check completion-at-point-functions I get the local value of:

(ebdb-mail-dwim-completion-at-point-function message-completion-function t)

The global value is (tags-completion-at-point-function) which must not
be what we are looking for.

-- 
Protesilaos Stavrou
https://protesilaos.com

oantolin:
capf! Great! I should switch to EBDB, I think this is one of its improvements over BBDB.

protesilaos:
> capf! Great! I should switch to EBDB, I think this is one of its
> improvements over BBDB.

Good!  I thought they were functionality equivalent...

-- 
Protesilaos Stavrou
https://protesilaos.com

oantolin:
I've migrated to EBDB and the capf email address completion is very nice! Thanks @protesilaos. By the way, I found a typo in your dotemacs: there is one occurrence of "EDBD" (I found it by making the same typo: I searched for "EDBD" and was surprised to see only one hit!).

protesilaos:
> I've migrated to EBDB and the capf email address completion is very
> nice! Thanks @protesilaos. By the way, I found a typo in your dotemacs:
> there is one occurrence of "EDBD" (I found it by making the same typo: I
> searched for "EDBD" and was surprised to see only one hit!).

Okay, corrected it!

By the way, if you ever use Github's ability to comment on a thread via
an email reply, you will want to append ***@***.***" to
"ebdb-permanent-ignores-file".

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #177, opened 2021-01-23
Save window excursion when previewing

Stebalien:
When consult is used with selectrum and mini-frame (minibuffer in a child-frame), window changes due to previews are not reverted on abort (`C-g`).

Saving the window excursion while selecting a candidate appears to fix the issue and doesn't appear to introduce any other issues. However, I'm not 100% sure _why_ this is an issue with mini-frame, and not an issue without it, so I have no idea if this is the correct fix.


minad:
This is not the correct fix since it is too intrusive. Can you please describe the issue precisely. Which kind of window changes are introduced during preview?

minad:
Please create a separate issue.

minad:
In case you are interested in investigating this, take a look at the consult--preview-position function which saves and restores the marker. Window configuration is not saved there as of now.

Stebalien:
https://github.com/minad/consult/issues/178

PR #182, opened 2021-01-25
consult--async-command: Flush candidates when input is too short

gexplorer:
When entering a not long enough input after getting some candidates, the candidate list was not being flushed so it was displaying a result from a previous input.
For example:
- Execute `consult-grep`
- Enter `#de`
  - No candidates because the input is too short
- Enter `#def`
  - Get candidates
- Remove the whole input
  - The candidates don't update and still show **def** to be matching
- Enter an input with no matches, like `#XX`
  - The candidates don't update and still show **def** to be matching

This pr prevents this bug, flushing the candidate list whenever the input is not long enough.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -1140,8 +1140,10 @@ The DEBOUNCE delay defaults to `consult-async-input-debounce'."
          (when debounce-timer
            (cancel-timer debounce-timer))
          (unless (string= action input)
-           (funcall async (setq input "")) ;; cancel running process
-           (unless (string= action "")
+           (setq input "")
+           (if (string= action "")
+               (funcall async 'flush)
+             (funcall async input) ;; cancel running process
              (setq debounce-timer (run-at-time
                                    (+ debounce (if unlocked 0 throttle)) nil
                                    (lambda () (funcall async (setq unlocked nil input action))))))))


minad:
Yes, I am aware of this. Actually this is intensional since I only wanted to flush when new candidates come in. For too short strings neither update nor flush is performed. Do you think it is too confusing?

Note that you can force searching for short strings by adding the second punctuation character, eg #a#

minad:
I just tried it again and don't find the current behavior good. In particular the example you are describing is not that convincing.

1. Type "#def" -> Found matches
2. Delete and type "#de" -> No update. Indicates to the user that the string is too short. There would be some matches.
3. Delete and type "#qq" -> No update. Indicates to the user that the string is too short. There would not be any matches.
4. Delete and type "#de#" -> Found matches (forced update)
5. Delete and type "#qq#" -> No matches (forced update)

In particular if we flush in steps 2 and 3, it would somehow give a wrong impression to the user. The current way things are done is that flushes never happen when the search is cancelled in order to avoid flicker, only when new candidates come in.

I made another experiment where I highlight the input string if it is too short - see https://github.com/minad/consult/tree/async-short. This indicates to the user that no update is happening. Maybe you want to give the branch a try? But I think it is a bit annoying and kind overkill to do this.

gexplorer:
Ok I see the experiment you made, but I still see a problem. Usually when filtering candidates the list reflects the input.

- I search `de` and I get no candidates, because the input is too short:
short

- I continue typing until I enter `def` which gives me a list of candidates with **def** highlighted
def

- Now if I delete the last char I get again `de` as input but now the candidate list is not empty as before but it shows the previous candidates with the previous **def** highlighted
de again

- If I start typing `XX` I still get the previous candidates and the minibuffer still tells me that is matching **def**
XX

I really think that the last 2 screenshots are wrong because I'm getting a list of candidates that does not match the input.

I understand that the implementation might not be the best, but I can't see the flicker you talk about. When I was trying to get it working I did notice that triggering a **flush** sometimes made things a bit crazy but as far as I can tell the way I did it it only happens when the input is forced to empty string by the input length check in `consult--async-split`.

minad:
I pushed a fix. Now the candidates list and the process indicator overlay should always be reset when the input string is empty. Please give it a test.

minad:
@gexplorer Unfortunately the flushing introduced some bug. When submitting the selected candidate, the input and the candidate list is flushed. Therefore this would require a more complicated solution. Since I don't feel strongly about this issue I decided to not introduce any flushing for the short input strings.

PR #183, opened 2021-01-25
consult-isearch: remove unnecessary line

astoff:
`with-suspended-isearch` actually starts a new Isearch if applicable.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -2651,7 +2651,6 @@ starts a new Isearch session otherwise."
                   (?r . "Regexp")
                   (?s . "Symbol")
                   (?w . "Word"))))
-    (unless isearch-mode (isearch-mode t))
     (with-isearch-suspended
      (setq isearch-new-string
            (consult--read


minad:
Thanks!

astoff:
Oh no, I just noticed this gives an error if you call `consult-isearch` before ever starting Isearch... After running Isearch at least one, all seem fine.

minad:
Okay.

astoff:
Apropos of the initial input: we can use `:initial isearch-string`, and revert this commit, which will have the side effect of clearing the global value of `isearch-string` when `consult-isearch` is called outside of Isearch.

minad:
> Apropos of the initial input: we can use :initial isearch-string, and revert this commit, which will have the side effect of clearing the global value of isearch-string when consult-isearch is called outside of Isearch.

Does not work. `isearch-string` seems to be "" in all cases.

astoff:
Ah, I see what's going on: it depends on whether you bind `isearch-consult` to `isearch-mode-map` or somewhere else. That's because Isearch uses, essentially, a transient map, so it suspends itself when a command from a different keymap is called.

PR #200, opened 2021-02-03
Add custom org source example

juliocc:
Small update to the README with another consult-buffer-sources example

### DIFF
--- a/README.org
+++ b/README.org
@@ -545,6 +545,17 @@ variables and functions with their descriptions.
   (add-to-list 'consult-buffer-sources 'term-source 'append)
   #+end_src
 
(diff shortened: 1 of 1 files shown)


minad:
Thanks! I think it is better to collect such configuration examples in the wiki, https://github.com/minad/consult/wiki. This is easier since it does not have to go through pull requests. Where it makes sense I will move examples from the wiki to the README. I just added links to the wiki to README in https://github.com/minad/consult/commit/f2960d6e958de83f16daf171b6c6397084721061.

One small correction regarding your source - the category should be `'buffer`. Besides that I recommend to use defvar for the source and add only the symbol to the list.

PR #213, opened 2021-02-08
consult-register-format: Do not truncate strings

minad:
@oantolin I just saw your commit, https://github.com/oantolin/emacs-config/commit/30443cc11078a2d5e8d3753efe95629461413b92. Do you think it makes sense to make this truncation behavior the default? I think I only truncated the register string because the original function does it. In our case showing the full string makes more sense and is more consistent with the consult-register ui, which does not truncate.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,5 +1,6 @@
 * Main
 
+- consult-register-format: Do not truncate register strings.
 - Multi sources/consult-buffer: Ensure that original buffer is
   shown, when the currently selected source does not perform preview.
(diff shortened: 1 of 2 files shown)


oantolin:
I like the register string shown in full, but since I often use Emacs un maximized, my lines aren't very wide, so I also like having the lines truncated. I feel it's the best of both worlds: things that should take up only one line do, but if I need to see the end of that line I can, if I want to.

Now that you added truncate-lines, I can remove that advice. Thanks!

PR #189, opened 2021-01-27
experiment: async find

minad:

Issue #89, opened 2020-12-24
Require style


minad:
I wonder about the best style to require libraries. We already require quite a few (heavy?) libraries which may not be loaded by default, should they be loaded via autoload-declaration, or required inside the function instead? The main problem is that the elisp checker does not like/check such declarations in contrast to proper require. Alternatively the consult.el file could be split up in multiple files? Maybe it is not necessary to do anything and just require everything in the main consult.el file. Would still be good to know the current best practices.

Ping @purcell for input

purcell:
Just `require` everything at the top level. That's equivalent to loading with `eval-and-compile`. If you know that you don't need certain libraries to be defined at either runtime or compile-time, you can potentially do things like `(eval-when-compile (require ...))` but this is very rarely the case. Keep it simple. Also, avoid requiring things from within functions.

minad:
Ok, so you are not recommending doing any kind of lazy loading of dependencies?

minad:
Thank you for answering here, today! Merry Christmas!🎄

This the style I am currently using. So you are not recommending doing any kind of lazy loading of dependencies?

purcell:
> So you are not recommending doing any kind of lazy loading of dependencies?

Correct. Usually trying to lazy load things manually is not worth the extra complication, and there is always the option of regular modularisation with generated autoloads.

purcell:
And... Merry Christmas!

minad:
> Correct. Usually trying to lazy load things manually is not worth the extra complication, and there is always the option of regular modularisation with generated autoloads.

Ok, thank you. Right now I am using the keep-it-simple-style, you are describing. I use this kind of autoload modularisation for consult-flymake.el. The other core dependencies are all required directly. Having a separate module only matters for dependencies which should not be loaded by default.

purcell:
`consult-flymake` is a perfect example, yes.

Issue #46, opened 2020-12-09
consult-rg/ag/grep?


liebkne:
Thanks for the great package. Is there any plan to implement search functions like `consult-rg/ag/grep`?

minad:
This is already on the wishlist. There is a roadblock though which prevents me from implementing this right now (see #10). But I am sure this will be resolved at some point. Furthermore I wonder why the other grep/rg packages do not fit the bill. I've read about live preview etc. But if other excellent packages already provide the functionality perfectly we may not implement it here.

PR #216, opened 2021-02-11
Add consult-xref.

JimDBh:
Add functions to let consult work with xref with preview. Half of it is basically copied from the `xref-show-definitions-completing-read` function in [xref.el ](https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/xref.el#L1027) and added a preview function for consult. 

The current xref branch looks a bit outdated and hopefully this could help with it.

### DIFF
--- a/consult-xref.el
+++ b/consult-xref.el
@@ -1,0 +1,103 @@
+;;; consult-xref.el --- Provides the command `consult-xref' -*- lexical-binding: t; -*-
+
+;; This file is not part of GNU Emacs.
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Provide support for showing xref with consult, with preview.
+;; Use the function (setup-consult-xref) to set it up.
+
+;;; Code:
+
+(require 'consult)
+(require 'xref)
+
+(defun xref-show-definitions-consult (fetcher alist)
+  "Let the user choose the target definition with completion.
+
+When there is more than one definition, let the user choose
+between them by typing in the minibuffer with completion."
+  (let* ((xrefs (funcall fetcher))
+         (xref-alist (xref--analyze xrefs))
+         xref-alist-with-line-info
+         xref
+         (group-prefix-length
+          ;; FIXME: Groups are not always file names, but they often
+          ;; are.  At least this shouldn't make the other kinds of
+          ;; groups look worse.
+          (let ((common-prefix (try-completion "" xref-alist)))
+            (if (> (length common-prefix) 0)
+                (length (file-name-directory common-prefix))
+              0))))
+
+    (cl-loop for ((group . xrefs) . more1) on xref-alist
+             do
+             (cl-loop for (xref . more2) on xrefs do
+                      (with-slots (summary location) xref
+                        (let* ((line (xref-location-line location))
+                               (line-fmt
+                                (if line
+                                    (format #("%d:" 0 2 (face xref-line-number))
+                                            line)
+                                  ""))
+                               (group-fmt
+                                (propertize
+                                 (substring group group-prefix-length)
+                                 'face 'xref-file-header))
+                               (candidate
+                                (format "%s:%s%s" group-fmt line-fmt summary)))
+                          (push (cons candidate xref) xref-alist-with-line-info)))))
+
+    (setq xref (if (not (cdr xrefs))
+                   (car xrefs)
+                 (let* ((collection (reverse xref-alist-with-line-info))
+                        (def (caar collection)))
+                   (cdr (assoc (consult--read collection
+                                              :prompt "xref results: "
+                                              :history nil
+                                              :require-match t
+                                              :predicate nil
+                                              :default def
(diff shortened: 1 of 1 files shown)


minad:
Thanks. This has been considered before in #189 and #46. I am not sure if is justified to have this function as part of Consult given that Emacs itself will offer this function soon. The preview function you proposed is not sufficiently robust since xref can open files which may not be desired. There is consult--jump-preview and consult--grep which implements a more robust preview. Consult xref preview should at least take a similar approach. I will add some more comments in the code.

consult-xref.el
```diff
@@ -0,0 +1,103 @@
...
+COLLECTION is the collection of definition candidates given by xref."
+  (let ((preview (consult--jump-preview))
+        (open (consult--temporary-files)))
+    (lambda (cand restore)
+      (cond
+       (restore
+        (funcall open)
+        (funcall preview nil t))
+       (t (let ((cand-xref (cdr (assoc cand collection))))
+                   (xref-pop-to-location cand-xref nil)))))))
```
minad, line 90: Instead of pop-to-location, preview and open should be used for a robust preview.

consult-xref.el
```diff
@@ -0,0 +1,103 @@
...
+
+(defun setup-consult-xref ()
+  "Set up consult with xref.
```
minad, line 93: This setup function should be removed, since the user can use setq directly.

consult-xref.el
```diff
@@ -0,0 +1,103 @@
...
+    (setq xref (if (not (cdr xrefs))
+                   (car xrefs)
+                 (let* ((collection (reverse xref-alist-with-line-info))
+                        (def (caar collection)))
+                   (cdr (assoc (consult--read collection
+                                              :prompt "xref results: "
+                                              :history nil
+                                              :require-match t
+                                              :predicate nil
+                                              :default def
```
minad, line 73: Predicate nil is unnecessary.

consult-xref.el
```diff
@@ -0,0 +1,103 @@
...
+                                (format "%s:%s%s" group-fmt line-fmt summary)))
+                          (push (cons candidate xref) xref-alist-with-line-info)))))
+
+    (setq xref (if (not (cdr xrefs))
+                   (car xrefs)
+                 (let* ((collection (reverse xref-alist-with-line-info))
+                        (def (caar collection)))
+                   (cdr (assoc (consult--read collection
+                                              :prompt "xref results: "
+                                              :history nil
```
minad, line 70: History nil is unnecessary or should be replaced with some history variable.

consult-xref.el
```diff
@@ -0,0 +1,103 @@
...
+;; Provide support for showing xref with consult, with preview.
+;; Use the function (setup-consult-xref) to set it up.
+
+;;; Code:
+
+(require 'consult)
+(require 'xref)
+
+(defun xref-show-definitions-consult (fetcher alist)
+  "Let the user choose the target definition with completion.
```
minad, line 29: The name should be consult-xref-... to follow the package convention.

consult-xref.el
```diff
@@ -0,0 +1,103 @@
...
+between them by typing in the minibuffer with completion."
+  (let* ((xrefs (funcall fetcher))
+         (xref-alist (xref--analyze xrefs))
+         xref-alist-with-line-info
+         xref
+         (group-prefix-length
+          ;; FIXME: Groups are not always file names, but they often
+          ;; are.  At least this shouldn't make the other kinds of
+          ;; groups look worse.
+          (let ((common-prefix (try-completion "" xref-alist)))
```
minad, line 41: Why do we want to determine the prefix using try-completion? Why is it necessary to mark the prefix with a special face?

oantolin: I think you misread the substring call below, the prefix isn't highlighted, but chopped off. The rest of of the file name is given a special face.

oantolin:
If I understand correctly, the only difference between this function  and the one in xref is the previews, but many of the search results will be in unopened files and consult usually doesn't have previews that require reading a file from disk. So, unless that policy changes, the previews here would have to be only for file that randomly happen to be open already, I guess.

Maybe file-opening previews could be admitted in Consult but off by default?

Is just having previews enough of a distinction for a command in Consult? I don't think there are currently examples of commands where the preview is the only distinguishing feature, are there? But also, is there anything wrong with that? Maybe previews are already worthwhile.

minad:
@oantolin consult-grep and consult-recent-file have file preview but it is written with some safety precautions. This should be added to the preview here. But I am unsure if it is sufficient as a distinction to justify a separate command. At some point I considered implementing preview in the style of Marginalia via the completion category for existing commands.

consult-xref.el
minad: Right, I still think it would be better to not use try-completion to find the common prefix.

minad:
When I merge consult-xref, I think I should also restructure consult a little bit such that we do not stuff more and more into the main consult.el file with the respective dependencies. This PR already proposes to create a consult-xref file.

* consult-outline: do not require outline, I am only using outline-regexp which does not require loading outline.
* consult-imenu -> move to consult-imenu.el, do not require imenu in consult.el
* consult-error -> move to consult-compile.el, rename to consult-compile-error, do not require compile in consult.el
* consult-flymake is already kept separately.
* consult-register -> move to consult-register.el. Maybe there are other candidates which could be moved out in order to keep the main library a little bit smaller. But this would not be be useful in order to improve lazy loading since register.el is always loaded. The rule could be to only move out components which pull in additional libraries.
* recentf and bookmark probably remain dependencies since they are used by consult-buffer and I want to keep this as is. But I am sure there are people who do not use recentf and bookmark.

Such a restructuring would come without downsides for the user and ensure that only the necessary components are loaded.  See also #89.

minad:
The consult-xref feature request has been noted on the wishlist #6. I close this PR since it is not mergeable in this form.

minad:
I added a `consult-xref` function which can be used as `xref-show-definitions-function` and `xref-show-xrefs-function`, see 29f6de93a9c920d2b03183e55f8f3454a19b8fc2.

JimDBh:
Thanks a lot for the update. I'll check out the new function.

Issue #222, opened 2021-02-16
Calling consul-compile-error from any buffer


astoff:
Shouldn't this be possible? I don't think `*compilation*` buffers are explicitly associated to a buffer or a project, but one could collect the compilation errors from the most recent compilation buffer, or all compilation buffers. If I have already switched to a compilation buffer, I see less value in calling `consult-compile-error`.

Then there's the case of auctex, which seems to have reinvented the compilation buffer. I wonder if `consult-compile--error-candidates` would be easy to adapt to this case (not that you go out of your way for compatibility with reinvented things).

minad:
Yes, this command is not super useful and fully baked in its current form. However I am also unsure if it makes sense to iterate over all compilation buffers, since for example grep buffers are also compilation buffers. Maybe one could use some filter function which selects the compilation buffers. But then it would also be neat if one could somehow use the consult-project-root-function to check if the current compile buffer belongs to the current project, but I have no idea if this is possible. Any ideas?

minad:
Regarding auctex, if it is not using the same compilation--message, compilation--loc structs, I don't have much hope in reusing code. I would rather try to modify auctex such that it uses the standard infrastructure. If this is not possible there is no other way than writing an extra consult-auctex package, if it is needed after all.

astoff:
The first line of ancompilation buffer reads
> But then it would also be neat if one could somehow use the consult-project-root-function in the filter function to check if the buffer belongs to the current project, but I have no idea if this is possible. Any ideas?

Sure: check the  `default-directory` of the compilation buffer. Also, the major mode of a compilation buffer and a grep buffer are different.

As to auctex, I agree. I think it's a recent TeX feature (meaning, from after 1982) to print error messages in a format compatible with other compilers. Still, it's funny how much stuff auctex reinvents.

minad:
> Sure: check the default-directory of the compilation buffer. Also, the major mode of a compilation buffer and a grep buffer are different.

Right, makes sense. Maybe then using a simple filter function would be the way to go. Either use all compilation buffers or if there is a project root function, check the default-directory. Are you interested in creating a PR?

astoff:
All right, I've sketched this and I will test it a bit before making the PR. Couple of questions in the meanwhile:

1. What's your preferred way to check that a file belongs to a project? We can select a compilation buffer by comparing its directory with the return value of `consult-project-root-function`, or just check that the current buffer's directory is a subdirectory of the compilation buffer's directory.
[...]

PR #225, opened 2021-02-17
consult-compile-error: Allow calling from any buffer

astoff:
This sketches a solution for #222. It jumbles together the results of all compilation-like buffers (including grep) relevant to the current buffer.

The function `compilation-find-file` sometimes asks interactively for a file name, which is not desired here. Maybe the same functions used for `consult-grep` preview should be used? (I'm not familiar with them.)

### DIFF
--- a/consult-compile.el
+++ b/consult-compile.el
@@ -25,33 +25,46 @@
 
 (require 'consult)
 (require 'compile)
+(require 'seq)
 
-(defun consult-compile--error-candidates ()
-  "Return alist of errors and positions."
-  (let ((candidates)
-        (pos (point-min)))
-    (save-excursion
-      (while (setq pos (compilation-next-single-property-change pos 'compilation-message))
-        (when-let (msg (get-text-property pos 'compilation-message))
-          (goto-char pos)
-          (push (list
-                 (consult--font-lock (consult--buffer-substring pos (line-end-position)))
-                 (point-marker)
-                 msg)
-                candidates))))
-    (nreverse candidates)))
+(defun consult-compile--error-candidates (buffer)
+  "Return alist of errors and positions in BUFFER, a compilation buffer."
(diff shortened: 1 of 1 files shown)


minad:
Thank you! Merged with minor changes (https://github.com/minad/consult/commit/7f75fe6e342a606e80488806dd7440ae9baee3f3). Please give it another check!

minad:
> The function compilation-find-file sometimes asks interactively for a file name, which is not desired here. Maybe the same functions used for consult-grep preview should be used? (I'm not familiar with them.)

Yes, we could do better by using our own functions, which do not ask the user. There is the grep state function which uses the `consult--temporary-files` closures to open files temporarily for preview. However in order to open a file it requires the file path, line and column. The compilation errors do not provide this but instead search for the file in some search path. All this is baked into the logic of `compilation-find-file`. We would have to replicate that, with the difference that the user should not be asked and the file opening should not use `find-file-noselect` but `consult--temporary-files` in order to restore the old state afterwards. Note that `consult--temporary-files` also has other safe guards in place, which prevent previewing large files or preview them in raw mode if their size surpasses some threshold.

~~~ elisp
(defun consult--grep-state ()
  "Grep preview state function."
  (let ((open (consult--temporary-files))
        (jump (consult--jump-state)))
    (lambda (cand restore)
      (when restore
        (funcall open))
      (funcall
       jump
       (when-let (buf (and cand (funcall (if restore #'find-file open)
                                         (car cand))))
         (with-current-buffer buf
           (save-restriction
             (save-excursion
               (widen)
               (goto-char (point-min))
               ;; Location data might be invalid by now!
               (ignore-errors
                 (forward-line (- (cadr cand) 1))
                 (forward-char (caddr cand)))
               (point-marker)))))
       restore))))
~~~

PR #226, opened 2021-02-18
Match prefix argument handling of insert-register

oantolin:
consult-register-load was using the opposite convention than
insert-register. Without a prefix argument, consult-register-load
leaves point before the inserted text, and with a prefix argument,
after; insert-register does the opposite. This simple change aligns
the two.

The discrepancy happened because jump-register and insert-register
commands use different conventions for translating the prefix argument
to their second function argument: jump-register uses the arg,
insert-register uses (not arg).


minad:
Thanks!

PR #227, opened 2021-02-18
consult-line: Quit isearch if applicable

astoff:
Currently Isearch is still on after finishing `consult-line` when it is invoked from isearch. I've also noticed some key conflicts between isearch and the minibuffer in my setup.

The argument to `isearch-done` makes the current search string not be saved in the isearch history. Feel free to remove it if you think saving is better.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -2051,7 +2051,8 @@ The symbol at point and the last `isearch-string' is added to the future history
      :lookup #'consult--line-match
      :default (car candidates)
      ;; Add isearch-string as initial input if starting from isearch
-     :initial (or initial (and isearch-mode isearch-string))
+     :initial (or initial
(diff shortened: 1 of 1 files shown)


minad:
Makes sense to end the isearch. Why not save it?

astoff:
Isearch doesn't save the string when you quit by `C-g`, which I don't like, but I thought not saving would be consistent.

Anyway, I updated the PR to save the string.

minad:
I think it is better to save in this case. It is like switching from isearch to occur. Does isearch save in that case?

astoff:
Occur doesn't even quit Isearch, but other ways to quit (movement commands, for instance) do save. So yes, makes more sense to save :-).

minad:
Right, occur just exports and keeps the search alive. Well I see consult-line triggered from isearch like switching to some live occur mode.

PR #228, opened 2021-02-19
Abbreviate names of recent files in consult--source-file

oantolin:
This is just a cosmetic improvement, to have shorter paths displayed for recent file candidates in consult buffer. Note that consult-recent-file already abbreviates file names.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -3286,7 +3286,8 @@ The command supports previewing the currently selected theme."
     :items
     ,(lambda ()
        (let ((ht (consult--cached-buffer-file-hash)))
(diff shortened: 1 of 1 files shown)


minad:
Thanks! I don't recall why I didn't do this right away. Maybe it was an oversight when I did the multi source refactoring or even earlier.

oantolin:
I figured it was an oversight. And it probably would not have even bothered me if it weren't for my stubborn insistence on grid view.  Those long file names are not _that_ long ---one per line you can read 'em--- but I want to see 2 or 3 per line, and using `~/` really helps them fit.

PR #230, opened 2021-02-21
Don't preview completion in region if its in the minibuffer

oantolin:
It doesn't work to show previews for completion in the minibuffer, since we prompt for a completion in a recursive minibuffer obscuring the one where we would be showing the previews.


minad:
I would prefer and not instead of unless ;)

oantolin:
> I would prefer and not instead of unless ;)

I did remember our conversation about `and` vs `when`, but I remember it too late, after submitting the pull request. Of course, you should feel free to change the `unless` to better match the rest of the code base.

PR #232, opened 2021-02-23
consult-grep: Use consult-grep category instead of xref-location

minad:


oantolin:
Oh, I somehow thought you had already merged this, before I merged your PR to Embark. There's still a good chance they land in the same MELPA cycle, I guess.

PR #235, opened 2021-02-25
consult--jump-preview: Extend overlay

angrybacon:
Hello :wave: 

This is a short PR to let preview overlays extend over the whole *physical* line. This is useful for `consult-line` mostly. Let me know if the change warrants a short introduction in the README.


minad:
Thanks! I am actually happy to make this the default. No need for an extra configuration option.

PR #265, opened 2021-04-08
consult-outline: Only fontify candidate lines

noctuid:
Instead of fontifying the entire buffer, only fontify candidates. This can greatly improve the speed of the first invocation in large buffers. It also allows getting fontification even in buffers larger than consult-fontify-max-size.

Let me know if this is okay. It seems to work well for me in testing. It is instant for me on the first run in files where the previous run took 6 seconds and in files bigger than `consult-fontify-max-size`. I'm not sure there would be a case where there would be enough headings to slow things down as much as `consult-line` in a buffer close to `consult-fontify-max-size` (it seems unlikely), but if there is a limit, I think it should be a separate variable. Let me know what you think.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -886,10 +886,15 @@ Since the line number is part of the candidate it will be matched-on during comp
                str))
             candidates)))
 
-(defsubst consult--buffer-substring (beg end)
-  "Return buffer substring between BEG and END."
+(defsubst consult--buffer-substring (beg end &optional fontifyp)
+  "Return buffer substring between BEG and END.
+If FONTIFYP and `consult-fontify-preserve' are non-nil, first ensure that the
+region has been fontified."
   (if consult-fontify-preserve
-      (buffer-substring beg end)
+      (progn
+        (when fontifyp
+          (consult--fontify-region beg end))
+        (buffer-substring beg end))
     (buffer-substring-no-properties beg end)))
 
 (defun consult--region-with-cursor (begin end marker)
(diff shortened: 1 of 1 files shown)


minad:
Thanks, I think this is good as is. If it turns out to be too slow we could still add another `fontify-outline-max-size` variable afterwards.

noctuid:
Yes, I tested it in a 2.2 megabyte org file with ~1.3k headings, and it is fast for me.

oantolin:
I don't doubt this faster, but I wonder if fontifying random lines works well. Doesn't fontifying accurately require context? I guess this is betting on the lines that form part of the outline being sort of synchronization points for fontification that can be fontified independently from the surrounding lines. I would bet that it is the case for common outline regexps.

minad:
@oantolin Yes, I always wondered what the region fontification function actually does and how it is supposed to work. I guess it just gives wrong results for regions. I hope it will work at least if the full lines are valid prefixes, which is often the case I guess (function definitions, org headlines, ...).

noctuid:
I think outline candidates should have the necessary context. I can't think of any cases where they wouldn't. If it's an issue, I guess this approach could be made optional.

u-w-u-u:
don't listen to @noctuid hes actually joe biden irl and its immoral to take the advice of someone who is a cannibalistic pedophile

u-w-u-u:
o u already merged i cant believe u sold out america like this

PR #276, opened 2021-04-18
consult-agenda and consult-clock-in

astoff:
Possible org-related commands were already discussed in the wishlist thread. In my opinion, the two most useful actions would be jumping to an agenda item and clocking in. This pull request is a draft for those two commands.

Possible improvements:
* Flatten the tree structure. That is, insetead of showing candidates
   ```
   * Heading 1
   ** Subheading 1.1
   * Heading 2
   ```
   display this as
   ```
   Heading 1
   Heading 1 > Subheading 1.1
   Heading 2
  ```
* Add narrowing. One could go wild here, and I don't even know what all the possibilities are. Some ideas:
   * By TODO/DONE state
   * By priority
   * By outline level ≤ n
* Performance: Building the candidate list could be slow, if one has lots of stuff in their agenda files. In my case it only takes a fraction of a second, but a lag is still noticeable.
* Category: should we use `consult-location` here, or invent a new category to allow attaching extra information to candidates, for instance the information needed for narrowing? Also, do I get it correctly that using lists of strings instead of lists of conses is the preferred way to build a candidate list now?
* Fontification: I think I'd rather remove the org heading faces from the candidates. I like to use larger, variable pitch faces for `outline-n`.
* In `consult-clock-in`, it would be nice to somehow inform the user about the currently clocked task, if any. One possibility would be to show the task name in the prompt, as in `Clock in (current: Heading 1): `. Any other suggestions?

What do you think?

### DIFF
--- a/consult-org.el
+++ b/consult-org.el
@@ -100,16 +100,17 @@ buffer are offered."
    (consult--read
     (consult--with-increased-gc (consult-org--entries match scope))
     :prompt "Go to heading: "
-    :category 'consult-org
+    :category 'consult-org-heading
     :sort nil
-    :title (lambda (cand)
-             (let ((marker (car (get-text-property 0 'consult-location cand))))
-               (buffer-name (marker-buffer marker))))
+    :title (unless (member scope '(nil 'tree 'region 'region-start-level 'file))
+             ;; Don't add titles when only showing entries from current buffer
+             (lambda (cand)
+               (let ((marker (car (get-text-property 0 'consult-location cand))))
+                 (buffer-name (marker-buffer marker)))))
     :narrow (consult-org--narrow)
     :require-match t
     :lookup #'consult--lookup-location
     :history '(:input consult-org--history)
-    :add-history (when (featurep 'org-clock) org-clock-current-task)
     :state (consult--jump-preview))))
 
 ;;;###autoload
@@ -150,7 +151,7 @@ recent clocked item."
            (and (member m1 org-clock-history)
                 (not (member m1 (member m2 org-clock-history))))))))
      :prompt "Clock in: "
-     :category 'consult-org
+     :category 'consult-org-heading
      :sort nil
      :title (lambda (cand)
               (let ((m (car (get-text-property 0 'consult-location cand))))
--- a/consult-org.el
+++ b/consult-org.el
@@ -1,0 +1,164 @@
+;;; consult-org.el --- Consult commands for org-mode -*- lexical-binding: t -*-
+
+;; Author: Daniel Mendler and Consult contributors
+;; Maintainer: Daniel Mendler <mail@daniel-mendler.de>
+;; Created: 2020
+;; License: GPL-3.0-or-later
+;; Version: 0.6
+;; Package-Requires: ((consult "0.6") (org "9.1") (emacs "26.1"))
+;; Homepage: https://github.com/minad/consult
+
+;; This file is not part of GNU Emacs.
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Provides a completing-read interface for some Org mode
+;; actions. This is an extra package, to allow lazy loading of Org.
+
+;;; Code:
+
+(require 'consult)
+(require 'org-agenda)
+(require 'org-clock)
+
+(defvar consult-org--history nil)
+
+(defvar consult-org--narrow
+  (append
+   ;; TODO keywords
+   (seq-filter
+    (lambda (it) (<= ?a (car it) ?z))
+    (mapcar (lambda (k)
+                  (list (downcase (string-to-char k)) k
+                        (lambda (cand)
+                          (string-equal k (get-text-property 0 'consult-org-todo cand)))))
+            (mapcan 'cdr org-todo-keywords)))
+   ;; Priorities
+   (mapcar (lambda (c)
+             (list c
+                   (format "Priority %c" c)
+                   (lambda (cand)
+                     (eq c (get-text-property 0 'consult-org-priority cand)))))
+           (number-sequence (max ?A org-highest-priority)
+                            (min ?Z org-lowest-priority)))
+   ;; Outline levels
+   (mapcar (lambda (i)
+             (list (+ i ?0)
+                   (format "Level %i" i)
+                   (lambda (cand)
+                     (>= i (get-text-property 0 'consult-org-level cand)))))
+           (number-sequence 1 9))))
+
+(defun consult-org--narrow ()
+  (cons (lambda (cand)
+          (funcall (nth 2 (assoc consult--narrow consult-org--narrow))
+                   cand))
+        (mapcar (pcase-lambda (`(,c ,s _)) (cons c s))
+                consult-org--narrow)))
+
+(defun consult-org--entries (match scope &rest skip)
+  "Return a list of consult locations from Org entries.
+
+MATCH, SCOPE and SKIP are as in `org-map-entries'."
+  (setq org-outline-path-cache nil)
+  (apply
+   #'org-map-entries
+   (lambda ()
+     (pcase-let ((`(_ ,level ,todo ,prio) (org-heading-components)))
+       (consult--location-candidate
+        (org-format-outline-path (org-get-outline-path t t))
+        (point-marker)
+        (current-line)
+        'consult-org-level level
+        'consult-org-todo todo
+        'consult-org-priority prio)))
+   match scope skip))
+
+;;;###autoload
+(defun consult-org-goto (&optional match scope)
+  "Jump to an Org heading.
+
+MATCH and SCOPE are as in `org-map-entries' and determine which
+entries are offered.  By default, all entries of the current
+buffer are offered."
+  (interactive (unless (derived-mode-p 'org-mode)
+                 (user-error "Must be called from an Org buffer.")))
+  (consult--jump
+   (consult--read
+    (consult--with-increased-gc (consult-org--entries match scope))
+    :prompt "Go to heading: "
+    :category 'consult-org
+    :sort nil
+    :title (lambda (cand)
+             (let ((marker (car (get-text-property 0 'consult-location cand))))
+               (buffer-name (marker-buffer marker))))
+    :narrow (consult-org--narrow)
+    :require-match t
+    :lookup #'consult--lookup-location
+    :history '(:input consult-org--history)
+    :add-history (when (featurep 'org-clock) org-clock-current-task)
+    :state (consult--jump-preview))))
+
+;;;###autoload
+(defun consult-org-agenda (&optional match)
+  "Jump to an Org agenda heading.
+
+By default, all agenda entries are offered.  MATCH is as in
+`org-map-entries' and can used to refine this."
+  (interactive)
+  (consult-org-goto match 'agenda))
+
+;;;###autoload
+(defun consult-org-clock-in (&optional match scope)
+  "Clock into an Org heading.
+
+MATCH and SCOPE are as in `org-map-entries' and determine which
+entries are offered.  By default, offer entries of files with a
+recent clocked item."
+  (interactive)
+  (unless scope
+    (org-clock-load)
+    (setq scope (thread-last org-clock-history
+                  (mapcar 'marker-buffer)
+                  (mapcar 'buffer-file-name)
+                  seq-uniq
+                  (seq-remove 'null))))
+  (org-clock-clock-in
+   (list
+    (consult--read
+     (consult--with-increased-gc
+      (sort ;; Recent items first, everything else in their original order
+       (consult-org--entries match scope)
+       (lambda (c1 c2)
+         (let ((m1 (car (get-text-property 0 'consult-location c1)))
+               (m2 (car (get-text-property 0 'consult-location c2))))
+           (and (member m1 org-clock-history)
+                (not (member m1 (member m2 org-clock-history))))))))
+     :prompt "Clock in: "
+     :category 'consult-org
+     :sort nil
+     :title (lambda (cand)
+              (let ((m (car (get-text-property 0 'consult-location cand))))
+                (if (member m org-clock-history)
+                    "Recent"
+                  (buffer-name (marker-buffer m)))))
+     :narrow (consult-org--narrow)
+     :require-match t
+     :lookup #'consult--lookup-location
+     :history '(:input consult-org--history)))))
+
+(provide 'consult-org)
+;;; consult-org.el ends here
--- a/consult-org.el
+++ b/consult-org.el
@@ -65,17 +65,19 @@ MATCH, SCOPE and SKIP are as in `org-map-entries'."
     (apply
      #'org-map-entries
      (lambda ()
+       ;; TODO: Is there no better way to obtain the line number? I guess it is guaranteed that
+       ;; `org-map-entries' iterates over all entries in a file in order.
        (unless (eq buffer (current-buffer))
          (setq opoint (point-min)
                line 1
                buffer (current-buffer)
-               org-outline-path-cache nil))
+               org-outline-path-cache nil)) ;; Reset the cache since `org-get-outline-path' uses the cache
        (setq line (+ line (consult--count-lines (prog1 (point)
                                                   (goto-char opoint))))
              opoint (point))
        (pcase-let ((`(_ ,level ,todo ,prio) (org-heading-components)))
          (consult--location-candidate
-          (org-format-outline-path (org-get-outline-path t t))
+          (org-format-outline-path (org-get-outline-path 'with-self 'use-cache))
           (point-marker)
           line
           'consult-org--level level
@@ -158,6 +160,7 @@ recent clocked item."
     (org-clock-clock-in
      (list
       (consult--read
+       ;; TODO This code should also go to `consult-org--clock-in-candidates'
        (nconc
         (seq-filter #'stringp
                     (mapcar (lambda (m) (gethash m recent))
--- a/consult-org.el
+++ b/consult-org.el
@@ -24,7 +24,6 @@
 
 (require 'consult)
 (require 'org-agenda)
-(require 'org-clock)
 
 (defvar consult-org--history nil)
 
@@ -121,73 +120,6 @@ By default, all agenda entries are offered. MATCH is as in
   (unless org-agenda-files
     (user-error "No agenda files"))
   (consult-org-heading match 'agenda))
-
-;; Only suitable for short-lived hash tables, since a marker's
-;; position can change.
-(define-hash-table-test 'consult-org--marker-test
-  'equal
-  (lambda (m)
-    (sxhash-equal (cons (marker-position m) (marker-buffer m)))))
-
-;;;###autoload
-(defun consult-org-clock-in (&optional match scope)
-  "Clock into an Org heading.
-
-MATCH and SCOPE are as in `org-map-entries' and determine which
-entries are offered. By default, offer entries of files with a
-recent clocked item."
-  (interactive)
-  ;; TODO move all preprocessing into the increased gc, into a `consult-org--clock-in-candidates'
-  ;; function, which returns candidates and recent hash
-  (let* ((scope (or scope
-                    (thread-last (progn (org-clock-load) org-clock-history)
-                      (mapcar #'marker-buffer)
-                      (mapcar #'buffer-file-name)
-                      (delete-dups)
-                      (delq nil))
-                    ;; TODO use current file as scope as fallback?
-                    ;; I am still not entirely happy with the idea of using the list of files
-                    ;; containing recent clock entries as scope. Here we invent some new scope
-                    ;; which is not present in Org. It may be preferrable to use either the current Org
-                    ;; file or the agenda or maybe we could even define our own org scope variable?
-                    (user-error "No recent clocked tasks")))
-         (candidates (consult--with-increased-gc
-                      (consult-org--entries match scope)))
-         (recent (let ((tbl (make-hash-table :test 'consult-org--marker-test
-                                             :size org-clock-history-length)))
-                   (dolist (m org-clock-history)
-                     (puthash m t tbl))
-                   (dolist (c candidates)
-                     (let ((m (car (get-text-property 0 'consult-location c))))
-                       (when (gethash m tbl)
-                         (puthash m c tbl))))
-                   tbl)))
-    (org-clock-clock-in
-     (list
-      (consult--read
-       ;; TODO This code should also go to `consult-org--clock-in-candidates'
-       (nconc
-        (seq-filter #'stringp
-                    (mapcar (lambda (m) (gethash m recent))
-                            org-clock-history))
-        (seq-remove (lambda (c) (gethash
-                                 (car (get-text-property 0 'consult-location c))
-                                 recent))
-                    candidates))
-       :prompt "Clock in: "
-       :category 'consult-location
-       :sort nil
-       :group
-       (apply-partially #'consult--group-by-title
-                        (lambda (cand)
-                          (let ((m (car (get-text-property 0 'consult-location cand))))
-                            (if (gethash m recent)
-                                "Recent"
-                              (buffer-name (marker-buffer m))))))
-       :narrow (consult-org--narrow)
-       :require-match t
-       :lookup #'consult--lookup-location
-       :history '(:input consult-org--history))))))
 
 (provide 'consult-org)
 ;;; consult-org.el ends here
--- a/consult-org.el
+++ b/consult-org.el
@@ -23,35 +23,32 @@
 ;;; Code:
 
 (require 'consult)
-(require 'org-agenda)
+(require 'org)
 
 (defvar consult-org--history nil)
 
 (defun consult-org--narrow ()
   "Narrowing configuration for `consult-org' commands."
-  (let ((todo-keywords
-         ;; TODO: Is there no function in org which provides keywords+keys?
-         ;; Why do we have to do our own splitting here?
+  (let ((todo-kws
          (seq-filter
-          (lambda (it) (<= ?a (car it) ?z))
+          (lambda (x) (<= ?a (car x) ?z))
           (mapcar (lambda (s)
                     (pcase-let ((`(,a ,b) (split-string s "(")))
                       (cons (downcase (string-to-char (or b a))) a)))
-                  (mapcan #'cdr org-todo-keywords)))))
+                  (apply #'append (mapcar #'cdr org-todo-keywords))))))
     (cons (lambda (cand)
             (pcase-let ((`(_ ,level ,todo ,prio)
                          (get-text-property 0 'consult-org-heading cand)))
               (cond
                ((<= ?1 consult--narrow ?9) (<= level (- consult--narrow ?0)))
                ((<= ?A consult--narrow ?Z) (eq prio consult--narrow))
-               ;; TODO instead attach a 'consult-org--todo-key to the candidate
-               (t (equal todo (alist-get consult--narrow todo-keywords))))))
-          (append (mapcar (lambda (c) (cons c (format "Level %c" c)))
-                          (number-sequence ?1 ?9))
-                  (mapcar (lambda (c) (cons c (format "Priority %c" c)))
-                          (number-sequence (max ?A org-highest-priority)
-                                           (min ?Z org-lowest-priority)))
-                  todo-keywords))))
+               (t (equal todo (alist-get consult--narrow todo-kws))))))
+          (nconc (mapcar (lambda (c) (cons c (format "Level %c" c)))
+                         (number-sequence ?1 ?9))
+                 (mapcar (lambda (c) (cons c (format "Priority %c" c)))
+                         (number-sequence (max ?A org-highest-priority)
+                                          (min ?Z org-lowest-priority)))
+                 todo-kws))))
 
 (defun consult-org--headings (match scope &rest skip)
   "Return a list of Org heading candidates.
--- a/consult-org.el
+++ b/consult-org.el
@@ -38,38 +38,28 @@
 
 (defun consult-org--narrow ()
   "Narrowing configuration for `consult-org' commands."
-  (let ((triples (append
-                  ;; TODO keywords
-                  (seq-filter
-                   (lambda (it) (<= ?a (car it) ?z))
-                   (mapcar (lambda (s)
-                             (list (downcase (string-to-char s))
-                                   s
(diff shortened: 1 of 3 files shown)


minad:
I am still a bit on the line regarding org commands. While I use org and like it, it is offers vast functionality and I am not always sure that new functionality is really needed inside Consult. Org already provides many fancy UIs on its own. So maybe I just feel like I am not enough of an org power user to maintain these commands and they are a better fit for a separate consult-org package? 

I observed for example that the command `org-set-tags-command` is broken if you use an enhanced completion UI (https://github.com/minad/vertico#org-set-tags-command). I think I've also seen the tags command which fixes the upstream version proposed here. I think such command proposals should be rejected or rather fixed directly in Org. 
But the commands you are proposing here are of a different quality, more useful and not just fix-ups of some upstream versions. 

To address your comments:

> Flatten the tree structure

I cannot tell what is better. For imenu it works well, for `consult-outline` it would not work well. It all depends on if you want to show the full line or only the heading. Maybe give it a try to see which works out better?

> Add narrowing

I personally wouldn't overdo it with the narrowing options. Right now, narrowing is used in a very "narrow" sense only, think like in consult-buffer/consult--multi, to unhide a few candidates for example. There is no support for fancy filtering. I thought before about supporting more options for narrowing, allowing to narrow according to multiple criteria etc. But these things are in conflict with completion filtering, why would we want to add another filtering system on top of basic completion? For fancy filtering narrowing one would rather want https://github.com/oantolin/orderless/issues/30.

> Performance: Building the candidate list could be slow, if one has lots of stuff in their agenda files. In my case it only takes a fraction of a second, but a lag is still noticeable.

Is it possible to cache the items somehow? If it is too slow, it should not be added here. But a noticeable lag does not sound too bad, it is like consult-line on a large file.

> Category: should we use consult-location here, or invent a new category to allow attaching extra information to candidates

It depends on if you want to attach Marginalia annotations or Embark actions.

> for instance the information needed for narrowing? 

The category is not needed for narrowing.
(There is a thought - one could write a narrowing package which enhances existing commands based on the completion category. Then one would need separate completion categories. However since narrowing is implemented via predicates it would not be so easy and could break commands. The predicates also receive the exact candidates, e.g., conses and not the candidate strings).

> Also, do I get it correctly that using lists of strings instead of lists of conses is the preferred way to build a candidate list now?

Yes, that's right.

> Fontification: I think I'd rather remove the org heading faces from the candidates. I like to use larger, variable pitch faces for outline-n.

This should be configurable if you use `consult--buffer-substring`, see `consult-fontify-preserve`.

> In consult-clock-in, it would be nice to somehow inform the user about the currently clocked task, if any. One possibility would be to show the task name in the prompt, as in Clock in (current: Heading 1): . Any other suggestions?

Sounds good. But I am not using the clock feature right now :shrug: 

Sorry for not being more concrete! I will add some more comments to the code.

consult.el
```diff
@@ -3699,6 +3700,74 @@ same major mode as the current buffer are used. See also
   (interactive)
   (consult--imenu (consult--imenu-all-items (consult--imenu-project-buffers))))
 
+;;;;; Command: consult-agenda
+
+(defun consult--org-entries (match scope &rest skip)
+  "Return a list of consult locations from Org entries.
+
+MATCH, SCOPE and SKIP are as in `org-map-entries'."
+  (require 'org)
```
minad, line 3709: Should go to a separate file, require org directly.

consult.el
```diff
@@ -3699,6 +3700,74 @@ same major mode as the current buffer are used. See also
...
+  (apply
+   #'org-map-entries
+   (lambda ()
+     (consult--location-candidate
+      (consult--buffer-substring (line-beginning-position)
+                                 (line-end-position)
+                                 'fontify)
+      (point-marker)
+      (current-line)))
```
minad, line 3718: This exact lambda exists in `consult-outline` and should be extracted to a common function `consult--line-candidate`. But maybe you will go with the flattening instead?

consult.el
```diff
@@ -3699,6 +3700,74 @@ same major mode as the current buffer are used. See also
...
+    :title (lambda (cand)
+             (let ((marker (car (get-text-property 0 'consult-location cand))))
+               (buffer-name (marker-buffer marker))))
+    :require-match t
+    :lookup #'consult--lookup-location
+    :history '(:input consult--agenda-history)
+    :add-history (when (featurep 'org-clock) org-clock-current-task)
+    :state (let ((preview (consult--jump-preview)))
+             (lambda (cand restore)
+               (funcall preview cand restore))))))
```
minad, line 3740: Eta expansion?

astoff:
> Org already provides many fancy UIs on its own.

True. It tends to use Magit-style menus (of course Org predates Magit, by far). In fact, `C-u M-x org-clock-in` provides a menu for clocking in. If you ask me, it's the wrong kind of UI for that operation. That's why I find the proposed `consult-clock-in` relevant.

> So maybe I just feel like I am not enough of an org power user to maintain these commands and they are a better fit for a separate consult-org package?

I am by no means an Org power user, and I wouldn't want to maintain a consult-org package. Now, I would argue the 2 proposed commands are of somewhat broad appeal. I can't think of a third org-related command that would be of broad appeal.

> I personally wouldn't overdo it with the narrowing options.

In this case I will not add any narrowing, at least for now.

I'll react to your other comments when I have time to work on this again. And we can also wait a bit to let the question of having or not Org-related stuff in Consult mature.

consult.el
astoff: :see_no_evil:

astoff:
So, I think I've addressed all your code comments above. I also added narrowing; I may even have gone a bit overboard with it. But it turns out that the most useful (narrow by TODO) was the trickier to get right. Concerning flattening of the tree structure, I noticed the command would be useless without it. Lastly, I created separated `consult-org-goto` and `consult-org-agenda` commands. The former by default only lists headings in the current buffer. I think it will be clear why when you see the code.

And again: I'm not insisting that this should be merged :-). I don't like the idea of having lots of separate little packages, but I might create it if you don't want to include this file in Consult. Or maybe not...

consult-org.el
```diff
@@ -0,0 +1,164 @@
...
+MATCH, SCOPE and SKIP are as in `org-map-entries'."
+  (setq org-outline-path-cache nil)
+  (apply
+   #'org-map-entries
+   (lambda ()
+     (pcase-let ((`(_ ,level ,todo ,prio) (org-heading-components)))
+       (consult--location-candidate
+        (org-format-outline-path (org-get-outline-path t t))
+        (point-marker)
+        (current-line)
```
astoff, line 84: Hah, so this is not a standard function. Is this line number used only for display? It seems safe to put a nil here.

minad: No this metadata is needed by embark-consult, e.g., if you want embark-export or embark actions to work.

consult-org.el
```diff
@@ -0,0 +1,164 @@
...
+;;; Code:
+
+(require 'consult)
+(require 'org-agenda)
+(require 'org-clock)
+
+(defvar consult-org--history nil)
+
+(defvar consult-org--narrow
+  (append
```
minad, line 40: Shouldn't this be dynamically computed since the values may change?

consult-org.el
```diff
@@ -0,0 +1,164 @@
...
+   (lambda ()
+     (pcase-let ((`(_ ,level ,todo ,prio) (org-heading-components)))
+       (consult--location-candidate
+        (org-format-outline-path (org-get-outline-path t t))
+        (point-marker)
+        (current-line)
+        'consult-org-level level
+        'consult-org-todo todo
+        'consult-org-priority prio)))
+   match scope skip))
```
minad, line 88: I prefer `consult-org--` for the property prefix. I think I have this consistently over the codebase?

minad:
Okay, I think this looks good besides a few minor nits.

consult-org.el
```diff
@@ -0,0 +1,164 @@
...
+
+;;;###autoload
+(defun consult-org-goto (&optional match scope)
```
minad, line 91: One more nitpick - Consult has this naming convention `consult-<thing>`, so maybe this should be `consult-org-heading`?

consult-org.el
astoff: At least one property, `consult-location`, is not double dashed. If you decide to change this, may I suggest separating `consult--marker` and `consult--line` to avoid some cars and pcases?

minad: No, consult-location is intentionally not double dashed since it is the public metadata for embark-consult.

consult-org.el
astoff: The thinking here was that this defvar, if not customizable, is at least hackable. Of course this is questionable. I'll change it.

consult.el
astoff: That's right: this lambda is gone, I'm using Org's headline formatter instead.

consult-org.el
minad: Okay, I understand. This makes sense. I still think the dynamic variant is better. If it turns out that we actually want to customize this we can either add more stuff or add some defcustom.

consult-org.el
astoff: I've optimized the line counting. It makes a quite big difference.

consult-org.el
astoff: I've simplified this part a bit.

consult-org.el
astoff: Okay, also renamed `consult-org-clock` (was `...-clock-in` before).

astoff:
You might want to add to the user configuration that `consult-org-agenda` requires `org-agenda-files` to be set, and `consult-org-clock` benefits from saving recent clock entries across Emacs sessions, which is done by calling `(org-clock-persistence-insinuate)` in the init file.

consult-org.el
minad: Ah, yes. This is important. I have this optimized everywhere in order to avoid the quadratic behavior.

consult-org.el
minad: This one I am not so sure about since we are basically replacing `org-clock-in`, see also `consult-goto-line` which is a replacement for `goto-line`. The `consult-<thing>` convention is not a hard rule. Please check out the other commands - maybe some should be renamed?

minad:
> You might want to add to the user configuration that consult-org-agenda requires org-agenda-files to be set, and consult-org-clock benefits from saving recent clock entries across Emacs sessions, which is done by calling (org-clock-persistence-insinuate) in the init file.

You mean to the readme?

consult-org.el
```diff
@@ -0,0 +1,167 @@
...
+     (consult--with-increased-gc
+      (sort ;; Recent items first, everything else in their original order
+       (consult-org--entries match scope)
+       (lambda (c1 c2)
+         (let ((m1 (car (get-text-property 0 'consult-location c1)))
+               (m2 (car (get-text-property 0 'consult-location c2))))
+           (and (member m1 org-clock-history)
+                (not (member m1 (member m2 org-clock-history))))))))
+     :prompt "Clock in: "
+     :category 'consult-org-heading
```
minad, line 154: Why did you decide for an extra category now? It may be better to use `consult-location` because this is what I used now for general location strings. However as I mentioned befre, if you want to attach org specific actions, to headings a separate category is better (maybe even 'org-heading). But as of now there are no embark actions for org headings and I am unsure if they are worth it since org already has its own set of commands to manipulate the headings. cc @oantolin

minad: See also https://github.com/oantolin/embark/issues/95#issuecomment-758246901 by  @sheijk

astoff: I think you are right, this would be too fine grained. This would be fine (but also maybe precious) if we could declare that `consult-org-heading` inherits from `consult-location`. That not being the case, I'll revert to `consult-location`. And I agree that some of the editing features discussed in the linked comment are already covered satisfactorily in Org itself.

minad: > This would be fine (but also maybe precious) if we could declare that consult-org-heading inherits from consult-location. That not being the case, I'll revert to consult-location.

You can actually do that in Embark, a `consult-org-heading` keymap could inherit from the `consult-location` keymap. But I think using a separate category (Embark target type) for org headings would only make sense if Embark would support actions at point for org headings (then the category 'org-heading would make more sense, since it could also work without consult).

astoff:
> You mean to the readme? Feel free to add your documentation there too.

I was going to add this, but since it's more of an org-mode hint, I'll let you decide what to do

```
 ;; Optionally configure Org.
 (use-package org
   :config
   ;; Add some files to your agenda, either with `C-c [` or manually:
   (setq org-agenda-files ...)
   ;; `consult-org-clock-in` offers recent clocked tasks for
   ;; selection. To remember those across Emacs sessions, use the
   ;; following:
   (org-clock-persistence-insinuate))
```
Otherwise, it's perhaps good to merge now :-)

minad:
Okay, I will probably not add the config example to the README. Org users will know. This looks pretty good now. I took another closer look and still have some improvement proposals. Sorry for the back and forth multiple times, I guess I am a bit more critical since these are the first org commands which are going to be added.

consult-org.el
```diff
@@ -0,0 +1,167 @@
...
+    (cons (lambda (cand)
+            (cond ((<= ?1 consult--narrow ?9)
+                   (<= (get-text-property 0 'consult-org--level cand)
+                       (- consult--narrow ?0)))
+                  ((<= ?A consult--narrow ?Z)
+                   (eq (get-text-property 0 'consult-org--priority cand)
+                       consult--narrow))
+                  ((when-let ((todo (alist-get consult--narrow todo-keywords)))
+                     (string-equal (get-text-property 0 'consult-org--todo cand)
+                                   todo)))))
```
minad, line 56: I would like to see the narrowing function to be simplified. You can either encode the possible keys into an integer as in `consult-minor-mode-menu`: 

https://github.com/minad/consult/blob/f22309220abbc77074eddd16530d98e268695105/consult.el#L3299-L3302

Alternatively store the possible narrow keys as a string in each candidate and then check if `consult--narrow` is an element of this string.

consult-org.el
```diff
@@ -0,0 +1,167 @@
...
+                                                  (goto-char opoint)))))
+       (setq opoint (point))
+       (pcase-let ((`(_ ,level ,todo ,prio) (org-heading-components)))
+         (consult--location-candidate
+          (org-format-outline-path (org-get-outline-path t t))
+          (point-marker)
+          line
+          'consult-org--level level
+          'consult-org--todo todo
+          'consult-org--priority prio)))
```
minad, line 87: Here only attach this single property which encodes the narrow keys.

consult-org.el
```diff
@@ -0,0 +1,167 @@
...
+buffer are offered."
+  (interactive (unless (derived-mode-p 'org-mode)
+                 (user-error "Must be called from an Org buffer.")))
+  (consult--jump
+   (consult--read
+    (consult--with-increased-gc (consult-org--entries match scope))
+    :prompt "Go to heading: "
+    :category 'consult-location
+    :sort nil
+    :title (unless (member scope '(nil 'tree 'region 'region-start-level 'file))
```
minad, line 105: This quoting seems incorrect?

consult-org.el
```diff
@@ -0,0 +1,167 @@
...
+         (let ((m1 (car (get-text-property 0 'consult-location c1)))
+               (m2 (car (get-text-property 0 'consult-location c2))))
+           (and (member m1 org-clock-history)
+                (not (member m1 (member m2 org-clock-history))))))))
+     :prompt "Clock in: "
+     :category 'consult-location
+     :sort nil
+     :title (lambda (cand)
+              (let ((m (car (get-text-property 0 'consult-location cand))))
+                (if (member m org-clock-history)
```
minad, line 158: Using member here is not good, since it will lead to O(n*m). Ideally attach the necessary information to the candidate directly?

consult-org.el
```diff
@@ -0,0 +1,167 @@
...
+  (org-clock-clock-in
+   (list
+    (consult--read
+     (consult--with-increased-gc
+      (sort ;; Recent items first, everything else in their original order
+       (consult-org--entries match scope)
+       (lambda (c1 c2)
+         (let ((m1 (car (get-text-property 0 'consult-location c1)))
+               (m2 (car (get-text-property 0 'consult-location c2))))
+           (and (member m1 org-clock-history)
```
minad, line 151: Same here, no member in the sorting because of `O(n*log(n)*m)`.

minad: But maybe the clock history is always very short or you could cut it off? As I said, I am not using the clocking feature yet. Generally I like to keep long histories.

consult-org.el
```diff
@@ -0,0 +1,167 @@
...
+  (consult-org-heading match 'agenda))
+
+;;;###autoload
+(defun consult-org-clock-in (&optional match scope)
+  "Clock into an Org heading.
+
+MATCH and SCOPE are as in `org-map-entries' and determine which
+entries are offered. By default, offer entries of files with a
+recent clocked item."
+  (interactive)
```
minad, line 134: Here I wonder about the default behavior. Wouldn't it make sense to use the agenda? Or write two commands `consult-org-clock-in-agenda` and `consult-org-clock-in-recent`?

consult-org.el
astoff: This won't help, since `consult-org--entries` would then take O(m*n). And one would have to pay this cost even if the UI doesn't use `:title`.

I also think creating a hash table to trade O(m) by O(log(m)) won't help in practice, since m (= `org-clock-history-length`) is typically small, 20 by default.

consult-org.el
astoff: Okay, let's think about this. Sorting usually takes O(n log(n)) operations. But now each operation takes O(m), so that makes it O(mn log(n)).

An alterative would be to split off the recent items (takes O(mn)) and then sort only the recent items (takes O(m^2 log(m^2)), and concatenate (takes O(n)). Overall this is O(mn), assuming n >> m.

Replacing the list by a hash table would cut it down to O(n log(m)), but is probably doesn't break even in practice.

Should we go for the O(mn) option?

minad: The proper solution is to replace the list by the hash table to ensure O(nlogn), but if the list is guaranteed to be short ("constant") then we already have that. 

I am a bit wary about the quadratic behavior, since in my Vertico package sorting by history position turned out to be a big bottleneck, which had to be fixed using a hashtable. But there one does not control the history since the UI should work for arbitrary commands. Here we know the history.

You are right about the break even, the hashtable is slow and one should decorate the items before sorting to cut down factors a bit. But breaking even is not what matters, since for small numbers we are already fast in any case. Even when the optimization hurts, we are not hurt much. I am only asking about retaining good behavior for long histories.

Btw, there is already the helper `consult--string-hash` if one would decide to turn the list into a hash.

consult-org.el
minad: > I also think creating a hash table to trade O(m) by O(log(m)) won't help in practice, since m (= org-clock-history-length) is typically small, 20 by default.

I don't understand. If you use a hashtable for the clock history we get O(n+m).

consult-org.el
minad: Hmm, assuming you have these 20 element histories, member is good enough and we can always use `consult--string-hash` later on if it turns out to be a problem. 

But then - it seems this clocking history is not really a history but you intend to use it here as a list of recent items which are still relevant for clocking, the elements which got pushed over the edge are not relevant anymore. I must say I am not really fond of this design. I think I would prefer to have a more deterministically behaving command `consult-clock-in-agenda` which uses the agenda, which is controlled by the user. And additionally one can still sort using the history. The history can be long in that case, requiring the hashtable.

The problem here is that I am not using the clocking currently, otherwise I could judge a bit better.

consult-org.el
astoff: One might often be interested in clocking into agenda entries, but clocking and agenda are two independent features. The first draft of this PR used the agenda, but then I realized there is this alternative way to get a list of clock-relevant files. Do you think it could be confusing / overly smart?

To clock into an agenda entry, one can call `(consult-org-clock-in nil 'agenda)`. Maybe it suffices to let the interested user define this in their config?

consult-org.el
astoff: Most of the time, the user will not use narrowing. Why would we spend time here to precompute a value that makes it easier to narrow? It also seems nicer to keep this metadata transparent, in case the user wants to redefine their narrow criteria or if https://github.com/oantolin/orderless/issues/30 comes into being.

Or did you say single property to mean something like `'consult-org--narrow` (list level todo prio)`?

Note also that the collection of TODO keywords is customizable, and would have to been computed here again if we wanted to encode it.

consult-org.el
astoff: Is the problem that this lambda isn't sufficiently elegant? I agree it's ugly, but I think your suggestion would spread the inelegance to other places (e.g., `todo-keywords` would have to be available in `consult-org--entries` as well).

consult-org.el
astoff: Fixed.

consult-org.el
minad: Yes, my problem here is that the narrow function is overly complicated. I prefer to push the complexity into the candidate generation. I think overall it will be a bit simpler.

consult-org.el
minad: > It also seems nicer to keep this metadata transparent, in case the user wants to redefine their narrow criteria or if oantolin/orderless#30 comes into being.

Yes, then we can reconsider to add special properties with public (non double dashed) names. I mean one can always find arguments either way. I think I prefer it the other way since it is more consistent with the rest of the code.

> Or did you say single property to mean something like 'consult-org--narrow (list level todo prio)?

I meant to attach a property to the candidates which is a string of the narrowing keys, e.g. "0tA". Then the narrow function only has to check using seq-position if consult--narrow is an element of the keys.

consult-org.el
minad: Where does the original org clocking command gets the clockable files/items? And yes I find it a bit "overly dwim" to scan the history for the clocking relevant files.

astoff:
Now that I think more about the narrow API, I get the impression that it was thought around splitting the candidate list into disjoint collections, and one needs to do some contortions when that's not the case.

Fundamentally, the narrowing data should just be a list of predicates. But since we need a UI to select a predicate, the narrowing data should be a list of elements of the form `'(character title predicate)`, where `predicate` is just a function that takes a candidate as argument.

In many cases this would seem like overkill, and for convenience one could accept
```
'(function (?a "Category A") (?b "Category B") ...)
```
as a synonym for
```
`((?a "Category A" ,(apply-partially function ?a)) (?b "Category B") ,(apply-partially function ?b)) ...)
```
or something similar using a dynamic var.

This would also make the narrowing data reusable. For instance, we could append the `consult-outline` narrowing data to that of the org commands, and only implement narrowing by priority and todo state in consult-org.el.

Or am I missing something?

consult-org.el
astoff: The basic way to clock in is to move to a headline and call `M-x org-clock-in`. The alternative is to type `C-u M-x org-clock-in`; then a window with 20 recent tasks appears and you can press a character to select one of them.

Originally I just wanted a command to clock into agenda entries. But now I realize that, if I were to clock into a task in a random file, then it would be quite likely that I would clock into another entry of that same file later on. On the other hand, I have 2 agenda files, and I only clock into items of one (the work related).

But maybe we should check what other people think. Perhaps @oantolin and @protesilaos have an opinion?

minad:
> Now that I think more about the narrow API, I get the impression that it was thought around splitting the candidate list into disjoint collections, and one needs to do some contortions when that's not the case.

You are right. I am sorry for the confusion. I missed that the levels are not disjoint. Then my narrowing proposal is not better to what you have. Todo status and priorities are disjoint, but not the levels. Actually not much thought has been put into the narrow API and I prefer to not over do this in order to avoid introducing another complicated filtering system on top of the completion style as I said in my first comment here.

> This would also make the narrowing data reusable. For instance, we could append the consult-outline narrowing data to that of the org commands, and only implement narrowing by priority and todo state in consult-org.el.

The scheme you describe is very similar to `consult-bookmark-narrow` or `consult-register-narrow`. However it is better to not rewrite everything as reusable predicates since then you have to always go over the whole list of predicates. As I said, the narrowing mechanism is limited. And it is probably better to keep it as is in order to not introduce a complicated filtering system on top of the completion style. I think we would not save much code by generalizing what we already have. outline and org is an example, are there more?

consult-org.el
astoff: See https://github.com/minad/consult/pull/276#discussion_r618666135 above.

minad:
> Perhaps not. But whatever you had in mind in oantolin/orderless#30 would be an example, right?

Let's keep it simple as is for now. In oantolin/orderless#30 I had in mind to add these filtering capabilities for exisiting commands depending on the category. Btw one could also implement a narrowing package in the style of Marginalia which enhances existing commands, e.g. adds narrowing/grouping to M-x.

protesilaos:
> @astoff commented on this pull request.
>
> [...]
>
> But maybe we should check what other people think. Perhaps @oantolin and
> @protesilaos have an opinion?

Thanks for considering me!  I do not have an opinion because I have
never done any real work with Org's clocking facilities: I have only
used them once or twice out of curiosity and to test some basic things.

-- 
Protesilaos Stavrou
https://protesilaos.com

consult-org.el
astoff: I see, you mean to use the markers as keys.

astoff:
I've pushed a commit that makes the algorithms O((m+n)log(m)) — I think.

Now, apparently hash tables don't work with markers out of the box, so one has to define a custom hash function, which is a bit crazy. I'm not sure what's the less unreasonable of the options here :-)

minad:
> Now, hash tables aren't meant to work with markers, since they get mutated all the time. But one can define a hash table test which is okay for short-lived hash tables.

This is indeed not reasonable. I had assumed that the history is actually a proper history containing strings which can also be stored. There is indeed such a thing called the `org-clock-stored-history`, but this one is probably not a good fit for the sorting since it may not be up to date.

minad:
I worked a little bit on the PR. There are still a few minor todos. Not all of them must be done, but these are the points where I would like to take a closer loo.

minad:
Currently I am thinking about adding only the `consult-org-heading` command, since this fits consult very well. The other command is basically `consult-org-heading` plus a special scope, a special recent heading sorting and a special action. The problem is that one could in principle write such a command for every command which operates on headings. Not that it would make sense, but I could define a `consult-org-archive` command which is `consult-org-heading` plus a special action. But this is an indication that this clock-in command is not general purpose enough.

Note that one can implement the clock in command somehow like this in the basic form.

~~~ elisp
(defun consult-org-clock-in (...)
    (interactive)
    (org-clock-clock-in
       (list
         (save-window-excursion
           (consult-org-heading ...)
           (point-marker))))
~~~

Then one can add grouping:

~~~ elisp
(defun consult-org-clock-in ()
   (interactive)
   (let ((consult-config '((consult-org-clock-in :group group-by-recent))))
      (org-clock-clock-in
         (list
           (save-window-excursion
             (consult-org-heading pass-the-magic-recent-scope-here)
             (point-marker)))))
~~~

minad:
One more argument - we also have embark actions. One could just put the clock-in on an Embark action which can be invoked on the selected heading. I think I mentioned Embark actions before and this comment https://github.com/oantolin/embark/issues/95#issuecomment-758246901.

minad:
I took the freedom to take a step back and concentrate on the heading command using a separate consult-org-heading category. Now it would be good to prototype how this could be glued together with org. It must be decided which information to attach to the candidates such that it is general purpose. Right now I attach the marker, level, todo and priority to the 'consult-org-heading property.

astoff:
> Currently I am thinking about adding only the consult-org-heading command, since this fits consult very well.

This is what I had imagined initially. Then I realized that live previews are not interesting (and perhaps annoying) in that case, and also that one could do something special about recent clocks. Hence I included this variant. But I don't mind leaving it out, either. Yes, maybe it's wiki-quality stuff.

> Maybe org already attaches properties internally to its heading and we could just reuse them or follow the org convention?

Org doesn't use completing read, so I don't know what you mean exactly. But we are keeping all metadata that comes out of `org-heading-components`, so I guess this is good enough?

minad:
> This is what I had imagined initially. Then I realized that live previews are not interesting (and perhaps annoying) in that case, and also that one could do something special about recent clocks. Hence I included this variant. But I don't mind leaving it out, either. Yes, maybe it's wiki-quality stuff.

Ah, I didn't see that you had disabled previews for the clocking. But fortunately Consult has a flexible configuration mechanism which allows to disable previews per command. So even if you base a clock-in command on consult-org-headline you could still disable the previews.

> Org doesn't use completing read, so I don't know what you mean exactly. But we are keeping all metadata that comes out of org-heading-components, so I guess this is good enough?

I meant maybe org internally keeps metadata for headings in some property attached to headings or in some data structure convention. But as you say we use `org-heading-components` and preserve it mostly. This is maybe just the convention I am looking for.

minad:
I am mostly happy with this now. The question is if we want to prepend the buffer name to the candidates if we are in a multi-buffer scope. Then we can hide the buffer name again using the `:title` function.

astoff:
I really liked the grouping/title thing, and I think the default should be with titles and no prepended buffer name, at least on the vertical UIs.

But we have all information to prepend the buffer name at hand, so it should be easy to add. If it's also easy to remove, I see no harm in this.

minad:
> I really liked the grouping/title thing, and I think the default should be with titles and no prepended buffer name, at least on the vertical UIs.

The idea here is to add it for the filtering. If the UI supports the title function the prefix will be removed again and will be shown only in the title. I work on a patch for this now :)

minad:
Okay, I added the commit. You may want to give it another try and then I will merge this.

astoff:
Looks good to me!

minad:
Thank you for the contribution!

gdindi:
Hi,
I think that commands allowing to navigate the clock history, jumping to the current clocked task and clocking in a task from the history would be very useful. This lacking feature is what keeps me using counsel. I clock most of my time in org-mode and I use these all the time:
    counsel-org-clock-history
    counsel-org-clock-goto
    counsel-org-clock-context

I can help testing them if needed.
Thanks!

astoff:
I've kept a dedicated clock-in command in my config, and I'm actually using it a lot (usually directly, more rarely as an Embark action). I don't know what these 3 variants that @gdindi mentions would be, though.

@minad (cc @oantolin) Aren't Embark action just plain old commands? So how would an Embark clock-in action differ from a `consult-org-clock-in` command, and why would it make more sense for Embark to provide it?

minad:
@astoff I meant that one could execute org-clock commands on the org headings in `consult-org-heading` for example. I don't know what alternative you are using right now, but if I recall correctly this org-clock-in command could be easily replicated based on consult-org-heading + plain org-clock-in in very few lines? Is this correct?

astoff:
> @astoff I meant that one could execute org-clock commands on the org headings in consult-org-heading for example.

That unfortunately doesn't work. One needs a wrapper command. FWIW, A bare-bones `consult-clock-in` is trivial to define, but adding bells and whistles is not.

@gdindi Jump to the current task already exists: `org-clock-goto`.

minad:
> FWIW, a bare-bones consult-clock-in is trivial to define, but adding bells and whistles is not.

Yes, this is what I meant. I don't understand what bells and whistles you would like to have?

astoff:
The main one is a `*Recent*` group — precisely the thing you didn't like, implementation-wise, when I originally made the MR.

Should I maybe add the simple and the fancy versions on the wiki? I just didn't want this to mean the command will never make it into consult-org.el.

minad:
Wiki is the right place for these commands now. I am looking for more general building blocks, for example consult-org-heading is great :)

astoff:
@gdindi Check out https://github.com/minad/consult/wiki#org-clock

gdindi:
@astoff Nice. Thank you!

Issue #75, opened 2020-12-18
Adding package to ELPA


phikal:
Hi,

I was wondering if there were any plans to add this package to GNU ELPA? It seems like it would be a good fit.

minad:
Yes, I thought about it before. I think it is not unreasonable if marginalia, consult etc move closer to upstream since they work well with icomplete too. 

The downside is that it slows down development due to the paperwork. I myself have the papers lying around here and haven't submitted them yet (I should though since I also want to contribute some things to other ELPA projects).

Do you know more about this non-GNU ELPA which is being planned? Is that an acceptable alternative?

phikal:
> The downside is that it slows down development due to the paperwork.

Do you mean because new contributors have to sign the papers? I guess, but in my case it took a week, and just two or three emails, so I can't comment on that.

> Do you know more about this non-GNU ELPA which is being planned? Is that an acceptable alternative?

Maybe this might help: https://git.savannah.gnu.org/cgit/emacs/nongnu.git/tree/README.org#n79

It's still not ready (see https://lists.gnu.org/archive/html/emacs-devel/2020-11/threads.html#00939), but it would also be a good alternative, if the contributors-agreement is an issue. But as I said, the main down-side is that it currently doesn't exist. It could be suggested to add it, but personally, I think that adding it to regular ELPA would be preferable, especially since the project isn't too old and it only gets harder as time goes on.

minad:
@phikal Thank you! I think about it! Regarding ELPA - what do you see as the advantages if the package is added there?

phikal:
Daniel Mendler <notifications@github.com> writes:

> @phikal Thank you! I think about it! Regarding ELPA - what do you see as the advantages if the package is added there?

The main ones are that it is available to all (GNU) Emacs users, and
that it might be merged into the core one day/the core could be adapted
to improve consult.

I, for example, don't use MELPA because it publishes a new version for
each commit, and therefore have to "git clone +
package-install-from-buffer"-it, to test it out. In my case this could
be fixed by adding a tag, because I have MELPA Stable configured, but
even then, it is not something you can depend on (MELPA Stable packages
can depend on MELPA packages, which can break stuff).

When I have a ELPA and a MELPA package, I usually prefer the ELPA one,
because it feels more reliable. It makes sense after all, because
basically the "worlds best Elisp experts" can comment on it, and suggest
idiomatic improvements. But I guess the same will also apply to NonGNU
ELPA.

-- 
	Philip K.

[...]

PR #279, opened 2021-04-20
Show in-buffer previews for consult-history

jakanakaevangeli:
Quick addition of previews to consult-history.

Important note: Previews don't work well in term-mode (for comint and eshell
they work OK). But then again, consult-yank previews don't work well in
term-mode either as this is a general `consult--region-preview` issue with
term-mode.

The problem is that in term mode, when running readline-based programs such as
bash, the last prompt will frequently be redrawn, moving point, markers and
overlays to the beginning or the end of this prompt. The terminal's process
filter might also manually move point over trailing newlines to the end of the
prompt, I'm not sure.

The best solution would probably be to make region previews more robust against
buffer modifications and somehow make them work in term-mode, but I admit that
my experience with Emacs's terminal emulation is lacking.

### DIFF
--- a/README.org
+++ b/README.org
@@ -864,9 +864,10 @@ contributed.
  ;; Note that you should bind <S-up> and <S-down> in the
  ;; `minibuffer-local-completion-map' or `selectrum-minibuffer-map'
  ;; to the commands which select the previous or next candidate.
- (setq consult-config `((consult-theme :preview-key nil)
-                        (consult-buffer :preview-key ,(kbd "M-p"))
-                        (consult-line :preview-key (,(kbd "<S-down>") ,(kbd "<S-up>")))))
+ (with-eval-after-load 'consult
+   (add-to-list 'consult-config '(consult-theme :preview-key nil))
+   (add-to-list 'consult-config '(consult-buffer :preview-key ,(kbd "M-p")))
+   (add-to-list 'consult-config '(consult-line :preview-key (,(kbd "<S-down>") ,(kbd "<S-up>")))))
  #+end_src
 
  Generally it is possible to modify commands for your individual needs by the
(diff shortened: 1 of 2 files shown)


minad:
Okay, good idea.

* Maybe the minibufferp check for region preview should go directly to `consult--region-preview`.
* Do we need an extra face? Maybe these region preview faces could be unified. Modus at least does not overwrite them and this is my reference theme and also the theme I use.
* Regarding term mode - I agree it would be good to support this better. But one can also disable the preview with an advice. Maybe we should add support for lambdas in `consult-config` as an easier to use dynamic configuration mechanism. 

Since this package is going to be submitted to ELPA at some point (#75): Do you have the fsf copyright assignment? In the current form, this PR is small enough to not fall within these rules, but all contributions sum up.

jakanakaevangeli:
> * Maybe the minibufferp check for region preview should go directly to
> `consult--region-preview`.

Agree, done.

> * Do we need an extra face? Maybe these region preview faces could be
> unified.

Agree, they are mostly the same thing IMO. I unified them into a new
consult-preview-insertion face.

> * Regarding term mode - I agree it would be good to support this
> better. But one can also disable the preview with an advice. Maybe we
> should add support for lambdas in `consult-config` as an easier to use
> dynamic configuration mechanism.

I went with supporting lambdas in consult-config and made it backward
compatible. What do you think?
Another possibility would be to disregard my 4 latest commits that add
support for lambdas and simply do something like this

    (add-hook 'term-mode-hook
      (lambda ()
        (setq-local consult-config
                    `((consult-yank :preview-key nil)
                      (consult-yank-pop :preview-key nil)
                      (consult-completion-in-region :preview-key nil)
                      (consult-history :preview-key nil))
                      ***@***.***))))

> Since this package is going to be submitted to ELPA at some point
> (#75): Do you have the fsf copyright assignment?

I have. I pushed the new commits with my real name.

minad:
Thanks! I think it is better to keep it simple for now. The term hook is good enough.

minad:
Merged partially

PR #292, opened 2021-04-27
Add consult-line-thing-at-point

condy0919:
It's modeled after the command `isearch-forward-thing-at-point`

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -295,6 +295,16 @@ command options."
 Each element of the list must have the form '(char name handler)."
   :type '(repeat (list character string function)))
 
+(defcustom consult-line-thing-at-point '(region url symbol sexp)
+  "A list of symbols to try to get the \"thing\" at point.
+
+Each element of the list should be one of the symbols supported
+by `thing-at-point'. This variable is used by the command
+`consult-line-thing-at-point' to yank the initial \"thing\" as
+text to the search string."
+  :type '(repeat (symbol :tag "Thing symbol"))
+  :group 'consult)
+
 (defcustom consult-config nil
   "Command configuration alist, which allows fine-grained configuration.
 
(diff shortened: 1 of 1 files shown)


minad:
Hi thank you for your PR. To be honest, I am not sure if the addition of such a command is advised in Consult. One can already start `consult-line` via Embark by placing the point on a thing. Furthermore the symbol at point is added to the future history. We had `consult-line-symbol-at-point` and removed it since it was not flexible enough. Your command provides a bit more functionality at the cost of a configuration option. While Consult is a kitchen sink, the idea is to stay a bit conservative regarding the addition of command variants, which only adjust some arguments, and focus on essential building block commands. Otherwise it is not clear what is to be added, do we also need `consult-imenu-thing-at-point` and `consult-outline-thing-at-point`? Users can then easily build special commands on top according to their preference and keep them around in their init.el. We maintain a list of additional commands and configurations in the [wiki](https://github.com/minad/consult/wiki). We could add this command to the wiki:

~~~ elisp
(defun consult-line-thing-at-point (&optional start)
  (interactive (list (not (not current-prefix-arg))))
  (let ((thing (or (seq-some (lambda (thing) (thing-at-point thing t))
                             '(region url symbol sexp))
                   (user-error "No thing at point"))))
    (when (use-region-p)
      (deactivate-mark))
    (consult-line thing start)))
~~~

However the search functionality is important enough, so I am not yet fully decided. And since the same command exists in isearch, there seems to be some common need for this. But then note that `consult-line` fundamentally is a line based command. It will never work well with multiple matches on a single line. I rather see it as an in-buffer grep but not as an isearch replacement. I think both have their reason for existence and both are useful.

condy0919:
> And since the same command exists in isearch, there seems to be some common need for this.

I agree with that. There are `swiper-isearch-thing-at-point`, `helm-swoop` in the wild.

condy0919:
Closed

PR #297, opened 2021-04-29
Pause async processes when entering a recursive minibuffer

### LINKED ISSUE #272: Check why running nested consult-ripgreps do not work
When starting multiple long running consult-ripgreps in recursive minibuffers it does not work for some reason with Vertico. Display update issue? Some state is not isolated well enough?

jakanakaevangeli:
Fixes #272.
In the linked issue, you suggested having a faster filter function to keep up
with multiple processes output. Trying this with the default filter, the
problem sadly persisted. That is why I opted for this approach where the
processes for non-innermost minibuffers are paused (by not accepting their
output).

The down side is that these processes stay paused until the inner minibuffer is
quit, even if its process finishes (or if it doesn't have a process in the
first place).

Ideally, we would make Emacs prioritize accepting inner minibuffer's process
output. I haven't found a way to set such a priority aside from deleting outer
minibuffer's process and creating it again (Emacs seems to prioritize older
processes), with the obvious disadvantage that we throw away results when
deleting the process.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -1268,7 +1268,7 @@ the comma is passed to ASYNC, the second part is used for filtering."
 ASYNC is the async function which receives the candidates.
 CMD is the command argument list.
 PROPS are optional properties passed to `make-process'."
-  (let ((proc) (last-args) (indicator) (count))
+  (let ((proc) (last-args) (indicator) (count) (proc-filter))
     (lambda (action)
       (pcase action
         ("" ;; If no input is provided kill current process
@@ -1299,17 +1299,18 @@ PROPS are optional properties passed to `make-process'."
                    :noquery t
                    :command args
                    :filter
-                   (lambda (_ out)
-                     (when flush
-                       (setq flush nil)
-                       (funcall async 'flush))
-                     (let ((lines (split-string out "\n")))
-                       (if (not (cdr lines))
-                           (setq rest (concat rest (car lines)))
-                         (setcar lines (concat rest (car lines)))
-                         (setq rest (car (last lines)))
-                         (setq count (+ count (length lines) -1))
-                         (funcall async (nbutlast lines)))))
+                   (setq proc-filter
+                         (lambda (_ out)
+                           (when flush
+                             (setq flush nil)
+                             (funcall async 'flush))
+                           (let ((lines (split-string out "\n")))
+                             (if (not (cdr lines))
+                                 (setq rest (concat rest (car lines)))
+                               (setcar lines (concat rest (car lines)))
+                               (setq rest (car (last lines)))
+                               (setq count (+ count (length lines) -1))
+                               (funcall async (nbutlast lines))))))
                    :sentinel
                    (lambda (_ event)
                      (when flush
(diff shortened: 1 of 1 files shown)


minad:
Thank you for looking into this! It is unfortunate that the `internal-default-process-filter` does not fix the issue. I had assumed that the switch to the elisp bytecode interpreter introduces the problem. I think the proper fix would be to ensure within Emacs that it receives input for all processes in some kind of fair scheduling scheme. It could still prioritize the inner/newer processes. I am unsure if we really need a fix for this problem in Consult for this since it is pretty much an edge case. I still think it would be good to somehow report this as an Emacs bug, but it is necessary to create a minimal reproducible for this.

minad:
Regarding multiple processes - does the problem persist if one has multiple slow long-running processes, which slowly produce their results? Or is it only a problem if you have processes like ripgrep as in this case which flood Emacs with their output? Furthermore is the problem related to the minibuffer (as I had speculated at some point) or does it also happen if I start a bunch of async processes in the background. Depending on the answers to these questions the issue is more or less serious and I would have assumed that someone already observed it.

jakanakaevangeli:
Yes, I tried multiple slow processes and emacs handled their fine in
parallel just fine, the problem only occurs with fast output processes.
The issue is not related to minibuffers.

Here is a simple reproducer, which shows that the second buffer doesn't
grow until we kill the first process:

    (progn
      (setq p-a (make-process :name "yes-a"
                              :command '("yes")
                              :connection-type 'pipe
                              :buffer (setq a (generate-new-buffer " *a*"))))
      (setq p-b (make-process :name "yes-b"
                              :command '("yes")
                              :connection-type 'pipe
                              :buffer (setq b (generate-new-buffer " *b*"))))
      (run-at-time 1 1
                   (lambda ()
                     (message "size a: %s\nsize b: %s" (buffer-size a) (buffer-size b)))))

I'm currently not really motivated enough to report this as Emacs bug,
so if you believe the problem is with Emacs, fell free to report,
thanks.

minad:
> Yes, I tried multiple slow processes and emacs handled their fine in
parallel just fine, the problem only occurs with fast output processes.
The issue is not related to minibuffers.

Thanks for your help!

> I'm currently not really motivated enough to report this as Emacs bug,
so if you believe the problem is with Emacs, fell free to report,
thanks.

That's fine. I will keep it in mind and maybe report it after investigating a bit more.

minad:
I reported the bug: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=48118

jakanakaevangeli:
Thanks, don't worry, I've already butted in.

minad:
Thanks. I am not actually worried. Your comment on the tracker is reasonable. The for loop is the "scheduling algorithm" one would naturally use with pselect/epoll/kqueue/etc. The actual scheduling work is then done by the OS. However Emacs could take the standpoint that they want to actually turn off the OS and do its own predictable "non-scheduling". Emacs also supports platforms without preemptive scheduling (DOS etc, does it still support that?), so it is not totally unreasonable that the Emacs implementation does things which seem weird on modern platforms.

minad:
I think you meant I should not worry because I didn't put you in cc ;)

Issue #299, opened 2021-05-01
consult-imenu may jump to a wrong position when items have same names


zbelial:
As it searches candidate using `assoc`, which returns the first item matched.

minad:
Yes. But notably the built-in `imenu` command also cannot handle with duplicates. The problem here is that duplicates in the imenu datastructure are probably not well defined. Therefore I am not sure if we should handle them in Consult. In which realistic scenarios do you see duplicate items?

You can try the following unrealistic examples with `imenu` and `consult-imenu`:

~~~ elisp
(defun x ())
(defun x ())
(defun y ())
~~~

~~~ org
* header
* header
* header
~~~

zbelial:
It's not unusual to see duplicate items when using programming languages that support override, such as java, cpp, etc.
Actually I found this problem when I wrote Java code.

minad:
> It's not unusual to see duplicate items when using programming languages that support override, such as java, cpp, etc.
Actually I found this problem when I wrote Java code.

Okay, I agree that this is an important use case then if many imenu backends generate ambiguous items. (Still it would be better if these backends disambiguate the items with some signatures).

Issue #212, opened 2021-02-07
Ivy support


zbelial:
I just tried consult with `completing-read-function` set to `ivy-compelting-read`, but I found that narrowing in `consult-buffer` does not work.  

File `consult-icomplete.el` provides a function `consult-icomplete--refresh` to refresh candidates, so to implement a similar function is enough?  If so, what this function should do?

Thanks.

minad:
Right now there is no Ivy support. Iris questionable if Ivy support is really useful to have since Ivy brings its own Counsel commands. However in principle there is nothing preventing us from having Ivy support here and it would be nice if all the packages could work together.

* [x] Basic functionality is working with Ivy
* [ ] Preview: Needs special Ivy support. A consult--completion-candidate-hook must be provided.
* [ ] Narrow: Needs special Ivy support, the narrowing is implemented via the completion predicate. Requires advices on Ivy functions or modifications of Ivy.
* [ ] Async: Needs special Ivy support. A consult--completion-refresh-hook must be provided. Additionally requires advices on Ivy functions or modifications of Ivy since Ivy does not reload async functions.

Bringing Ivy support here will require someone with Ivy experience. I looked in it for a few minutes but could not figure it out quickly. 

Furthermore it would be nice to have Marginalia support for Ivy. See https://github.com/abo-abo/swiper/issues/2780.

zbelial:
Thanks for you reply and all the detailed information, @minad .
I must say that implementing all of these functions is beyond my ability now, but maybe I can give it a try, eg. to implement narrow as the first step, which I like very much. I guess this is a time-consuming task for me, so you cannot see good results soon.

minad:
@zbelial Sure, give it a try. Note that I will only merge a version here which provides full support. I added two more points regarding consult-line and consult--async which may also require special casing.

zbelial:
Small progress:
Preview and narrow work now, but the code is ugly as it use some of ivy's internal functions and variables, and I'm not sure whether it's effective enough. 
Here is the code:
```elisp
            (with-eval-after-load 'ivy

              (defun consult-ivy--candidate ()
                "Return current ivy candidate."
                (and ivy-mode (nth ivy--index (ivy--filter ivy-text ivy--all-candidates)))
                )

              (defun consult-ivy--refresh ()
                "Refresh ivy view, keep current candidate selected if possible."

                (when ivy-mode
                  (let ((ivy-old-text ivy-text)
                        (cands nil))
[...]

PR #300, opened 2021-05-02
check marker when looking up imenu item

zbelial:
This PR aims to solve issue #299 , it puts the marker as a text property to each item.

### DIFF
--- a/consult-imenu.el
+++ b/consult-imenu.el
@@ -82,9 +82,9 @@ TYPES is the mode-specific types configuration."
            (consult-imenu--flatten next-prefix next-face (cdr item) types))
        (let* ((name (car item))
               (key (if prefix (concat prefix " " (propertize name 'face face)) name))
-              (payload (cdr item)))
-         (list (cons key
-                     (pcase payload
+              (payload (cdr item))
+              marker)
+         (setq marker (pcase payload
                        ;; Simple marker item
                        ((pred markerp) payload)
                        ;; Simple integer item
@@ -96,7 +96,10 @@ TYPES is the mode-specific types configuration."
                         (nconc
                          (list pos #'consult-imenu--special (current-buffer) name fn)
                          args))
-                       (_ (error "Unknown imenu item: %S" item))))))))
+                       (_ (error "Unknown imenu item: %S" item))))
+         (setq key (propertize key 'marker marker))
+         (list (cons key
+                     payload)))))
    list))
 
 (defun consult-imenu--compute ()
@@ -185,10 +188,21 @@ The symbol at point is added to the future history."
            (eq (get-text-property 0 'consult--type (car cand)) consult--narrow))
          narrow))
       :category 'imenu
-      :lookup #'consult--lookup-cons
+      :lookup #'consult-imenu--lookup-cons
       :history 'consult-imenu--history
       :add-history (thing-at-point 'symbol)
       :sort nil))))
+
+(defun consult-imenu--lookup-testfn (a b)
+  (let ((am (get-text-property 0 'marker a))
+        (bm (get-text-property 0 'marker b)))
+    (and (equal a b)
+         (equal am bm))))
+
+(defun consult-imenu--lookup-cons (_ candidates cand)
+  "Lookup CAND in CANDIDATES alist, return cons."
+  (assoc cand candidates #'consult-imenu--lookup-testfn))
+
 
 ;;;###autoload
 (defun consult-imenu ()


minad:
This is not a correct patch since completing-read does not retain the properties.

zbelial:
Sorry for making this incorrect patch and thanks for letting me know the reason why it's incorrect. 
I use `ivy-mode`, it sets `completing-read-function` to `ivy-completing-read`, which seems to retain properties, so it worked when I tested this patch.  I forgot to test it with the default `completing-read`. Sorry.

minad:
No worries. I appreciate your bug report and your attempt to fix the issue. Generally try to test your patches with completion systems which are fully supported by Consult (Selectrum, Vertico, Icomplete, default completion). Ivy is not yet supported (#212) as you are aware. Out of interest, which commands do you miss in the Selectrum/Consult/Embark ecosystem in contrast to the Ivy/Counsel set of packages?

zbelial:
> Out of interest, which commands do you miss in the Selectrum/Consult/Embark ecosystem in contrast to the Ivy/Counsel set of packages?

There are several things that are important to me: 
1. Ivy's :preselect, which makes me preselect a item. 
    I'll use `counsel-imenu` as a use case. When I invoke `counsel-imenu` and the cursor is in a function, for example `Func`, I want to preselect it, but not the first one(You know functions that are relevant are close in files, so I can select them more easily).  I use this feature in a command to show pdf files' table of  contents too.
2. If candidates are a list of cons, `ivy-read` will return the whole cons(including properties), not just the car of it.  `consult-imenu` can also benefit from this kind of feature I think. 
3. I can use different filters for different commands. 

These are the features I miss in Consult/Selectrum. 
Maybe ivy's internal is more sophisticated, I find `ivy-read` is powerful and convenient. 

As a user, what I want the most from Consult are preview and narrow, that's the reason why I want to add ivy support to Consult (I wish I can finish it some day, but Emacs completing system is complex for me, and Ivy's internal is really sohisticated).

zbelial:
Actually I wish I can write commands use Emacs completing system's api but not a specific one (such as ivy-read) some day.  And I know you're making some improvement in this aspect (I've seen some patches you commit in emacs mailing list). Thanks.

minad:
I am not sure I understand these points. These are not strong arguments. All of them are possible in Consult/selectrum/vertico, but maybe not in exactly the same form.

1. How is :preselect different from :default which is supported by consult--read and completing-read?
2. If you want conses set the :lookup function to consult--lookup-cons
3. Different completion styles? This is possible by setting completion-category-defaults or completion-category-defaults.

zbelial:
Most ivy commands I use(swiper, counsel-imenu, counsel-outline, counsel-rg, counsel-yank-pop)  have equivalents with similar behaviour in Consult/Selectrum, I've not noticed any significant differences.  I choose ivy because of its api ( I've wrote several commands with it), which I'm more familiar with and is more intuitive IMO.

Actually I don't have enough knowledge of Emacs completion, I've made a little effort to learn it, but I can't say I succeeded.  So, the following examples and explainations may not be corrent/good enough. 

> * How is :preselect different from :default which is supported by consult--read and completing-read?
The following examples can show the difference.  In both examples, do not input anything but just press <return> .
``` elisp 
;; use ivy-read 
(setq r (ivy-read "test: "
         `((,(propertize "foobar1" 'marker "abc") 1) ("barfoo" 2) ("foobaz" 3) ("foobar1" 4) )
         :preselect 0
         ))
=> #("foobar1" 0 7 (marker "abc"))
```

```elisp
;; use completing-read
(setq r (completing-read "Test: "
         '(("foobar1" 1) ("barfoo" 2) ("foobaz" 3) ("foobar1" 4))
         nil nil nil nil "foobar1"
         ))
=> "foobar1"
``` 
With `ivy-read`, you can know exactly which candidate is returned , but with `completing-read`, you have no idea of which "foobar1" it is.  I know you can use, like consult--lookup-cons, to look up the candidate, but you have to deduplicate the candidates, like what you did in `consult-imenu`, to get the exact one. 

> * If you want conses set the :lookup function to consult--lookup-cons.
It has minor drawback as described above. 

> * Different completion styles? This is possible by setting completion-category-defaults or completion-category-defaults, which are an Emacs built-in configuration variables.
I had no idea about this, and I cannot test it right now, but I'll try it. Thanks.

minad:
Okay, your point here is that Ivy supports duplicate items. I made a proposal on emacs-devel to support this duplicates also with `completing-read`, but I am not yet sure if I will proceed with it. Ivy implements this as an extension, which is okay since it uses its own api. The Consult set of packages cannot make use of this functionality since we have to rely on the built-in functionality. I have somehow come to the conclusion that deduplication is not that bad after all (as I did for `consult-imenu`, `consult-line` etc.)

zbelial:
Another thing is `completing-read` does not retain properties(as you pointed out in a previous comment), which ivy supports. It's a useful feature in some cases. 

All in all, ivy is the first completing system I'm familiar with as a user, and I get used to it as a daily toolset. This is the reason why I'm still stick to it now.

minad:
> Another thing is completing-read does not retain properties(as you pointed out in a previous comment), which ivy supports. It's a useful feature in some cases.

This is not really "another thing". It is just the same thing as - the completion system supports duplicates, since string equality does not take properties into account. If `completing-read` would allow duplicates the preservation of text properties would b implied.

> All in all, ivy is the first completing system I'm familiar with as a user, and I get used to it as a daily toolset. This is the reason why I'm still stick to it now.

I understood that. You should stay with it if you like it. I wonder why you are even considering Consult which explicitly does not support Ivy, given that there is Counsel which is a strong library. I guess there is nothing wrong with having multiple libraries installed doing similar things, many people have both Helm and Ivy installed for example. But there is also a large group of people who somehow try to use a streamlined Emacs, with a focus on minimalism and consistency.

zbelial:
> I wonder why you are even considering Consult which explicitly does not support Ivy, given that there is Counsel which is a strong library.

I'd like to try out new good packages, so I installed Selectrum and then Consult when I found them. And I like Consult's narrow and preview(I found preview is pretty useful when I want to have a quick look at the content of a buffer or a function but don't want to switch to it).  I'm one of the people who have Helm and Ivy (and Consult) installed at the same time.

PR #303, opened 2021-05-05
Add system recent files to `consult-recent-file`.

okamsn:
Hello Minad,

This change adds the option to include files recently used by other programs to the candidate list in `consult-recent-file`.

This is based on a similar option in [`counsel-recentf`](https://github.com/abo-abo/swiper/blob/99bf708b5fff21c9434c9f5a0336f236066fd002/counsel.el#L2309) and in the [Selectrum wiki](https://github.com/raxod502/selectrum/wiki/Useful-Commands#including-files-listed-in-the-xdg-specification-used-on-linux). I saw that they were mentioned on Reddit, and thought that this feature could be worth including in Consult.

It currently only supports the XDG spec used on Linux. Unlike those other versions, in the function `consult--recent-files-mixed-candidates`, I included a `cl-case` to allow for other systems in the future. I myself am only using Linux these days, so I can't test what's needed for MacOS or Windows.

To parse and navigate the XML data file named by the XDG spec, this change needs to use only a few functions, but those functions are spread out over a few libraries. Do you want to make any changes regarding the `require` uses?

What would you like changed?

Thank you.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -52,6 +52,10 @@
 (require 'recentf)
 (require 'seq)
 
+(autoload 'dom-attr "dom")
+(autoload 'dom-by-tag "dom")
+(autoload 'xdg-data-home "xdg")
+
 (defgroup consult nil
   "Consulting `completing-read'."
   :group 'convenience
@@ -298,6 +302,10 @@ command options."
 
 Each element of the list must have the form '(char name handler)."
   :type '(repeat (list character string function)))
+
+(defcustom consult-include-system-recent-files nil
+  "Whether to include files used by other programs in `consult-recent-file'."
+  :type 'boolean)
 
 (defcustom consult-config nil
   "Command configuration alist, which allows fine-grained configuration.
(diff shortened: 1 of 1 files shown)


minad:
Thanks, I've also seen that feature mentioned on Reddit. A few comments:

* The minimum requirement for inclusion is that it is supported and tested on the major platforms (Linux, Mac, Win). Generally I don't support platform-specific commands here, since I would also have difficulties in testing them.
* I am not sure if I even want this command, since I am not sure why it is helpful to open files of non-Emacs file types in Emacs. I think it would hurt more than help. I assume that the common Emacs user edits *everything text-based* in Emacs. Then all the binary formats, images, etc, which are edited in other applications are not really useful here.
* I am not fond of the additional requires. Currently Consult relies only on core functionality, now you add dom, xdg, xml and url-util for a command I don't see being used often.
* There is also `consult--source-file` which includes the recent files. If it is decided to support recent files in Consult this buffer source must be adjusted too. It would be even better to provide a separate source. 

So overall the requirements for inclusion of such a feature in Consult are a bit high. I omitted this feature initially due to these reasons.

okamsn:
> So overall the requirements for inclusion of such a feature in Consult are a bit high. I omitted this feature initially partly due to these reasons. Maybe this command is better maintained in the wiki or in an external package.

I added the information to the Wiki. If the situation improves, adding it can be revisited later.

> Thanks, I've also seen that feature mentioned on Reddit. A few comments:
> 
>  * The minimum requirement for inclusion is that it is supported and tested on the major platforms (Linux, Mac, Win). Generally I don't support platform-specific commands here, since I would also have difficulties in testing them.

This is reasonable. I was hoping that others would come along and add support for other systems, but this feature is already niche. 
 
>  * I am not sure if I even want this command, since I am unsure if it is helpful to open files of non-Emacs file types in Emacs. Maybe it would hurt more than help. I assume that the common Emacs user edits _everything text-based_ in Emacs; this is at least the use case I am optimizing for. Then all the binary formats, images, etc, which are edited in other applications are not really useful here.

My use case is that I don't always use Emacs for working with files. For example, I often edited MATLAB files in their own IDE, even though I find the editing experience limited compared to Emacs. Other times, I might wish to re-view an image that I was looking at a few days ago when taking notes in Org.

Adding these files into the standard Emacs interface avoids "mental hiccups" when I try to view these files in Emacs. If they are available in the list of candidates, then it is less to remember.
 
>  * I am not fond of the additional requires. Currently Consult relies only on core functionality, now you add dom, xdg, xml and url-util for a functionality I don't see being used often.

Upon review, the only functions that aren't `autoload`-ed are `dom-attr`, `dom-by-tag`, `xdg-data-home`, and `xml-parse-region` (only used if Emacs compiled without `libxml2`). 

I had assumed that I needed to `require` the libraries. I have tried just `autoload`-ing these functions, and it seems to work.
 
>  * There is also `consult--source-file` which includes the recent files. If it is decided to support xdg recent files in Consult this buffer source must be adjusted too. It would be even better to provide a separate source.

Added on the wiki entry.

minad:
> This is reasonable. I was hoping that others would come along and add support for other systems, but this feature is already niche.

Yes, but this should happen before it gets added to the package. If it gets added over time to the wiki version we can reconsider.

> My use case is that I don't always use Emacs for working with files. For example, I often edited MATLAB files in their own IDE
> ...
> Adding these files into the standard Emacs interface avoids "mental hiccups" when I try to view these files in Emacs. If they are available in the list of candidates, then it is less to remember.

Yes, this is understandable. But as soon as you have opened the file once in Emacs it will be present in the recentf-list, so there is only a tiny advantage in reading the xdg list. It may be useful if you are inspecting a large number of files partially with Emacs and partially with some other program, such that you open each file only once in Emacs. Then it is certainly nice to have the ability to switch back and forth. 

For many users the overall advantages seem not that significant, otherwise the feature would have been requested more often. As far as I know it has not been mentioned before on the wishlist (#6), while other features have been mentioned multiple times. Then the cost value ratio is not good. We have to do this xml parsing and don't even support the other platforms yet. Maybe it would be better if Emacs included some xdg/win/mac recent file importing functionality itself such that it could synchronize its recentf list with the global list. Maybe this has been proposed before but didn't make it into Emacs for similar reasons as I stated here.

Issue #251, opened 2021-03-12
Is there an option to initially narrow?


antifuchs:
I would love to use `consult-buffer` with a setup similar to what `bufler` workspaces do (limiting the selection of buffers to the buffers "in" the workspace in my frame -- usually the active projectile project). That means that when I invoke `consult-buffer`, I'd love to have the selection of buffers already pre-narrowed to the current projectile project.

So far, I've tried:

```elisp
(setq consult-config '((consult-buffer :initial "p ")))
```

But that only inserts the literal string "p " on the input field.

This felt interesting, but results in no narrowing (and you can't narrow at all afterwards):
```elisp
(setq consult-config
      '((consult-buffer :narrow '(consult--source-project-buffers))))
```

So - is what I would like to accomplish possible with consult?

antifuchs:
It just occurred to me that I wished for something similar with `consult-flycheck` also: I'd love to set it up such that if there are errors, it's narrowed to the errors in the files initially.

minad:
Initial narrowing is supported with a little bit of configuration effort (https://github.com/minad/consult/wiki#start-command-with-initial-narrowing). See also the prior discussion in #203.

minad:
There is only one downside of the current approach - if you are not in a project, you probably don't want to narrow down to the empty list :laughing:  This means a little bit of additional logic is needed.

antifuchs:
Thanks so much - I entirely missed this section in the wiki. That's exactly what I wanted & agreed about the additional logic. I'm writing that up now (-:

antifuchs:
In case anyone else wants a similar thing as me, here's what I did to get flycheck errors narrowed to errors only, if there are any:

``` elisp
(defun asf--consult-initially-narrow-buffers ()
  (let ((buffer (if (minibufferp)
                    (window-buffer (minibuffer-selected-window))
                  (current-buffer))))
    (cond
     ;; Narrow buffers to "F" ones (bufler):
     ((eql this-command 'consult-buffer)
      (setq unread-command-events (append unread-command-events (list ?F 32))))

     ;; Narrow flycheck diagnostics to errors if there are any:
     ((and (eql this-command 'consult-flycheck)
           (with-current-buffer buffer
             (cl-find 'error flycheck-current-errors :key 'flycheck-error-level)))
      (setq unread-command-events (append unread-command-events (list ?e 32)))))))

(add-hook 'minibuffer-setup-hook #'asf--consult-initially-narrow-buffers)
```

[...]

Issue #203, opened 2021-02-04
consult--read able to initiate already narrowed down


0x462e41:
I think it would be useful to be able to evoke this function with the sources already narrowed. By doing this, it would be possible, for example, to call `consult-buffer` already restricted to the buffer subset and, only if desired, remove the narrowing or even change it.

Sorry if it's already possible, I haven't read the code base entirely yet.

minad:
This is not possible out of the box, but you can easily achieve this behavior as follows:

~~~ elisp
;; Configure initial narrowing per command
(defvar consult-initial-narrow-config
  '((consult-buffer . ?b)))

;; Add initial narrowing hook
(defun consult-initial-narrow ()
  (when-let (key (alist-get this-command consult-initial-narrow-config))
    (setq unread-command-events (append unread-command-events (list key 32)))))
(add-hook 'minibuffer-setup-hook #'consult-initial-narrow)
~~~

minad:
Added to the wiki, https://github.com/minad/consult/wiki#start-command-with-initial-narrowing. Please let me know if it works for you.

0x462e41:
Yes, it works. Thanks!

I find it useful when we want to combine several sources in the same "completing read", but generally in our workflow, we usually narrow it to a subset. For me, an example might be what I described: when I call `consult-buffer`, I usually instantly narrow to the buffer subset, although I sometimes want to select a recent file. So, just using "f SPC" on these occasions is easier than almost always use "b SPC". I do this because my job involves opening several files, sometimes with a similar name, and dealing with them. In these cases, mixing buffers and recent files is confusing, making it difficult to discern which files I am currently dealing with and which files I have already dealt with.

Another example is to narrow `consult-imenu` programmatically to show functions or variables, depending on the context (point position, for example), but also being able to change the narrowing if needed.

Anyhow, the solution you included on the wiki is good enough, considering that not all users would benefit from it, I think.

minad:
I just type the name of the buffer or the file and orderless gives me the candidates I want to see :) But maybe if you have many similar candidates as in your case it becomes useful to use narrowing for that. For now I think it is better to not included this by default since it will probably be used very rarely. 

[...]

PR #316, opened 2021-05-26
Add helper macro for pre-filling the minibuffer

Zetagon:
Hello,

I created this helper macro for prefilling the minibuffer based on this: https://github.com/minad/consult/wiki#start-command-with-initial-narrowing
I wasn't sure if this project would want this, but I thought I would file this pull request anyways in case it is deemed useful.

There are two things I wasn't sure of: 1) Naming of identifiers and 2) where to put the new definitions.

I have signed FSF copyright papers for Emacs.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -525,6 +525,30 @@ ARGS is a list of commands or sources followed by the list of keyword-value pair
 
 ;;;; Helper functions and macros
 
+(defvar consult-prefill-minibuffer-config
+  '()
+  "An alist of (COMMAND . STRING) pairs describing which commands gets prefilled with what string.
+If COMMAND prompts the user via the minibufer, the minibuffer
+gets the string STRING as initial input. See
+`consult--prefill-minibuffer' and
+`consult-with-prefilled-minibuffer'")
+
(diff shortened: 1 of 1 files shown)


minad:
There is already `:initial` which can be `consult-customize`d per command. There is one thing which is lacking though - you cannot use this to implement initial narrowing. A cleaner way to implement initial narrowing would be to add an `:initial-narrow` option to `consult--read` (See #251 and #203). So far I've not seen a use case which needs this but it has been requested a few times now, so one may want to add this.

oantolin:
Can't people who want initial narrowing do whatever `consult-buffer` does to hide hidden buffers and apply that to all groups they want hidden by default?

Zetagon:
For me personally, the reason I created this macro is that it makes it easier for me to provide initial narrowing for existing commands, and only for specific calls to that command. For that usecase it's more likely to be used in people's personal configs I guess.

But yeah if you are tentative about this I have no problem with just dumping it in my own config.

minad:
> But yeah if you are tentative about this I have no problem with just dumping it in my own config.

Okay, closing this then.

PR #352, opened 2021-07-06
consult-completing-read-multiple

minad:
@bdarcus @oantolin 

This is the generalized version of `simple-crm`. While this crm implementation is not the perfect solution and may still have other subtle issues, I am actually quite happy with it. I think it is a usability improvement in particular in combination with UIs like Vertico  or Selectrum, but it also works well with default completion. The function is a pretty cheap and non-intrusive addition, independent of the completion UI. This is much more maintainable, in contrast to baking some special crm support directly in the completion UI like Vertico.

My problem here is that I could not get this to work with Embark until now. My main concern is not the multi selection target finder, but the ability to allow `completing-read-multiple` commands to be used as Embark actions. Example: `M-x describe-face RET <select-face> M-.` or `M-x describe-face RET <select-face> C-. M-x describe-face RET`, where `M-./C-.`  are `embark-dwim/act` respectively. The issue is that the action gets stuck in the CRM loop.

Please give it a test. Maybe you have a good idea how to enable Embark support?

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -2,6 +2,11 @@
 #+author: Daniel Mendler
 #+language: en
 
+* Development
+
+- =consult-mark=, =consult-global-mark=: Added optional marker list argument
+- =consult-completing-read-multiple=: New function
+
 * Version 0.9 (2021-06-22)
 
 - Add =consult-preview-excluded-hooks=
--- a/README.org
+++ b/README.org
@@ -343,6 +343,7 @@ their descriptions.
  #+findex: consult-apropos
  #+findex: consult-file-externally
  #+findex: consult-completion-in-region
+ #+findex: consult-completing-read-multiple
  #+findex: consult-theme
  #+findex: consult-man
  #+findex: consult-preview-at-point
@@ -384,6 +385,9 @@ their descriptions.
    buffer to the minibuffer, the server does not receive the updated input. Lsp
    completion should work with Corfu or Company though, which perform the
    completion directly in the original buffer.
+  - =consult-completing-read-multiple=: Enhanced drop-in replacement for
+    =completing-read-multiple= which works better for long candidates.
+
 * Special features
   :properties:
   :description: Enhancements over built-in `completing-read'
@@ -834,6 +838,9 @@ contributed.
      ;; Use Consult to select xref locations with preview
      (setq xref-show-xrefs-function #'consult-xref
            xref-show-definitions-function #'consult-xref)
+
+     ;; Optionally replace `completing-read-multiple' with an enhanced version.
+     (advice-add #'completing-read-multiple :override #'consult-completing-read-multiple)
 
      ;; Configure other variables and modes in the :config section,
      ;; after lazily loading the package.
--- a/consult-icomplete.el
+++ b/consult-icomplete.el
@@ -27,14 +27,14 @@
 (require 'consult)
 (require 'icomplete)
 
-(defun consult-icomplete--refresh ()
-  "Refresh icomplete view, keep current candidate selected if possible."
+(defun consult-icomplete--refresh (&optional reset)
+  "Refresh icomplete view, keep current candidate unless RESET is non-nil."
   (when icomplete-mode
     (let ((top (car completion-all-sorted-completions)))
       (completion--flush-all-sorted-completions)
       ;; force flushing, otherwise narrowing is broken!
       (setq completion-all-sorted-completions nil)
-      (when top
+      (when (and top (not reset))
         (let* ((completions (completion-all-sorted-completions))
                (last (last completions))
                (before)) ;; completions before top
--- a/consult-selectrum.el
+++ b/consult-selectrum.el
@@ -60,11 +60,10 @@ See `consult--completion-filter' for arguments PATTERN, CANDS, CATEGORY and HIGH
 (defun consult-selectrum--refresh (&optional reset)
   "Refresh completion UI, keep current candidate unless RESET is non-nil."
   (when selectrum-is-active
-    (if consult--narrow
-        (setq-local selectrum-default-value-format nil)
-      (kill-local-variable 'selectrum-default-value-format))
+    (when consult--narrow
+      (setq-local selectrum-default-value-format nil))
     (when reset
-      (setq selectrum--history-hash nil))
+      (setq-local selectrum--history-hash nil))
     (selectrum-exhibit (not reset))))
 
 (defun consult-selectrum--split-wrap (orig split)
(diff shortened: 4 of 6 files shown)


bdarcus:
Tested, and works as expected (though I have not yet tested with very large candidate lists). Nice!

Still wish we could find a workaround for the "double-RET" issue.

minad:
> Still wish we could find a workaround for the "double-RET" issue.

This is easy to solve by adding custom keybindings. For Vertico one could bind

* TAB -> If candidate is selected, submit this candidate. If no candidate is selected, do nothing.
* RET -> If candidate is selected, submit this candidate and afterwards submit the empty string, to terminate CRM. If no candidate is selected, submit empty string.

Unfortunately there is no general solution which works across all UIs. The double RET is okay in case one selects multiple candidates. This also allows double checking. However in case one wants to select a single candidate most of the time it might be a bit annoying. I think the UI is quite intuitive since it only requires RET - a user will probably understand that immediately, in contrast to the default CRM implementation, where the user must know about the crm-separator.

bdarcus:
> This is easy to solve by adding custom keybindings. For Vertico one could bind ...

Can you give me an example for Vertico please? What's the correct way to do this?

> TAB -> If candidate is selected, submit this candidate. If no candidate is selected, do nothing.

> Unfortunately there is no general solution which works across all UIs.

That's a minor cost though.

Is it feasible also to configure Vertico/Selectrum to use `shift-TAB` to do as we were earlier discussing: select, and keep the search string?

> The double RET is okay in case one selects multiple candidates. This also allows double checking. However in case one wants to select a single candidate most of the time it might be a bit annoying. I think the UI is quite intuitive since it only requires RET - a user will probably understand that immediately, in contrast to the default CRM implementation, where the user must know about the crm-separator. So overall I don't feel that strongly about this and I believe the benefits outweigh the disadvantages here.

You summarize the pros and cons well. Indeed, it is super simple and consistent.

It's just maybe too much so, given that you use `RET` to select both individual candidates, and the full selected set.

minad:
Okay, I got this to work with Embark. It requires a tiny bit of integration code in embark-consult.el though.

~~~ elisp
(add-hook 'embark-setup-hook #'embark-consult--crm-setup)
(defun embark-consult--crm-setup ()
  (when (eq minibuffer-history-variable 'consult--crm-history)
    (run-at-time 0 nil #'exit-minibuffer)))
~~~

minad:
> Is it feasible also to configure Vertico/Selectrum to use shift-TAB to do as we were earlier discussing: select, and keep the search string? I'm guessing no, but worth confirming.

Yes, this is possible but needs some effort since you have to reinject the last input string with the next selection. If we add such a `consult-crm-initial-input` variable, we can facilitate this use case.

> Can you give me an example for Vertico please? What's the correct way to do this?

I will try to implement these three things:

1. Command for selection only
2. Command for selection and exit
3. Command for selection preserving the input

minad:
This is the implementation. It works quite well but overall I am a bit unhappy with the way the implementation is split across vertico, embark and `consult-completing-read-multiple`. Now I am tempted to just package up all of this as a `vertico-crm` package instead. It would be even better than what we have here now. Only downside is that we would lose the consult portability benefit.

~~~ elisp
;; -*- lexical-binding: t -*-

(defun vertico-crm-select (&optional arg)
  "Select/deselect current candidate.
If prefix ARG is given, preserve minibuffer input."
  (interactive "P")
  (when (>= vertico--index 0)
    (when arg
      (let ((content (minibuffer-contents-no-properties)))
        (run-at-time 0 nil (lambda ()
                             (delete-minibuffer-contents)
                             (insert content)
                             (vertico--exhibit)))))
    (vertico-exit)))

(defun vertico-crm-done ()
  "Finish with currently selected candidate.
If no candidate is selected, exit directly."
  (interactive)
  (when (>= vertico--index 0)
    (run-at-time 0 nil #'exit-minibuffer))
  (vertico-exit))

(defun vertico-crm-setup (&rest args)
  "Setup keymap for `completing-read-multiple'."
  (let* ((depth (1+ (recursion-depth)))
         (hook (lambda ()
                 (when (= (recursion-depth) depth)
                   (let ((map (make-composed-keymap nil (current-local-map))))
                     (define-key map "\t" #'vertico-crm-select)
                     (define-key map "\r" #'vertico-crm-done)
                     (use-local-map map))))))
    (unwind-protect
        (progn
          (add-hook 'minibuffer-setup-hook hook)
          (apply args))
      (remove-hook 'minibuffer-setup-hook hook))))

(advice-add #'consult-completing-read-multiple :around #'vertico-crm-setup)
~~~

bdarcus:
> This is the implementation. It works quite well but overall 

Do I need some additional code to get the bindings?

> I am a bit unhappy with the way the implementation is split across vertico, embark and `consult-completing-read-multiple`. 

I can see that, but am agnostic myself.

> Now I am tempted to just package up all of this as a `vertico-crm` package instead. It would be even better than what we have here now. Only downside is that we would lose the consult portability benefit.

I guess a question would be whether an equivalent to such code might be appropriate for Selectrum?

If you did do a `vertico-crm` instead, for example, it would be a very small package, easy to adapt by others to other contexts.

Or, couldn't this still work in Selectrum or icomplete-vertical as is now, just without the additional commands?

``` elisp
(advice-add #'completing-read-multiple :override #'vertico-crm-completing-read-multiple)
```

minad:
I am closing this in favor of https://github.com/minad/vertico/pull/74.

oantolin:
Ah, you beat me to suggesting running `exit-minibuffer` an extra time in the setup hook.

I'm a little sad to see this closed. I guess users of default completion (like me) can just put this function in their personal configuration.

bdarcus:
> EDIT: Ignore that last remark, I somehow failed to notice the PR was reopened.

It's easy to miss @oantolin - @minad's been pretty active with these experiments the past few days, and has gone back and forth on what changes should live where :-)

minad:
@oantolin

I decided to merge the general consult version. I even enhanced it over the original version - I am now catching the `exit-minibuffer` to perform the selection instead of using a loop. We will see if this makes problems :D

@bdarcus

> I forget: is it still possible to do the same keybindings as in vertico-crm? To me, that is the key usability enhancement compared to this.

It is possible to add keybindings. I will probably add this in the consult-vertico/consult-selectrum integration code in the next step. But for now I wanted to get the basic version merged.

bdarcus:
Thanks for all your work on this @minad!

minad:
Thanks for your help and pushing this :+1:

I am looking forward to some feedback. I hope the current version works as expected, except for the improved keybindings.

PR #367, opened 2021-07-26
Add consult-line-everywhere

minad:
Search in lines across project buffers or all buffers. The name is up to debate.

cc @iyefrat, if you have time to give this a test I would appreciate it.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -4,9 +4,10 @@
 
 * Development
 
-- =consult-mark=, =consult-global-mark=: Added optional marker list argument
+- =consult-mark=, =consult-global-mark=: Add optional marker list argument
 - =consult-completing-read-multiple=: New function
-- Renamed =consult-project-imenu= to =consult-imenu-project=
+- Rename =consult-project-imenu= to =consult-imenu-project=
+- Add =consult-line-multi= to search multiple buffers
 
 * Version 0.9 (2021-06-22)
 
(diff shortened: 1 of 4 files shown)


hmelman:
I don't have an opinion on the name but for info:

- At least in what I have loaded the only "everywhere" I have is `ido-everywhere`.  
- I have 3 `-all-buffers` commands, for wgrep, tramp and Buffer-menu.  
- Another option is using `multi-` as in `multi-isearch-buffers` and  `multi-occur` 
- Since this is mostly about searching in projects, why not `consult-line-in-project`?

minad:
If a prefix argument is given, the search includes all buffers, not only the project buffers.

hmelman:
> If a prefix argument is given, the search includes all buffers, not only the project buffers.

I saw that, but I think you name a command based on it's default behavior, not based on what happens with an optional argument.  Obviously it's better if you can account for both.

Since it only searches open buffers, `consult-line-in-project` is probably wrong since it implies to me "all project files".  Since this searches several buffers, either those in the same project or all buffers, it seems closest to the "multi" commands.

Would you consider including other groupings of buffers, like all with the same major mode?  I could see that being useful in say all elisp buffers even if they're in different projects.

minad:
> Would you consider including other groupings of buffers, like all with the same major mode? I could see that being useful in say all elisp buffers even if they're in different projects.

Not out of the box, but if you look at the implementation - it uses the `consult--buffer-query` function, which allows to search buffers by multiple criteria. Users can easily write their own command based on these functions.

PR #371, opened 2021-07-28
Add async status action and decouple indicator

### LINKED ISSUE #370: Async: Decouple the process indicator from the process source
Add a status action to the async functions. When an async function is called with a list of the form `(status :key value...)`, it should append its status information to the plist and return the updated status information. This way status information can propagate up and down through an async pipeline.

* The process indicator can be separated from the process source.
* The throttler could benefit from the process runtime information and adapt its delays accordingly.
* Non-process backends can reuse the indicator overlay, e.g., a web backend.

cc @tecosaur

tecosaur:
Here's a start. There is one unresolved issues I'm aware of

+ without the `(consult--async-indicator)` the status can be passed through as a list of candidates

This should close #370.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -1427,7 +1427,8 @@ SPLIT is the splitting function."
 ASYNC is the async function which receives the candidates.
 CMD is the command argument list.
 PROPS are optional properties passed to `make-process'."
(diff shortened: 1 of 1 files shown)


minad:
Putting this on hold for now as discussed.

PR #374, opened 2021-07-30
feat : add source `consult--source-dired'

liuyinz:
I write a dired source `consult--source-dired` and filter `consult--source-buffer` with today's `:predicate` update.

Any advice for hidden and order with dired source?


minad:
Yes, this is a useful buffer source! However I don't want to add further sources to the Consult package. You can document the source in the [Consult wiki](https://github.com/minad/consult/wiki). It is generally encouraged to write custom sources specific to your preferences and requirements and add them to your init.el.

Issue #317, opened 2021-05-26
Windows: Process arguments must be manually escaped and encoded


oantolin:
I managed to get `consult-find` to work in Windows, but oddly this required to escape the asterisks with backslashes:

```find . -not ( -wholename \*/.\* -prune ) -ipath \*ARG\* OPTS```

With the backslashes it does _not_ work on Linux. I don't know if you want an OS dependent default value for the variable, to mention the backslashes for Windows users in the manual, or just to have people find this issue here.

By the way, the hidden ` *consult-async*` buffer was invaluable in figuring this out, and I wish affe had something like it.

(Another pitfall on Windows is that there is a `find.exe` in `C:\WINDOWS\System32` that you want to avoid, so you need to make sure that GNU find is earlier on the path, but that pitfall I was aware of!)

minad:
Uh why is that? On which level does windows interpret the wildcards? The escaping is not necessary on Unix since the arguments are passed directly to the process without going through the shell.

I would like to avoid platform dependent defaults. I prefer if the issue could be fixed by adding some additional escaping where necessary automatically. Any idea how one could avoid the find.exe issue?

oantolin:
> Uh why is that? On which level does windows interpret the wildcards?

I have no idea, it makes no sense to me. Maybe on Windows `make-process` shells out because they didn't find an appropriate Windows API function? I don't know.



> I would like to avoid platform dependent defaults. I prefer if the issue could be fixed by adding some additional escaping where necessary automatically.



> Any idea how one could avoid the find.exe issue?

Well, there is a `find-program` variable in `grep.el`, which you can point to the location of your find executable. It's default value is just `"find"`, of course. I imagine Windows users that do install GNU find typically set that variable to point to it, rather than mess with paths? I've decided to try the drastic approach of putting Git's `bin/` ahead of `C:\Windows\System32` on the path, but I guess that might cause problems (so far it hasn't).

vincentjgoh:
I mentioned this on reddit: consult-locate apparently has trouble with passing the arguments unadulterated to everything search under windows.

It's worth noting that helm isn't having this same problem; I can get the same results there as on the command line. 

Helm's command string looks like this:
`es %s -instance locate-instance -s %s`
vs consult's:
`es OPTS -instance locate-instance -s ARG`

Helm's docs say this:

[...]

PR #380, opened 2021-08-04
Transform input of consult-grep and consult-find commands

minad:
The input string is transformed before passing it to the backend. Spaces are
replaced by `.*`. Furthermore the input is interpreted as Emacs regular
expression and transformed to a Perl-compatible regular expression (PCRE), by a
handful of basic substitutions. The matching substrings are not obtained anymore
from the backend by looking for ANSI escape sequences. Instead the input regular
expression is used to heuristically apply highlighting.

Advantages:

- No parsing of ANSI escape sequences, which is fragile and inefficient.
- More flexibility, input string can be transformed, e.g., "def consult" -> "def.*consult".
- Highlighting becomes possible for backends which don't return highlights, e.g., find.

Disadvantages:

- Highlighting may be incorrect. In practice this is rare, since most search inputs
  are simple regular expressions.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -10,6 +10,18 @@
 - Add =consult-line-multi= to search multiple buffers
 - Removed obsolete =consult-yank=, =consult-async-default-split=, =consult-config=
 - =consult-ripgrep=: Use =--smart-case=
+- =consult-grep/git-grep=: Use =--ignore-case=
+- Deprecate =consult-<cmd>-command= in favor of =consult-<cmd>-config.=
+- =consult-grep/git-grep/ripgrep/find/locate=: Add support for multiple patterns.
+  + =consult-git-grep= and =consult-find= support unordered patterns out of the box.
+    For example: "first second third" is transformed to "first -and second -and third".
+  + =consult-grep= and =consult-ripgrep= support unordered patterns if the program supports it.
+  + =consult-locate= only support ordered patterns.
+    For example: "first second third" is transformed to "first.*second.*third".
+- =consult-find/locate/man=: Add highlighting.
+- =consult-grep/git-grep/ripgrep=: Compute the highlighting based on the input,
+  instead of relying on the ANSI-escaped output. This works better with multiple
+  patterns, but may occasionally produce false highlighting.
 
 * Version 0.9 (2021-06-22)
 
--- a/README.org
+++ b/README.org
@@ -250,22 +250,19 @@ their descriptions.
    search gets started. Consult splits the input string into two parts, if the
    first character is a punctuation character, like =#=. For example
    =#grep-regexp#filter-string=, is split at the second =#=. The string =grep-regexp=
-   is passed to Grep, the =filter-string= is passed to the /fast/ Emacs filtering to
-   further narrow down the list of matches. This is particularly useful if you
-   are using an advanced completion style like orderless. =consult-grep= supports
-   preview. If the =consult-project-root-function= is [[#use-package-example][configured]] and returns
-   non-nil, =consult-grep= searches the current project directory. Otherwise the
-   =default-directory= is searched. If =consult-grep= is invoked with prefix
-   argument =C-u M-s g=, you can specify the directory manually.
+   is passed to Grep. By default, spaces are replaced by ~.*~. The =filter-string=
+   is passed to the /fast/ Emacs filtering to further narrow down the list of
+   matches. This is particularly useful if you are using an advanced completion
+   style like orderless. =consult-grep= supports preview. If the
+   =consult-project-root-function= is [[#use-package-example][configured]] and returns non-nil, =consult-grep=
+   searches the current project directory. Otherwise the =default-directory= is
+   searched. If =consult-grep= is invoked with prefix argument =C-u M-s g=, you can
+   specify the directory manually.
  - =consult-find=, =consult-locate=: Find file by
    matching the path against a regexp. Like for =consult-grep,= either the project
    root or the current directory is the root directory for the search. The input
    string is treated similarly to =consult-grep=, where the first part is passed
-   to find, and the second part is used for Emacs filtering. Note that the
-   standard =find= command uses wildcards in contrast to the popular =fd=, which
-   uses regular expressions. In case you want to use =fd=, you can either change
-   the =consult-find-command= configuration variable or define a small command as
-   described in the [[https://github.com/minad/consult/wiki][Consult wiki]].
+   to find, and the second part is used for Emacs filtering.
 
 ** Compilation
    :properties:
@@ -534,6 +531,9 @@ their descriptions.
  Examples:
 
  - =#defun=: Search for "defun" using grep.
+ - =#consult embark=: Search for "consult.*embark" using grep.
+   If the grep program has support for Perl-compatible regular expressions
+   (PCRE), the search includes both "consult.*embark" and "embark.*consult".
  - =#defun#consult=: Search for "defun" using grep, filter with the word
    "consult".
  - =/defun/consult=: It is also possible to use other punctuation
--- a/consult.el
+++ b/consult.el
@@ -221,47 +221,100 @@ See `consult--multi' for a description of the source values."
   "Maximal number of columns of grep output."
   :type 'integer)
 
-(defcustom consult-grep-command
-  "grep --null --line-buffered --color=always --extended-regexp\
-   --exclude-dir=.git --line-number -I -r . -e ARG OPTS"
-  "Command line string for grep, see `consult-grep'.
+(defconst consult--grep-match-regexp
+  "\\`\\(?:./\\)?\\([^\n:]+\\):\\([0-9]+\\):"
+  "Regexp used to match file and line of grep output.")
 
-The command string must have a specific format, including ARG and OPTS
-substrings. ARG is replaced by the filter string and OPTS by the auxillary
-command options."
-  :type 'string)
+(defcustom consult-grep-command
+  (list :args "grep --line-buffered --color=never --ignore-case\
+               --exclude-dir=.git --line-number -I -r ."
+        :command (let (type)
+                   (lambda (config input)
+                     (setq type (or type (consult--grep-supported-regexp "grep"))
+                           input (consult--command-split input))
+                     (append (split-string-and-unquote (plist-get config :args))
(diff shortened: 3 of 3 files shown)


oantolin:
It's tempting to reuse `orderless-pattern-compiler`, translate each regexp to PCRE, and join with `" -and "`.

minad:
Yes, it is! I also had that planned originally but then somehow moved away from it because of the more complicated translations. If you look into the code I have some basic infrastructure there (very basic!). Maybe I should cut this down even more and focus on the Orderless integration. Probably one needs pcre2el to make this robust.

oantolin:
Does pcre2el also have el2pcre functions?

minad:
Yes, both directions. This is quite nice since then the users can configure the style as they like :)

minad:
Maybe a simpler converter would suffice for Orderless as long as you don't write complicated custom regexps yourself?

oantolin:
> Maybe a simpler converter would suffice for Orderless as long as you don't write complicated custom regexps yourself?

I think that's right. The regexp features the mathcing styles that come with orderless are not that fancy. Grouping without capturing, word boundaries, maybe a little more.

minad:
I refactored everything a bit, which should allow us to integrate orderless. Now we have these regexp functions:

* `consult--emacs-to-extended-regexp`: Very basic converter
* `consult--compile-regexp`: Compile string to list of regexps of type basic, extended, emacs or lookahead
* `consult--join-regexp`: Compile string to a single joined regexp of type basic, extended, emacs or lookahead

https://github.com/minad/consult/blob/2539fd71e65d52db4e5854acb4bc367818e3680d/consult.el#L582-L609

In order to support Orderless it is only necessary to replace `consult--compile-regexp`. If a basic regexp is requested, Orderless should bail out, but for the other types Orderless can be used. Furthermore `consult--emacs-to-extended-regexp` must be extended such that it translates the regexp features used by Orderless.

grolongo:
Just to make sure I understand this, is the intended way to configure grep/ripgrep/find through `consult-customize` as we already do for consult-buffer completions?

Example:
```elisp
(consult-customize
 consult-ripgrep-config
 :args "rg --line-buffered --color=never --max-columns=1000 --no-heading\
 --line-number --follow --hidden --search-zip --smart-case .")
```

`consult-find` doesn't work out of the box on windows-nt.
`:args "find . -not ( -wholename \\*/.\\* -prune )"` fixes it, though as an UX suggestion I think it would be better for windows-nt users to not have to look for a solution and make it just work for them too. Maybe add an os detection to tweak the command or just don't include to ignore dotfiles by default?

Last question, would you think it would be a good idea to make `consult-grep` to also rely on `grep-find-ignored-files` and `grep-find-ignored-directories`?

minad:
> Just to make sure I understand this, is the intended way to configure consult-grep/ripgrep/find through consult-customize as we already do for consult-buffer completions?

No you configure the commands by modifying the `consult-ripgrep-config` variable directly. `consult-customize` is only for command and buffer configurations, it is a feature of the underlying `consult--read` API. This is unfortunately a bit misleading, but `consult-customize` is a special case.

> consult-find doesn't work out of the box on windows-nt.

Yes, this is a known issue, see #317. Unfortunately I don't have the possibility to test or fix this. But if someone implements a proper fix I would appreciate it. Furthermore I suggest to report this as an upstream bug to emacs-devel. This is a bug in the process API, the API abstraction is broken.

> Last question, would you think it would be a good idea to make consult-grep to also rely on grep-find-ignored-files and grep-find-ignored-directories?

It doesn't sound like a bad idea at first. I also considered this. Downside is that we have to require the grep library then, which is not good for lazy loading. The `consult-grep` etc commands are completely decoupled from the Emacs grep library, it does not make use of any of its infrastructure. `embark-export` is the glue which allows to transfer from `consult-grep` to a `grep-mode` buffer.

oantolin:
While `consult-find` doesn't work out of the box on Windows, I  had found a workaround that's documented in that issue that @minad linked, #317. I guess that workaround might need updating for the recent changed in consult.

minad:
@oantolin Do you have an idea how the issue could be fixed properly at the core? It doesn't make sense to me why this problem exists at all.

grolongo:
> No you configure the commands by modifying the consult-ripgrep-config variable directly. consult-customize is only for command and buffer configurations, it is a feature of the underlying consult--read API. This is unfortunately a bit misleading, but consult-customize is a special case.

Understood, therefore `plist-put` is the way to go unless you use the customize interface.

> Downside is that we have to require the grep library then, which is not good for lazy loading. The consult-grep etc commands are completely decoupled from the Emacs grep library, it does not make use of any of its infrastructure.

After thinking about it I agree with you it's better to keep it separated. Also those options are made to be used with `find` coupled with `grep` after all, not just `grep` alone.

minad:
> Understood, therefore plist-put is the way to go unless you use the customize interface.

Yes, but I am currently overhauling this again. I am not yet satisfied with the current design and it is still lacking flexibility. The goal is to have the ability to plug in orderless as a pattern transformer, such that you can search via ripgrep the usual orderless patterns.

rswgnu:
Could you add a comment here on whether/how you have added orderless support for grep/ripgrep commands.  Please give a complete example as I find the documentation on how to send arguments to the grep commands very hard to follow.  For example, give an example where a multi-word pattern is searched regardless of order but only within a single, specific file.  Thanks.  (Just starting with consult and the grep functionality is really great.  Would like to be able to search over either multiple directories or within a specific list of files too).

minad:
@rswgnu It is in the wiki: https://github.com/minad/consult/wiki#use-orderless-as-pattern-compiler-for-consult-grepripgrepfind

rswgnu:
This still does not supply a precise example of the syntax to specify for the initial input args.  Is that fully documented anywhere with specific examples?  It would really help if you just gave one detailed example wherever you mention complex syntax and then explain that example rather than just listing it.  For example, if I say use this:  "#expr -- -g *.c#filter", no one will no what that means.  But if you break it down and explain each part, then it will be clear.  Thanks.

rswgnu:
To make it even simpler, I know how to write a grep command like:   grep -n my-expr file1 file2.  What is the exact expression to send those args to consult-grep and consult-ripgrep (or their associated commands)?

rswgnu:
> @rswgnu It is in the wiki: https://github.com/minad/consult/wiki#use-orderless-as-pattern-compiler-for-consult-grepripgrepfind

I tried that and it errored out since 'input' is not known within the lambda expression (using latest consult and orderless packages).  This sort of runs but only matches to the first character typed and only after removing the leading # async symbol:

```elisp
(defun consult--orderless-regexp-compiler (input type &rest _config)
  (setq input (orderless-pattern-compiler input))
  (cons
   (mapcar (lambda (r) (consult--convert-regexp r type)) input)
   `(lambda (str) (orderless--highlight ',input str))))

;; OPTION 1: Activate globally for all consult-grep/ripgrep/find/...
(setq consult--regexp-compiler #'consult--orderless-regexp-compiler)
```

minad:
@rswgnu I think you want to do something which is not directly supported by consult-grep (but you can of course configure and adjust it). You probably want to adjust everything to your Hyperbole ways but that's not what most users necessarily want. ;)

I suggest you try Consult as is for a while. Then you will see how everything fits together. Also you should read the docs completely. Most of the information you asked for should be there. Since you are an expert you could also skim the wiki and the source for additional tweaks.

rswgnu:
Hi Daniel:

I have read through a great deal of the documentation and have not been
able to solve this.  It is not a Hyperbole-specific thing.  I have seen
many requests for how to specify a set of files (with full or relative
paths) for consult-grep type commands to utilize and none of the proposed
solutions work with the current code.

If consult supports such a command, please show the needed inputs.  If not,
I would suggest there is definitely demand.  It seems like such a basic use
case.  Typical examples are grepping over files that only match a glob
pattern, like *foo*.el or recursively grepping over multiple directories
potentially with a file glob as well.

In the meantime, I am utilizing consult's default behavior but this is
limiting my use.

minad:
On 2/20/23 08:17, Robert Weiner wrote:
> I have seen many requests for how to specify a set of files (with full or relative
> paths) for consult-grep type commands to utilize and none of the proposed
> solutions work with the current code.

Then could try if you can adapt any of the existing dysfunctional
solutions, such that it works again.

> If not, I would suggest there is definitely demand. It seems like such a basic use
> case. Typical examples are grepping over files that only match a glob
> pattern, like *foo*.el or recursively grepping over multiple directories
> potentially with a file glob as well.

Sure there is always need for more commands. But we still have to draw a
line somewhere and rather provide building blocks which can then be
adapted instead of adding more and more specialized commands. How should
such the file search command look like exactly? Can you provide a
prototype? How would the workflow look like? First select some files,
then grep or first provide a wildcard pattern, then grep?

rswgnu:
On Mon, Feb 20, 2023 at 10:35 AM Daniel Mendler ***@***.***>
wrote:

> On 2/20/23 08:17, Robert Weiner wrote:
> > I have seen many requests for how to specify a set of files (with full
> or relative
> > paths) for consult-grep type commands to utilize and none of the proposed
> > solutions work with the current code.
>


> How would the workflow look like? First select some files,
> then grep or first provide a wildcard pattern, then grep?
>

The grep commands would take an extra argument that could be a list of
paths with glob wildcards.  The Emacs wildcard expander is first used to
expand the wildcards.  Then for each path that is a directory, the grep is
done like you do it now with a single directory.  For each path that is a
file, it is added to the end of the grep command and simply grepped over.
If you make such updates, then we can be assured that it will be a part of
consult going forward.

Makes sense?

minad:
On 2/20/23 16:49, Robert Weiner wrote:
> The grep commands would take an extra argument that could be a list of
> paths with glob wildcards. The Emacs wildcard expander is first used to
> expand the wildcards. Then for each path that is a directory, the grep is
> done like you do it now with a single directory. For each path that is a
> file, it is added to the end of the grep command and simply grepped over.
> If you make such updates, then we can be assured that it will be a part of
> consult going forward.

Makes sense. Not sure how difficult it is to implement it. But if you
want to work on it, go ahead. I don't see myself using this since I am
happy with the existing Consult commands. I like that I don't have to
select multiple directories. I mostly use consult-ripgrep to search
recursively through projects.

rswgnu:
Hi Daniel:

I'll work on it when I get time on the weekend and will submit a PR when
ready.

PR #386, opened 2021-08-07
Fix `consult--command-split`

mohamed-abdelnour:
Hello, @minad!

This is a small fix to make `consult--command-split` remove the `--` from the arguments when splitting an input.

#### Before: `--` is included in the arguments

```elisp
~ $ (consult--command-split "case -- --hidden --glob *backup")
("case -- " "--hidden" "--glob" "*backup")
```

`*consult-async*`: `--` is included as a string to match, `(?=.*--)`:

```txt
consult--async-process started ("rg" "--line-buffered" "--color=never" "--max-columns=1000" "--path-separator" "/" "--smart-case" "--no-heading" "--line-number" "." "-P" "-e" "^(?=.*case)(?=.*--)" "--hidden" "--glob" "*backup")
```

#### After: `--` is removed

```elisp
~ $ (consult--command-split "case -- --hidden --glob *backup")
("case" "--hidden" "--glob" "*backup")
```

`*consult-async*`:

```txt
consult--async-process started ("rg" "--line-buffered" "--color=never" "--max-columns=1000" "--path-separator" "/" "--smart-case" "--no-heading" "--line-number" "." "-P" "-e" "case" "--hidden" "--glob" "*backup")
```

On a different note, I quickly benchmarked ripgrep with [hyperfine](https://github.com/sharkdp/hyperfine). Running ripgrep with the PCRE2 engine seems to be generally slower than running it with the Rust engine. Do you think `consult-ripgrep` should default to using the PCRE2 engine, or shall we look into ways of using the Rust engine whenever possible?

An easy step would be to use ripgrep's `--auto-hybrid-regex` flag; this uses the PCRE2 engine only if needed. That would still use the PCRE2 engine a lot of the time (because Consult uses positive lookahead to match multiple strings), so we'd have to change that if we want to rely on the Rust engine.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -536,7 +536,7 @@ ARGS is a list of commands or sources followed by the list of keyword-value pair
   (save-match-data
     (let ((opts (when (string-match " +--\\( +\\|\\'\\)" str)
                   (prog1 (substring str (match-end 0))
-                    (setq str (substring str 0 (match-end 0)))))))
+                    (setq str (substring str 0 (match-beginning 0)))))))
       (unless (string-blank-p str)
         ;; split-string-and-unquote fails if the quotes are invalid. Ignore it.
         (cons str (and opts (ignore-errors (split-string-and-unquote opts))))))))


minad:
> This is a small fix to make consult--command-split remove the -- from the arguments when splitting an input.

Oh good, I probably messed this up in my recent (large) refactorings. If you find other issues please let me know!

minad:
Regarding PCRE - there is no other way than using it if we want positive lookahead. In case you want to disable PCRE, just set `consult--ripgrep-regexp-type=extended`.

minad:
I looked into `--auto-hybrid-regexp`. Unfortunately it is not available for the (probably outdated) ripgrep version I have installed, so we won't add this. But you can add it to your user configuration of course.

mohamed-abdelnour:
> Oh good, I probably messed this up in my recent (large) refactorings. If you find other issues please let me know!

I've seen how much refactoring you've been doing, keep up the excellent work!

mohamed-abdelnour:
> I looked into `--auto-hybrid-regexp`. Unfortunately it is not available for the (probably outdated) ripgrep version I have installed, so we won't add this. But you can add it to your user configuration of course.

Good to know, I'll add it to my config for now, then. Maybe we'll look into it again in the future! Quick note: `--auto-hybrid-regex` is deprecated; the way to do it now is: `--engine auto`.

minad:
> Quick note: --auto-hybrid-regex is deprecated; the way to do it now is: --engine auto.

Good to know. I also don't have the flag unfortunately.

mohamed-abdelnour:
> Good to know. I also don't have the flag unfortunately.

We'll revisit it later, then.

For the sake of completeness:
- `--auto-hybrid-regex` was added in ripgrep [11.0.0](https://github.com/BurntSushi/ripgrep/blob/master/CHANGELOG.md#1100-2019-04-15),
- and it was deprecated in favour of `--engine auto` in ripgrep [12.0.0](https://github.com/BurntSushi/ripgrep/blob/master/CHANGELOG.md#1200-2020-03-15).

We can consider adding it when ripgrep 11.0 is the standard, aka when it hits Debian Stable. ;)

minad:
> We can consider adding it when ripgrep 11.0 is the standard, aka when it hits Debian Stable. ;)

That's a good standard generally. I went through many distributions over the years but mostly settled on Debian stable for the time being. But in the future I may consider nix/guix :)

mohamed-abdelnour:
Oh, you're on Debian—that's awesome! I'm perfectly happy with the distribution I'm on, and I have no reason to switch, but I constantly contemplate switching to Debian or NixOS.

Anyway, this accidentaly turned into a bit of a chat, and I don't want to waste any more of your time, so I'll leave you to it! :)

PR #391, opened 2021-08-09
Query prompt for buffer selection

jdtsmith:
This gives more feedback on the filtering and number of buffers being search for `consult-line-multi`.

### DIFF
--- a/consult-imenu.el
+++ b/consult-imenu.el
@@ -147,11 +147,9 @@ TYPES is the mode-specific types configuration."
                 (buffer-name) (error-message-string err))
        nil)))
 
-(defun consult-imenu--multi-items (query)
-  "Return all imenu items from buffers matching QUERY."
-  (apply #'append (consult--buffer-map
-                   (apply #'consult--buffer-query query)
-                   #'consult-imenu--items-safe)))
+(defun consult-imenu--multi-items (buffers)
+  "Return all imenu items from BUFFERS."
+  (apply #'append (consult--buffer-map buffers #'consult-imenu--items-safe)))
 
 (defun consult-imenu--jump (item)
   "Jump to imenu ITEM via `consult--jump'.
@@ -222,17 +220,14 @@ In order to determine the buffers belonging to the same project, the
 same major mode as the current buffer are used. See also
 `consult-imenu' for more details. In order to search a subset of filters,
 QUERY can be set to a plist according to `consult--buffer-query'."
-  (interactive)
-  (let ((scope "Multiple buffers"))
-    (when-let (project (and (not (keywordp (car-safe query)))
-                            (consult--project-root)))
-      (setq scope (format "Project %s" (consult--project-name project))
-            query `(:directory ,project :mode ,major-mode :sort alpha)))
-    (if query
-        (consult-imenu--select
-         (format "Go to item (%s): " scope)
-         (consult-imenu--multi-items query))
-      (consult-imenu))))
+  (interactive "P")
+  (unless (keywordp (car-safe query))
+    (setq query (list :sort 'alpha :mode major-mode
+                      :directory (and (not query) 'project))))
+  (let ((result (apply #'consult--buffer-query-scope query)))
+    (consult-imenu--select
+     (format "Go to item (%s): " (cdr result))
+     (consult-imenu--multi-items (car result)))))
 
 (define-obsolete-function-alias
   'consult-project-imenu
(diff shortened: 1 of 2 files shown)


minad:
Thanks! I played around with this and tried to make the behavior more uniform. Please take a look. The main problem is that the prompt gets a bit long.

jdtsmith:
Looks good.  Good idea to get buffers and scope message all in one place to avoid all the plist overhead.  The other thought I had to cut down on length of the prompt, that would actually add more value to the search overall: move the number of buffers, as `num_shown/num_searched`, to its own auto-updating indicator at the front of the string, like `1/4451 [9/9] Go to line (Project consult): `.  Then as the searching whittles down the number of actual matching buffers shown, it might drop to, say, `[2/9]`.  This only shortens for certain cases, but I think it's the problematic ones.

jdtsmith:
One other tweak.  If I'm searching for just `:mode` matching buffers, or trimming the list with `:include`/`:exclude`, `All buffers` doesn't feel appropriate.  I'd reserve the word `All` for truly _all_ buffers (modulo `consult-buffer-filter`).  This is a shortening change.

jdtsmith:
We could place an absolute limit on the length of the directory name, and drop down to showing only the final directory path unless below some limit.  Project name could also be shortened similarly.

jdtsmith:
And this one is more radical: it's a little jarring that the position of the "Go to line (blah blah): " cursor prompt moves around so much as you switch scope parameters.  You could _right-align_ the variable length scope info (All buffers, Project foo, etc.).  And just let the user-entered search stuff "push it off the right side" if they type too much.  Then the "Go to line: " stays ~in the same place within the minibuffer.

jdtsmith:
And then, once the scope prompt text/position is sorted, we're left with a multi-parameter command with a combinatorial space:

- `:directory`: project, path, `nil`
- `:sort`: `'visibility`, `'alpha`, `nil`
- `:filter`: `t`, `nil`, `'inverted`
- `:exclude`: list of regexps, nil
- `:include`: list of regexps, nil
- `:mode`: a (list of) modes, nil
- `:predicate`: function or nil
- `:as`: function or nil

I.e. **864** different possible combinations! (Or 216 if you don't expose the final 2.)  Only so much one can do with prefix key binding to span this input parameter space.  

A transient would be an _ideal_ interface.  Default buffers if called normally, pop the transient if called with a prefix.  They are somewhat hard to construct though.

minad:
> Then as the searching whittles down the number of actual matching buffers shown, it might drop to, say, [2/9].

This is impossible/unrealistic to implement with reasonable effort due to the interaction between UI and Consult.

> We could place an absolute limit on the length of the directory name, and drop down to showing only the final directory path unless below some limit. 

Agree. But truncation of the project name is unnecessary.

 > Then the "Go to line: " stays ~in the same place within the minibuffer. This one would require an addition to consult--read to support a :right-prompt or so.

The idea is not bad, but it is a bit of a mess to get this working with different UIs. Does not seem worth the effort.

> A transient would be an ideal interface. Default buffers if called normally, pop the transient if called with a prefix. They are somewhat hard to construct though.

The transient idea is not bad of course. But this is something which should be implemented *on top* of consult-line-multi. It should be easy to write a consult-line-transient. I would not want to bake transient right into consult-line-multi. Transient is even part of Emacs core now.

jdtsmith:
> The transient idea is not bad of course. But this is something which should be implemented _on top_ of consult-line-multi. It should be easy to write a consult-line-transient. I would not want to bake transient right into consult-line-multi. Transient is even part of Emacs core now.

The main issue is the transient documentation is quite... dense.  So non-trivial to implement.  I like the idea of a small package to do this.  You could envision many, such as ibuffer-transient.

minad:
What do you think about https://github.com/minad/consult/pull/391/commits/a91a21369df481ade3b7015366d7832e43e06232? I like this option the best I think.

minad:
For now I merged a reduced version https://github.com/minad/consult/commit/3b3e62b0fe84d195e8c7fd9f034691a20d3f476c.

jdtsmith:
Sorry I didn't see your final commit.  I agree keeping it to one line is preferable.  I do miss the `:mode` info.  Since you can afford the space to say "Project consult", I don't think "<emacs-lisp-mode>" is problematic.  Could also shorten that to remove any final "-mode".

minad:
> Since you can afford the space to say "Project consult", I don't think emacs-lisp-mode is problematic.

Only because I can afford "Project consult", this does not mean that I can have "emacs-lisp" too or even both project and mode. I tried it for a while and it is too verbose, of course I tried with the `-mode` stripped off. I don't miss the mode since the mode is quite obvious from the content. I want to keep the more minimal prompt for now and put my hopes in a potential consult-transient interface for some of the consult commands (consult-grep family of async commands and the commands which take a buffer query argument).

PR #402, opened 2021-08-16
consult-xref--preview: Handle xref-elisp-location

### LINKED ISSUE #400: consult--preview-post-command: wrong-type-argument
Hi, I'm noticing some errors using consult with xref. This is with `emacs -Q` (emacs 28):

```elisp
(add-to-list 'load-path "/path/to/consult")
(require 'consult-xref)
(setq xref-show-xrefs-function #'consult-xref
      xref-show-definitions-function #'consult-xref)

(xref-find-definitions "ediff") ; works
(require 'ediff)
(xref-find-definitions "ediff") ; more candidates - now it errors
```

```
Error in post-command-hook (consult--preview-post-command): (wrong-type-argument integer-or-marker-p "No preview for xref-elisp-location")
```

fbergroth:
Fixes #400. But the `t` clause will still likely error, so perhaps that should be adjusted as well?

### DIFF
--- a/consult-xref.el
+++ b/consult-xref.el
@@ -60,6 +60,8 @@
                    (cl-typecase loc
                      (xref-buffer-location
                       (xref-location-marker loc))
+                     (xref-elisp-location
+                      (xref-location-marker loc))
                      (xref-file-location
                       (consult--position-marker
                        (funcall open (oref loc file))


minad:
The problem with this patch is that it will open new files which will then be kept open. Therefore I cannot accept the patch in this form. We need the same special handling for `xref-elisp-location` like we have it for `xref-file-location`.

fbergroth:
> The problem with this patch is that it will open new files which will then be kept open even after preview. Therefore I cannot accept the patch in this form.

Right. Thanks for the fix you committed.

> We need the same special handling for `xref-elisp-location` like we have it for `xref-file-location`.

It gets a bit hairy, and hard to do without copying logic from the `xref-location-marker` implementation.

Rather than the `consult--temporary-files` machinery, can't we just compare `(buffer-list)` before and after `(xref-location-marker loc)`, and record any buffers that were opened? Those buffers would then be closed when restoring. Then no special handling would be necessary. But perhaps you already thought of that, and saw some problems?

Alternatively, just check if `loc` has a `file` slot, and if there is no buffer visiting it, we close any buffer visiting the file afterwards?

minad:
> It gets a bit hairy, and hard to do without copying logic from the xref-location-marker implementation.

Yes, it is okay to just copy this logic in an adjusted form.

> Rather than the consult--temporary-files machinery, can't we just compare (buffer-list) before and after (xref-location-marker loc), and record any buffers that were opened? Those buffers would then be closed when restoring. Then no special handling would be necessary. But perhaps you already thought of that, and saw some problems?

Of course I thought of this. The downside is that then the preview will only be cleaned up in the very end. Maybe I am a bit overly cautious here, but the goal is that preview only opens up to consult-preview-max-count files.

fbergroth:
@minad I'm a bit hesitant to copy the logic.

What do you think of the following? The opener is used as a guard, and we assume that `xref-location-marker` won't open other files.
```patch
@@ -60,11 +60,12 @@
                    (cl-typecase loc
                      (xref-buffer-location
                       (xref-location-marker loc))
+                     (xref-elisp-location
+                      (and (funcall open (xref-elisp-location-file loc))
+                           (xref-location-marker loc)))
                      (xref-file-location
-                      (consult--position-marker
-                       (funcall open (oref loc file))
-                       (oref loc line)
-                       (oref loc column)))
+                      (and (funcall open (oref loc file))
+                           (xref-location-marker loc)))
                      (t (message "No preview for %s" (type-of loc)) nil))
                    nil)))))))
```

minad:
This could work if you can guarantee that `xref-location-marker` does indeed not open a file if the funcall open opened it first. Then we can add this.

PR #423, opened 2021-09-15
Call compilation-find-file from the correct directory

### LINKED ISSUE #301: consult-compile-error starts recursive minibuffers
`consult-compile-error` stopped working for me unless I enable recursive minibuffers. And even if I do, a query with prompt `Find this error in (default ...):` appears each time I highlight a different candidate.

Does that happen to you? If not I'll try to provide more specific details.

astoff:
This does the trick for me. We might also either enable recursive minibuffers, or just give up if `compilation-find-file` asks for a directory, but with an appropriate error message.

Fixes #301

### DIFF
--- a/consult-compile.el
+++ b/consult-compile.el
@@ -67,20 +67,21 @@
 
 (defun consult-compile--error-lookup (_ candidates cand)
   "Lookup marker of CAND by accessing CANDIDATES list."
-  (when-let (cand (car (member cand candidates)))
-    (let ((marker (get-text-property 0 'consult-compile--marker cand))
-          (loc (get-text-property 0 'consult-compile--loc cand)))
-      (when (buffer-live-p (marker-buffer marker))
-        (consult--position-marker
-         ;; taken from compile.el
-         (apply #'compilation-find-file
-                marker
-                (caar (compilation--loc->file-struct loc))
-                (cadar (compilation--loc->file-struct loc))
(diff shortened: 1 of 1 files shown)


minad:
Thank you! The buffer live check comes after the default-directory let binding. Can you fix this please?

astoff:
Ah, sure. Any ideas about popping a recursive minibuffer versus giving up?

astoff:
My choice would be to just not do any live preview if the file is not found (if you enable recursive minibuffers, you'll get that query anyway). But just wrapping everything in `ignore-errors` doesn't seems to work.

consult-compile.el
```diff
@@ -67,20 +67,21 @@
 
 (defun consult-compile--error-lookup (_ candidates cand)
   "Lookup marker of CAND by accessing CANDIDATES list."
-  (when-let (cand (car (member cand candidates)))
-    (let ((marker (get-text-property 0 'consult-compile--marker cand))
-          (loc (get-text-property 0 'consult-compile--loc cand)))
-      (when (buffer-live-p (marker-buffer marker))
```
minad, line 73: Oh, I meant to keep the buffer live check but move the buffer-local-value inside the check.

astoff: The nice thing is, `marker-buffer` already returns nil in that case.

minad: Perfect.

PR #435, opened 2021-10-05
Fix consult-xref preview on emacs master

northgaard:
The internals of xref were changed to use cl-defstruct instead of
EIEIO, which breaks the use of `oref' in consult-xref--preview. With
this commit an additional code path is introduced which handles the
cl-defstruct case.

See [this commit](https://github.com/emacs-mirror/emacs/commit/86da812afb2572c7fead2bb07570b976bffd7c55).

### DIFF
--- a/consult-xref.el
+++ b/consult-xref.el
@@ -61,10 +61,17 @@
                      (xref-buffer-location
                       (xref-location-marker loc))
                      (xref-file-location
-                      (consult--position-marker
-                       (funcall open (oref loc file))
-                       (oref loc line)
-                       (oref loc column)))
+                      ;; As of Emacs 28, xref uses cl-defstruct,
+                      ;; whereas earlier versions use EIEIO
+                      (if (cl-struct-p loc)
+                          (consult--position-marker
+                           (funcall open (xref-file-location-file loc))
+                           (xref-file-location-line loc)
+                           (xref-file-location-column loc))
+                        (consult--position-marker
+                         (funcall open (oref loc file))
+                         (oref loc line)
+                         (oref loc column))))
                      (t (message "No preview for %s" (type-of loc)) nil))
                    nil)))))))
 


minad:
> The internals of xref were changed to use cl-defstruct instead of
> EIEIO, which breaks the use of `oref' in consult-xref--preview. With
> this commit an additional code path is introduced which handles the
> cl-defstruct case.

Great, I like that xref changed from clos to structs! But I am not very
happy with the upgrade path here. An alternative could be to depend on
an explicit xref version. But Consult aims to work with Emacs out of the
box without additional dependencies, so it seems there is no other way
than the code you are proposing here. In any case we should add a
comment to the conditional, why the check is needed and from which
Emacs/Xref version on the check could be dropped in the future. Is the
Xref version already released which makes this switch?

northgaard:
This change to xref will be included in the soon (?) to be released Emacs 28. I've added a comment to reflect this.

Strictly speaking this change could be reverted before release, of course. The commit to change it is only a few days old as of this writing, so if you want to play it safe you could wait a bit before considering a merge.

minad:
> Strictly speaking this change could be reverted before release, of
> course. The commit to change it is only a few days old as of this
> writing, so if you want to play it safe you could wait a bit before
> considering a merge.

I hope the commit is kept. I like structs better than clos.

All I am waiting for is another xref release which includes this change
- I mean the xref package on ELPA, which can also be installed on Emacs
26 and 27.

northgaard:
> All I am waiting for is another xref release which includes this change - I mean the xref package on ELPA, which can also be installed on Emacs 26 and 27.

Ah, gotcha. As far as I can tell there has not yet been a new ELPA release.

minad:
@dgutov Do you plan to release a new xref version with this change sometimes soon?

dgutov:
@minad Hi Daniel,

The new version will be tagged soon, but I think you don't need to worry about that: you should be able to write the code in a way that works the both the new and the old versions.

`oref` is indeed out of the question now, but try using `xref-location-group`, `xref-location-line` and `xref-file-location-column`: they should work with both underlying implementations.

Not sure what you need `xref-file-location-column` for, though. The visualization code usually doesn't need that.

Also: why aren't you using `xref-location-marker` with `xref-file-location` type as well? If you did, this issue would not even arise.

minad:
Hi Dmitry, thanks for answering. The reason why I break the xref abstraction barrier is that I want to have fine grained control over the files being opened during preview in the consult completion commands. Consult ensures that previewed files are closed afterwards. I could also determine opened files via some other means, e.g., hooks. This would allow to use xref-location-marker directly. However Consult is written in a very conservative way currently, e.g., large files are opened in raw mode for preview. This matters in particular for automatic preview. You are right that one can also write this code with the existing accessor functions, but I think the accessor xref-location-group does a bit too much. That's why I went with oref originally. Maybe it is still possible to use it and then one would end up with better more abstract code. Overall the Consult preview for closed files (in comparison to buffer preview) is a bit brittle and over engineered. I am not sure I would implement it again like this when starting anew, in particular since I don't generally recommend automatic preview for files. Despite all these efforts file preview is still slow. So given that file preview should be mostly manual one could also do away with all the complications and just open the file without cleaning it up afterwards. But then this kind of "preview" is not really a preview anymore but just like an action on the candidate (Helm,Ivy,Embark-action I mean). So this is why I am sticking to the design.

dgutov:
> But it is a bit like breaking the abstraction in another subtle way ;)

A little bit. But less so than checking the type explicitly. Or even using `oref` to peek at the object's private parts. :)

dgutov:
Thanks for verifying the migration path, BTW.

This ended up being the first third-party package to be updated. :+1:

PR #460, opened 2021-11-03
consult--format-location: remove extra colon if not needed

gagbo:
Ref gagbo/consult-lsp#21


minad:
I added the change in a slightly different form. Note that I relied on the current behavior of always appending a colon in `consult-global-mark`.

Issue #461, opened 2021-11-07
Eshell completion with `consult-completion-in-region` unexpected behaviors


svaante:
First off, great package 👍 

I have a feeling this is due to eshell completion quirks and not consult. So feel free to close this if this is the case, or if anybody has any suggestion on how to make this work.

The expected behavior is how the \*Completions\* buffer is populated and what text is inserted when the completion is expanded.

`emacs -Q`

```emacs-lisp
(package-initialize)
(require 'consult)
(require 'vertico)
(vertico-mode)

;; substring doesn't work with the completions described below
(setq completion-styles '(basic)) 

(setq completion-in-region-function
      (lambda (&rest args)
        (apply (if vertico-mode
                   #'consult-completion-in-region
                 #'completion--in-region)
               args)))
```
## First example

M-x eshell
```sh
$ mkdir somefolder
$ cd somefolder
$ touch somefile
$ chmod +x ./somefile
$ ./some<Tab>
```
Outcome:
![Screenshot 2021-11-07 at 01 53 34](https://user-images.githubusercontent.com/1129021/140628302-fb722591-83e1-4401-9ddb-d3bb784c451a.png)

Expected result would be `$ ./some<Tab> -> ./somefile`

## Second example
M-x eshell
```sh
$ mkdir somefolder
$ cd somefolder
$ touch 1 2 3
$ ls <Tab>
```
Outcome:
![Screenshot 2021-11-07 at 01 43 45](https://user-images.githubusercontent.com/1129021/140628105-d04bdeea-2df0-4821-9ee3-9da306c04aa2.png)

Expected result would be that the completion list is populated with 1 2 3

minad:
I just tried to reproduce this and it works as expected on Emacs 27.1. I don't see your results but the result you described as expected. Is your Consult version up to date? Closing this for now. Please let me know how it goes on your end.

svaante:
Thanks for looking into it. I am still able to reproduce the issue on ﻿﻿12ec003. I am using emacs 28.0.50.

Did some digging around for the second issue.
https://github.com/minad/consult/blob/3b18b04524f0e963070ab6c61d4d3a3f10507d12/consult.el#L2343
At this point `initial` is `""` which is then passed as `dir` which might be undefined behavior. If we pass `nil` `read-file-name-default` will use `default-directory` which then leads to the expected behavior. I can't seem to find the commit responsible on the 28 branch. Which changes the behavior of `read-file-name-default` or if that is the case at all.

[...]

PR #463, opened 2021-11-07
consult-completion-in-region: insert initial text for file completion

oantolin:
Previously this was done by using initial as the DIR argument of
read-file-name, but that relied on undocumented behavior of that
function. With this change, we explicitly insert the initial text.

This fixes one of the two issues mentioned in minad/consult#461

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -2334,13 +2334,15 @@ These configuration options are supported:
                                         'consult--completion-candidate-hook)
                      (let ((enable-recursive-minibuffers t))
                        (if (eq category 'file)
-                           ;; When completing files with consult-completion-in-region, the point in the
-                           ;; minibuffer gets placed initially at the beginning of the last path component.
-                           ;; By using the filename as DIR argument (second argument of read-file-name), it
-                           ;; starts at the end of minibuffer contents, as for other types of completion.
-                           ;; However this is undefined behavior since initial does not only contain the
-                           ;; directory, but also the filename.
-                           (read-file-name prompt initial initial require-match nil predicate)
+                           ;; We use read-file-name, since many completion UIs make it nicer to
+                           ;; navigate the file system this way; and we insert the initial text
+                           ;; directly into the minibuffer to allow the user's completion
+                           ;; styles to expand it as appropriate (particularly useful for the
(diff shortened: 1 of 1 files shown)


consult.el
```diff
@@ -2334,13 +2334,9 @@ These configuration options are supported:
...
                      (let ((enable-recursive-minibuffers t))
                        (if (eq category 'file)
-                           ;; When completing files with consult-completion-in-region, the point in the
-                           ;; minibuffer gets placed initially at the beginning of the last path component.
-                           ;; By using the filename as DIR argument (second argument of read-file-name), it
-                           ;; starts at the end of minibuffer contents, as for other types of completion.
-                           ;; However this is undefined behavior since initial does not only contain the
-                           ;; directory, but also the filename.
-                           (read-file-name prompt initial initial require-match nil predicate)
+                           (minibuffer-with-setup-hook
```
minad, line 2337: Two small suggestions:

1. Use consult--minibuffer-with-setup-hook which fixes some issue with minibuffer-with-setup-hook. I do this consistently throughout consult I think.
2. Still keep the comment partially or explain the history of the change? Usually such subtleties are forgotten again after some time.

oantolin: 1. Sure, I'll change it. I forgot about that problem. It was a memory leak, right? And your macro uses a symbol, setting its function value, for the hook instead, I believe.
2. In my view, the comment explains a hack and the new code is a more or less direct expression of intent, so it doesn't need a comment. But I can certainly add some form of comment.

minad: Let's just add a comment which links to the issue.

oantolin: Oh, sorry, I made some changes before I saw that. I added a comment that explains why it's important to insert the full initial text into the minibuffer. But I still don't like to the issue.

minad:
Thanks!

oantolin:
No problem! Unlike you, I actually use consult-completion-in-region a lot. (Although more and more I turn on Corfu, I might actually switch to Corfu full time at some point.)

minad:
> (Although more and more I turn on Corfu, I might actually switch to Corfu full time at some point.)

That's good to hear - as long as you continue to look into consult-completion-in-region bugs ;)

PR #467, opened 2021-11-12
Reformat wide docstring in consult and consult-selectrum

piotrkwiecinski:
Fix warnings from the compiler about docstrings longer than 80 characters.


minad:
Thank you!

PR #495, opened 2022-01-10
Improve focus-lines speed using multi-line overlays

### LINKED ISSUE #493: consult-focus-lines point remains in invisible region
I enjoy consult-focus-lines in the package-list-packages listing.  I've noticed however that if point is in the text to be hidden, it remains there.  So that if you focus to "consult" (for example) and attempt to mark the first item with "i", the item in the full buffer you were on before focus is marked instead.  Perhaps it would be possible to move point to the nearest position outside any invisible overlay?

jdtsmith:
This improves the speed of `consult-focus-lines` by:

- Storing line number + line begin & end as text properties on the line candidates to be matched.
- Extracting this information from the text properties in resulting matches.
- Using the matched line data to merge contiguous blocks of lines, substantially reducing the number of overlays needed.  Overlays are not pre-allocated, but instead deleted and re-created on each new input.

To ensure every line has the `'line` text property added, blank lines are collected as `\n`.  I tested this on a file with >250,000 lines (`/usr/share/dict/words`).  The current version does not make it to the prompt after several minutes of waiting.  This is likely due to the pre-allocation of one overlay per line.  This version works reasonably well on such a long file.  With fewer overlays, it also modestly improves navigation speed in the buffer, likely due to reducing the need to skip over the many invisible overlays.

In addition, it:

- Correctly exits the current search on interruption
- Provides a message update during the search with the number of included/excluded lines.
- Moves point past the enclosing invisible overlay when finished [1]. Fixes #493.

[1] Note that emacs normally moves point out of invisible regions after commands, but not at/near the beginning of buffer (where it moves to `(point-min)`, which may be hidden).

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -3045,7 +3045,6 @@ INITIAL is the initial input."
 		   (finish nil)
 		   (max (point-max)))
 	      (unless (eq matches t)	;input arrived
-		(message "%s %d lines" (if not "Excluded" "Included") (length matches))
 		(while matches
 		  (pcase-let* ((`(,ind ,beg ,end) (car matches))
 			       (new-block (> (- ind old-ind) 1)))
--- a/consult.el
+++ b/consult.el
@@ -2993,10 +2993,6 @@ INITIAL is the initial input."
 
 ;;;;; Command: consult-focus-lines
 
-(defsubst consult--extract-line-ind (matches)
-  "Extract and sort (line beg end) from MATCHES."
-  (mapcar (apply-partially #'get-text-property 0 'line) matches))
-
 (defun consult--focus-lines-state (filter)
   "State function for `consult-focus-lines' with FILTER function."
   (let ((lines) (overlays) (last-input) (point-orig (point)))
@@ -3027,55 +3023,54 @@ INITIAL is the initial input."
       (when (and input (not (equal input last-input)))
 	(mapc #'delete-overlay overlays)
 	(setq last-input input overlays nil)
-        (unless (string-match-p "\\`!? ?\\'" input) ; empty input.
-          (let* ((not (string-prefix-p "! " input))
-                 (stripped (string-remove-prefix "! " input))
-                 ;; Heavy computation is interruptible if *not* committing!
-                 (matches (if restore
-			      (consult--extract-line-ind
-			       (funcall filter stripped lines))
-			    (while-no-input
-			      (consult--extract-line-ind
-			       (funcall filter stripped lines)))))
+	(unless (string-match-p "\\`!? ?\\'" input) ; empty input.
+	  (let* ((not (string-prefix-p "! " input))
+		 (stripped (string-remove-prefix "! " input))
+		 ;; Heavy computation is interruptible if *not* committing!
+		 (matches (if restore
+			      (funcall filter stripped lines)
+			    (while-no-input (funcall filter stripped lines))))
 		 (old-ind 0)
 		 (start (point-min))
 		 (finish nil)
 		 (max (point-max)))
 	    (unless (eq matches t)	;input arrived
 	      (while matches
-		(pcase-let* ((`(,ind ,beg ,end) (car matches))
+		(pcase-let* ((`(,ind ,beg ,end)
+			      (get-text-property 0 'line (car matches)))
 			     (new-block (> (- ind old-ind) 1)))
 		  (if not
 		      ;; exclude: hide from beg to 1+end of the last contiguous block
 		      (progn (when new-block
 			       (if finish
-				   (push (consult--overlay start finish 'invisible t) overlays)
-				 )
+				   (push (consult--overlay start finish 'invisible t) overlays))
 			       (setq start beg))
 			     (setq finish (1+ end)))
 		    ;; include: hide from 1+end of last block to beg of new block
 		    (when new-block
-		      (push (consult--overlay start beg  'invisible t) overlays))
+		      (push (consult--overlay start beg 'invisible t) overlays))
 		    (setq start (1+ end)))
 		  (setq matches (cdr matches) old-ind ind)))
-	      (if not (when finish	; finish up
-			(setq finish (min finish max))
-			(if (> finish start)
-			  (push (consult--overlay start finish 'invisible t) overlays)))
-		(if (> max start) ; may hide ALL
+	      (if not
+		  (when finish		; finish up
+		    (setq finish (min finish max))
+		    (if (> finish start)
+			(push (consult--overlay start finish 'invisible t) overlays)))
+		(if (> max start)	; may hide ALL
 		    (push (consult--overlay start max 'invisible t) overlays)))))))
       (when restore
-        (cond
-         ((not input)
+	(cond
+	 ((not input)
 	  (mapc #'delete-overlay overlays)
-          (goto-char point-orig))
+	  (goto-char point-orig))
 	 ((equal input "")
-          (consult-focus-lines 'show)
-          (goto-char point-orig))
-         (t
-          ;; Sucessfully terminated -> Remember invisible overlays
-	  (setq consult--focus-lines-overlays (seq-copy overlays))
-	  (if-let ((invisible-p point-orig) ;move point past invisible  
+	  (setq consult--focus-lines-overlays (seq-copy overlays)) ;;XXX loses old overlays
+	  (consult-focus-lines 'show)
+	  (goto-char point-orig))
+	 (t
+	  ;; Sucessfully terminated -> Remember invisible overlays
+	  (consult--focus-lines-overlays (seq-copy overlays))
+	  (if-let ((invisible-p point-orig) ;move point past invisible
 		   (ovs (overlays-at point-orig))
 		   (ov (seq-find (lambda (ov) (overlay-get ov 'invisible)) ovs)))
 	      (goto-char (overlay-end ov))
--- a/consult.el
+++ b/consult.el
@@ -2995,8 +2995,7 @@ INITIAL is the initial input."
 
 (defsubst consult--extract-line-ind (matches)
   "Extract and sort (line beg end) from MATCHES."
-  (sort (mapcar (apply-partially #'get-text-property 0 'line) matches)
-	(lambda (a b) (< (car a) (car b)))))
+  (mapcar (apply-partially #'get-text-property 0 'line) matches))
 
 (defun consult--focus-lines-state (filter)
   "State function for `consult-focus-lines' with FILTER function."
@@ -3021,7 +3020,8 @@ INITIAL is the initial input."
 		  (if (eq beg end) (char-to-string ?\n) ; "\n" reuses string! 
 		    (buffer-substring-no-properties beg end)))
 	    (add-text-properties 0 1 `(line (,(cl-incf i) ,beg ,end)) line)
-	    (push line lines)))))
+	    (push line lines))
+	  (setq lines (nreverse lines)))))
     (cl-labels ((add-inv-ov (beg end)
 			    (push (consult--overlay beg end 'invisible t)
 				  overlays)))
--- a/consult.el
+++ b/consult.el
@@ -3022,64 +3022,66 @@ INITIAL is the initial input."
 	    (add-text-properties 0 1 `(line (,(cl-incf i) ,beg ,end)) line)
 	    (push line lines))
 	  (setq lines (nreverse lines)))))
-    (cl-labels ((add-inv-ov (beg end)
-			    (push (consult--overlay beg end 'invisible t)
-				  overlays)))
-      (lambda (input restore)
-	;; New input provided -> Update
-	(when (and input (not (equal input last-input)))
+    (lambda (input restore)
+      ;; New input provided -> Update
+      (when (and input (not (equal input last-input)))
+	(mapc #'delete-overlay overlays)
+	(setq last-input input overlays nil)
+        (unless (string-match-p "\\`!? ?\\'" input) ; empty input.
+          (let* ((not (string-prefix-p "! " input))
+                 (stripped (string-remove-prefix "! " input))
+                 ;; Heavy computation is interruptible if *not* committing!
+                 (matches (if restore
+			      (consult--extract-line-ind
+			       (funcall filter stripped lines))
+			    (while-no-input
+			      (consult--extract-line-ind
+			       (funcall filter stripped lines)))))
+		 (old-ind 0)
+		 (start (point-min))
+		 (finish nil)
+		 (max (point-max)))
+	    (unless (eq matches t)	;input arrived
+	      (while matches
+		(pcase-let* ((`(,ind ,beg ,end) (car matches))
+			     (new-block (> (- ind old-ind) 1)))
+		  (if not
+		      ;; exclude: hide from beg to 1+end of the last contiguous block
+		      (progn (when new-block
+			       (if finish
+				   (push (consult--overlay start finish 'invisible t) overlays)
+				 )
+			       (setq start beg))
+			     (setq finish (1+ end)))
+		    ;; include: hide from 1+end of last block to beg of new block
+		    (when new-block
+		      (push (consult--overlay start beg  'invisible t) overlays))
+		    (setq start (1+ end)))
+		  (setq matches (cdr matches) old-ind ind)))
+	      (if not (when finish	; finish up
+			(setq finish (min finish max))
+			(if (> finish start)
+			  (push (consult--overlay start finish 'invisible t) overlays)))
+		(if (> max start) ; may hide ALL
+		    (push (consult--overlay start max 'invisible t) overlays)))))))
+      (when restore
+        (cond
+         ((not input)
 	  (mapc #'delete-overlay overlays)
-	  (setq last-input input overlays nil)
-          (unless (string-match-p "\\`!? ?\\'" input) ; empty input.
-            (let* ((not (string-prefix-p "! " input))
-                   (stripped (string-remove-prefix "! " input))
-                   ;; Heavy computation is interruptible if *not* committing!
-                   (matches (if restore
-				(consult--extract-line-ind
-				 (funcall filter stripped lines))
-			      (while-no-input
-				(consult--extract-line-ind
-				 (funcall filter stripped lines)))))
-		   (old-ind 0)
-		   (start (point-min))
-		   (finish nil)
-		   (max (point-max)))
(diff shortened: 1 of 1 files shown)


minad:
Thank you for looking into this! I will add a few comments.

consult.el
```diff
@@ -3010,48 +3015,72 @@ INITIAL is the initial input."
...
+				 (funcall filter stripped lines))
+			      (while-no-input
+				(consult--extract-line-ind
+				 (funcall filter stripped lines)))))
+		   (old-ind 0)
+		   (start (point-min))
+		   (finish nil)
+		   (max (point-max)))
+	      (unless (eq matches t)	;input arrived
+		(message "%s %d lines" (if not "Excluded" "Included") (length matches))
```
minad, line 3048: Please remove this message. We can discuss this separately, but I'd like to have the changes at least as independent commits.

consult.el
```diff
@@ -3010,48 +3015,72 @@ INITIAL is the initial input."
...
+        (let ((i 0) line)
+	  (consult--each-line beg end
+	    (setq line
+		  (if (eq beg end) (char-to-string ?\n) ; "\n" reuses string! 
+		    (buffer-substring-no-properties beg end)))
+	    (add-text-properties 0 1 `(line (,(cl-incf i) ,beg ,end)) line)
+	    (push line lines)))))
+    (cl-labels ((add-inv-ov (beg end)
+			    (push (consult--overlay beg end 'invisible t)
+				  overlays)))
```
minad, line 3027: This function should be inlined. I rarely use cl-labels and in this case it is not justified, since the function is tiny.

consult.el
```diff
@@ -2993,6 +2993,11 @@ INITIAL is the initial input."
 
 ;;;;; Command: consult-focus-lines
 
+(defsubst consult--extract-line-ind (matches)
+  "Extract and sort (line beg end) from MATCHES."
+  (sort (mapcar (apply-partially #'get-text-property 0 'line) matches)
+	(lambda (a b) (< (car a) (car b)))))
```
minad, line 2999: Why do we have to sort? Aren't the lines already sorted?

jdtsmith: They are reversed via push, but that's an easy fix.  I was concerned that `completions-all-completions` doesn't have set the `display-sort-function` metadata when called, so that someone's completion style could interfere.  Is that a real issue?

consult.el
jdtsmith: It's used 4 times in the potentially hard to parse and understand core algorithm, so I thought it would be easier for future readers.

consult.el
jdtsmith: OK figured, but thought I'd prime the pump.  The better solution of course would be a consult-buffer style prefix of `#lines shown/#total lines`, but I didn't know how to do that.

consult.el
minad: It is still only a single line and I don't consider this a "logical functional unit", which deserves to be factored out. Also it potentially makes the code slower. In other parts of Consult I use a functional programming style heavily, e.g., check out the async pipeline, but from my experience it does not feel justified here.

consult.el
minad: There is no issue with sorting. The filter function preserves the original order. So we should go with the reverse.

consult.el
minad: Yes, I would accept a PR which adds such an overlay for `consult-focus-lines` and `consult-keep-lines`. But it should be a separate PR.

jdtsmith:
OK can refactor.  Let me know your thoughts on sort necessity and cl-labels.  All, one other thing I wasn't clear on that should be fixed now: the current behavior adds to the overlays on sequential focus calls, so that you can stack filters on top of each other.  This might be useful, but as it is, it's broken because only one list of overlays is saved (the latest), so on "clearing" (blank input) only the most recent are unhidden.  

I think the best approach is to have one level deep of filter, and so `(consult-focus-lines 'show)` on each new call to focus-lines.  Thoughts?

consult.el
jdtsmith: Fixed.

consult.el
jdtsmith: I understand.  Is a `defsubst` also out of the question?

minad:
> All, one other thing I wasn't clear on that should be fixed now: the current behavior adds to the overlays on sequential focus calls, so that you can stack filters on top of each other. This might be useful, but as it is, it's broken because only one list of overlays is saved (the latest), so on "clearing" (blank input) only the most recent are unhidden.

Oh, I had assumed that this works correctly. Please preserve the existing behavior. But I was also about to ask if we should improve the situation somehow. Instead of layering the overlays, combine them for better efficiency for example. I wouldn't go with the idea of stacking, which is too complex. But any of this can be addressed separately.

jdtsmith:
> > All, one other thing I wasn't clear on that should be fixed now: the current behavior adds to the overlays on sequential focus calls, so that you can stack filters on top of each other. This might be useful, but as it is, it's broken because only one list of overlays is saved (the latest), so on "clearing" (blank input) only the most recent are unhidden.
> 
> Oh, I had assumed that this works correctly. Please preserve the existing behavior. 

Since overlays are recreated on each call to the closure, the issues is we'll have duplicates if we allow the old overlays to persist.  BTW, I was always confused by this: taking the `package-list-packages` example, I don't expect to have to "clear" the search to start a new one.  I just want to `M-s f consult`, check out some cool packages, then `M-s f eldoc`, check some more, etc.  In fact I do that _every time_ and spend a moment perplexed by the blank buffer.  I'd strongly suggest we just adopt a "single filter at a time" approach, unless you've seen usage of stacked filters in the wild. 

If that's still the approach we want, we must think about _merging_ the new and saved overlay lists.  But to do that well, it will require merging/coalescing overlays that form contiguous series.  Painful, and hard to get right for all corner cases.  I'd like to avoid this if at all possible.  

> I wouldn't go with the idea of stacking, which is too complex.

That's how I'm describing the current behavior of "add more filters blocks to the filters".  So IMO current "stacking" is the default.  If we really go this route, I think some feedback to the user that this is happening would be appropriate (filter in place & lines shown, for example).  And then rather than merging overlays, the easiest way might in fact be to compose a filter as `filter1 AND filter2`, delete all overlays and regenerate.  So save the former input as well.  Getting pretty complicated for such a simple tool!

minad:
Currently the behavior is quite simple and reminiscent of the `consult-keep-lines` behavior. Therefore it is designed like this. I prefer to keep this behavior. If you want to clear first it is easy to add such a small helper command to your config.

jdtsmith:
> Currently the behavior is quite simple and reminiscent of the `consult-keep-lines` behavior. Therefore it is designed like this. I prefer to keep this behavior. If you want to clear first it is easy to add such a small helper command to your config.

In that case I'd say we need:

1. A way to indicate to the user what filter(s) are _already in place_.
2. To _compose filters_ rather than try to _merge overlays_.  Something like a saved list of `consult--focus-line-filter-inputs`, that get prepended to the current filter (and cleared on reset), ala 
    ```
    (filter_n ( ... ( filter2 (filter 2 inputs lines))))...
    ``` 
    Such a list would obviously help with 1 as well.

minad:
> In that case I'd say we need:

No, we don't need that here. Please let's keep this PR focused. I am open to discussing the overlay composition in a separate issue however. Currently the composition is simply achieved by reusing the existing overlays and putting all the load on the Emacs display engine. This PR should do the same. 

I assume that we both agree that this is not particularly friendly, so it deserves to be improved in the next step. :)

consult.el
```diff
@@ -3010,48 +3014,74 @@ INITIAL is the initial input."
...
-                       (while-no-input
-                         (consult--string-hash (funcall filter stripped lines))))))
-            (when (hash-table-p ht)
-              (let ((ov overlays) (li lines))
-                (while ov
-                  (overlay-put (car ov) 'invisible (eq not (gethash (car li) ht)))
-                  (setq li (cdr li) ov (cdr ov))))
-              (setq last-input input)))))
+                 (matches (if restore
+			      (consult--extract-line-ind
```
minad, line 3035: consult--extract-line-ind should be inlined. Not here but later in the while loop. We should avoid allocating a new list with the mapcar.

consult.el
```diff
@@ -3010,48 +3014,74 @@ INITIAL is the initial input."
...
+			    (while-no-input
+			      (consult--extract-line-ind
+			       (funcall filter stripped lines)))))
+		 (old-ind 0)
+		 (start (point-min))
+		 (finish nil)
+		 (max (point-max)))
+	    (unless (eq matches t)	;input arrived
+	      (while matches
+		(pcase-let* ((`(,ind ,beg ,end) (car matches))
```
minad, line 3046: Extract the text property here?

jdtsmith:
> > In that case I'd say we need:
> 
> No, we don't need that here. Please let's keep this PR focused. I am open to discussing the overlay composition in a separate issue however. Currently the composition is simply achieved by reusing the existing overlays and putting all the load on the Emacs display engine. This PR should do the same.

There was never overlap before because you had only one overlay per line. Overlays can now be thousands of lines long or more.  I suppose I can just let overlays overlap willy nilly and let emacs sort it out (as there's no such thing as "double invisible").  Feels wrong...
 
> I assume that we both agree that this is not particularly friendly, so it deserves to be improved in the next step. :)

Yes.  Here's a potential UI I sketched up for later:

image

minad:
> There was never overlap before because you had only one overlay per line. Overlays can now be thousands of lines long or more. I suppose I can just let overlays overlap willy nilly and let emacs sort it out (as there's no such thing as "double invisible"). Feels wrong...

Well, with the current implementation you can also get two overlays at the same place both saying "invisible". If this is wrong or not is debatable. Emacs allows me to do this. I doubt that having overlaps is an issue here, as far as I understood the performance issue for Emacs is the sheer number of overlays. And since we are getting that number down we will be better than before, with overlap or without.

consult.el
jdtsmith: Good idea, done.

jdtsmith:
OK the ugly overlap deed is done.  Seems to work; can you give a quick test?

minad:
I found one issue. Narrowing didn't work. I also simplified it a little bit on the go, see
https://github.com/minad/consult/commit/92fd226ba8c87543d6e08632c980722cf79f67f7.

jdtsmith:
How was narrowing fixed?

Also, this is indeed simple, but does not produce correct results for `not` matches at the end.  E.g. consider:

```
aaa
bbb
aaa
bbb
ccc
ccc
ccc
```

with input `! c`.  The reason is the clever new "out of matches" end finisher never gets a chance to run when block-end = point-max.

minad:
Thanks, for checking! Fixed. See the latest commits.

minad:
I think this issue was introduced by the unrelated optimization https://github.com/minad/consult/commit/21501968fe3256ad23ed09b98d5e6bc435a9f69f. The "\n" hack you used is not really a good idea good. I'd like to replace this with something else.

minad:
Hmm, or maybe your proposed fix to check for "\n" handles all the scenarios? I will merge this for now.

jdtsmith:
Yep, you are re-treading the slippery path I traversed ;).  Only text properties survive.  You can also check that length is 1, but given the construction of lines, they will never have a newline character other than if blank.

minad:
That's not exactly true. The minor optimization 21501968fe3256ad23ed09b98d5e6bc435a9f69f goes a little bit beyond. Anyways dissecting the code again is what I call a thorough review. ;)

Generally I try to avoid a finalization step if I can pull it inside the loop.

jdtsmith:
I just meant the pitfalls I encountered.  And great work finding a way to avoid finalization — I agree that's not ideal. With the latest, it's a pretty lean and mean (and simple!) algorithm: good teamwork.  

One thing I've noticed playing with long files is that while-no-input doesn't seem to do much for you.  The reason is that it is wrapping the relatively inexpensive (i.e. not so heavy) part of the algorithm: the completion-all-completions match.  The fix was easy and in fact simplifies things even further; see #497.

minad:
> it's a pretty lean and mean (and simple!) algorithm: good teamwork.

Yes, thank you for that! By overhauling the code again I played the ball back to you for another review. :)

> One thing I've noticed playing with long files is that while-no-input doesn't seem to do much for you. The reason is that it is wrapping the relatively inexpensive (i.e. not so heavy) part of the algorithm: the completion-all-completions match. The fix was easy and in fact simplifies things even further; see #497.

Yes, I am aware of this. Filtering is usually the heaviest part, e.g., in Vertico. But here the overlay creation may be more costly. I have some doubts about this though.

oantolin:
Woah, did you guys know about the built-in `rtree` library? It provides an efficient tree-based data structure for storing sets of disjoints intervals of numbers and common operations for these sets. Maybe it could be used to implement the "update overlays" approach in a way that isn't horribly inconvenient.

minad:
I don't know about this library. But I doubt that we need anything sophisticated here. The current implementation is straightforward. We don't implement overlay updates/merging.

oantolin:
> We don't implement overlay updates/merging.

I know it's not currently done, I was just saying that perhaps with that library's help it wouldn't be hard to do so. But now that I look more at the API, probably not.

minad:
> I know it's not currently done, I was just saying that perhaps with that library's help it wouldn't be hard to do so.

Sure, I understood you that way. I'd argue that the current implementation is good enough. It is much more usable than before. Also `consult-keep-lines` unlocks read-only buffers now. I am much more happy with these commands now. :)

But @jdtsmith has some ideas on how to improve the UI (#496).

jdtsmith:
The latest greatly reduces the number of overlays used, and cuts out the "shortest match" bottleneck by making all the expensive operations (overlays included) interruptible.  So no longer do you type a single letter and wait a while for it to update.  

If we wanted to make focus-lines a fancier tool that allows stacking/unstacking multiple filters, maybe a library like rtree could be useful.  Right now, multiple rounds of filtering just create independent overlays and let the display engine sort it out.  With orderless, I claim even that is unnecessary (as you can stack include & exclude filters all in one filtering level).  Part of the UI issue is that allowing multiple independent sequential rounds of filtering will complicate keeping track of how many lines are still showing, whereas with "one level of filtering only", it's trivial.

BTW, in another thread someone mentioned the power of embark-export + focus-lines + marginalia for selecting on "all metadata", which hadn't occurred to me.  The one issue I found is that embark-export is quite slow on large lists of the sort I'd like to filter this way (e.g. `C-h v C, E`), maybe because it's building a customize widget for each variable.

oantolin:
I think that the annotators are the slowest part of embark export. You normally don't notice how slow they are because Vertico, Selectrum or Icomplete-vertical annotate only 10 candidates or so at a time. But embark collect or the built-in `*Completions*` buffer annotate all candidates and then you can really tell it takes a while.

minad:
@jdtsmith 

> If we wanted to make focus-lines a fancier tool that allows stacking/unstacking multiple filters, maybe a library like rtree could be useful. Right now, multiple rounds of filtering just create independent overlays and let the display engine sort it out. With orderless, I claim even that is unnecessary (as you can stack include & exclude filters all in one filtering level). Part of the UI issue is that allowing multiple independent sequential rounds of filtering will complicate keeping track of how many lines are still showing, whereas with "one level of filtering only", it's trivial.

To be honest, I am not sure I want this. `consult-keep-lines` and `consult-focus-lines` are simple tools, which aim to be better `keep-lines` commands. But I am not sure we should make them arbitrarily complex. The commands should solve the most common need that users have! But for now I think having a "count/total" indicator is a good idea, as proposed in #496.

> BTW, in another thread someone mentioned the power of embark-export + focus-lines + marginalia for selecting on "all metadata", which hadn't occurred to me. The one issue I found is that embark-export is quite slow on large lists of the sort I'd like to filter this way (e.g. C-h v C, E), maybe because it's building a customize widget for each variable.

Actually `consult-focus-lines` and `consult-keep-lines` were specifically developed for the `embark-collect-snapshot` workflow described by @oantolin. At that time I didn't want to implement a `consult-descbinds` command. If I recall correctly, we argued that it is better to have a general export an post filtering functionality. However in the end we still got `embark-bindings`. :)

@oantolin

> I think that the annotators are the slowest part of embark export. You normally don't notice how slow they are because Vertico, Selectrum or Icomplete-vertical annotate only 10 candidates or so at a time. But embark collect or the built-in *Completions* buffer annotate all candidates and then you can really tell it takes a while.

Yes, but for the rare occasions one uses this it is good enough. Recently I didn't make much use of this functionality except for testing Marginalia annotators, `M-x`, or `C-h v` and then snapshot everything :laughing:

jdtsmith:
> consult-keep-lines and consult-focus-lines are simple tools, which aim to be better keep-lines commands

I agree on keeping it simple.  In fact I advocate making focus-lines even _more simple_: only one filter allowed at a time (new filter overrides any previous).  For people who want/need to add multiple filters, recommending an orderless search like `Focus lines: keep ![0-9]\{2,\}` or similar is a good workaround.  It is now perfectly fast enough to build this up interactively.  And it will make `count/total` trivial to do.

> M-x, or C-h v and then snapshot everything 😆

Oops I forgot the distinction between export and collect snapshot.  For me, even exporting after a variable search like `consult` is a 5-10s affair. I think it must be because of the fancy customize interface.  Must remember collect...  

BTW, somebody should make a gist with "awesome consult + embark + marginalia + orderless tricks" and one-liners.  Probably could post on reddit and collect a bunch of these from people's workflows.

oantolin:
Oh, sorry, I read your earlier comment too quickly and didn't notice you said *export*! Yeah, it's probably slow for different reasons than collect. 😛

oantolin:
> I agree on keeping it simple. In fact I advocate making focus-lines even _more simple_: only one filter allowed at a time (new filter overrides any previous).

I'd like that for consult-focus-line, but I *also* like it being semantically similar to keep-lines. One of those two wishes will go unfulfilled.

minad:
@jdtsmith

> For people who want/need to add multiple filters, recommending an orderless search like...

Note that you cannot assume Orderless.

> BTW, somebody should make a gist with "awesome consult + embark + marginalia + orderless tricks" and one-liners. Probably could post on reddit and collect a bunch of these from people's workflows.

These should be put them in the respective wikis. This is why the wikis exist.

@oantolin 

> I'd like that for consult-focus-line, but I also like it being semantically similar to keep-lines. One of those two wishes will go unfulfilled.

Since we've already got the version which is semantically similar to keep-lines it will stay that way. This makes it harder or less efficient to implement the count/total indicator, but I also argue that we don't really need that.

PR #497, opened 2022-01-11
Wrap all heavy computations in while-no-input

jdtsmith:
Currently `while-no-input` wraps the filtering portion of focus-lines.  But in fact the computation and creation of overlays is the heavier computational load, especially when there are many matches.  This PR wraps both filtering and overlay computation in `while-no-input`, disabling this if `restore` is set.  Since the first (least general) part of the input matches the most lines and is therefore (by far) the slowest, this _really_ helps with apparent performance for large files.  With this fix, searching with several letter long inputs in `/usr/dict/words` (236k lines) happens ~instantaneously.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -3023,32 +3023,34 @@ INITIAL is the initial input."
     (lambda (input restore)
       ;; New input provided -> Update
       (when (and input (not (equal input last-input)))
-        (mapc #'delete-overlay overlays)
+	(mapc #'delete-overlay overlays)
         (setq last-input input overlays nil)
         (unless (string-match-p "\\`!? ?\\'" input) ;; empty input.
-	  (while-no-input
-	    (let* ((inhibit-quit restore)
-		   (not (string-prefix-p "! " input))
-		   (stripped (string-remove-prefix "! " input))
-		   (matches (funcall filter stripped lines))
-		   (old-ind 0)
-		   (block-beg pt-min)
-		   (block-end pt-min))
-	      (while old-ind
-		(let ((match (pop matches)) (ind nil) (beg pt-max) (end pt-max) prop)
-		  (when match
-		    (setq prop (get-text-property 0 'consult--focus-line match)
-			  ind (car prop)
-			  beg (cdr prop)
-			  ;; NOTE: Check for empty lines, see above!
-			  end (+ 1 beg (if (equal match "\n") 0 (length match)))))
-		  (unless (eq ind (1+ old-ind))
-		    (let ((a (if not block-beg block-end))
-			  (b (if not block-end beg)))
-		      (when (/= a b)
-			(push (consult--overlay a b 'invisible t) overlays)))
-		    (setq block-beg beg))
-		  (setq block-end end old-ind ind)))))))
+	  (when (while-no-input
+		  (let* ((inhibit-quit restore)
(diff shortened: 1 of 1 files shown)


minad:
I've already considered this, but I am skeptical. Does this not lead to inconsistent state, where the buffer is displayed partially filtered?

jdtsmith:
Yes I suppose it it can, but since with the current design, the very next input removes all overlays and starts over again, the situation does not persist.  I think checking for interruption and clearing the current (possibly partial) overlays would suffice, on the notion that another input will soon arrive. I can look at that if you like. The performance benefits are really impressive.

minad:
Okay, it should work like this:

1. Preserve old overlays
2. Create new overlays in an interruptible fashion
3. If interrupted, delete new overlays again, preserve old ones
4. If not interrupted, keep the new ones, delete the old ones

jdtsmith:
I think what I just pushed does this (effectively), since old overlays are all stored externally to the closure.  Since overlays are always re-computed fresh for each round of input now, this makes it easier to cope with one interruption.

minad:
Hmm, I don't think the current code is correct. First you delete the overlays, which is wrong, then we create the new ones in an interruptible fashion. If interrupted we also delete the new ones.

jdtsmith:
The initial deletion of old overlays (on new input) is what we are already doing now.  The only addition here is to delete any "recently accumulated but necessarily incomplete due to interruption" overlays.

minad:
> The initial deletion of old overlays (on new input) is what we are already doing now. The only addition here is to delete any "recently accumulated but necessarily incomplete due to interruption" overlays.

Yes, I know. We should correct this.

jdtsmith:
> Yes, I know. We should correct this.

IMO, there is nothing to correct.  With our new algorithm, the old overlays span many many lines.  New overlays will span many _other_ lines.  The effort required to retain and mutate old overlays is far greater than any potential savings.  Since they are few in number, simply dumping them and starting again (with a few others) is the right approach.

minad:
You misunderstood. I don't mean to mutate the old overlays. I meant to change the way the interruption is handled such that it is transactional.

jdtsmith:
I see, sounds sensible.  One way to cement the transaction is to refuse to update last-input unless it runs to completion.  But for this as you say you need to "fallback" to the last overlays you had. So something like:

1. on new changed input, cache the old overlays.
2. create new overlays as before.
3. If this runs to completion, replace overlays with the new ones, and update the last_input.
4. If interrupted, fall back on the old overlays, and leave last_input alone.

minad:
Please take a look at 82b1c543db4df1567f1a828c7aed0e3ef55e09ac. What do you think about it?

jdtsmith:
Oops crossed paths.

minad:
Sorry about the interrupt. I didn't see that you had answered in the meantime. Does it work well for you?

jdtsmith:
> Sorry about the interrupt

I see what you did there ;).   

Yes this tests well.  Although my [version](https://github.com/jdtsmith/consult/commit/53ee32893baa316447373a3b74efe6f709d7e076) of the same was simpler ;).   

If you haven't you should really try it on a long file.  I could see using this instead of grep if I wasn't sure just what I wanted to search for.

Issue #517, opened 2022-02-19
consult-line makes subsequent navigation in large file buffer very slow


bbktsk:
I have a very large file (4214473 lines, ~520MB) open in the Fundamental mode, with emacs 27.2 running in a text mode in a terminal over ssh. Initially, everything works as expected, but once I invoke `consult-line` and quit it, navigation in the buffer suddenly becomes very slow with screen redraws taking several seconds.

Steps to reproduce:
- `emacs -nw -q`
- eval minimal config:
```
(package-initialize)
(require 'consult)
(setq completion-styles '(substring))
 ;; gc setting does not seem to have much impact on the issue, it is included just to indicate I've already tried it
(setq gc-cons-threshold 1000000000) 
```
- `C-x C-f verylargefile.txt` , select "yes" on the confirmation prompt
- confirm that navigation in file is fast: `PgDown`, `PgUp`, `M-<`, `M->` executes  "immediately" (maybe 0.1sec? The speed I'm used to from regular sized files.)
- `M-x consult-line`. Wait ~60 sec for the prompt (that's unfortunate but expected and not the topic of this issue). 
- Do not search for anything, just `C-g` to abort search.
- `PgDown`, `PgUp`, `M-<`, `M->` are now very slow, with screen updates after each taking several seconds. Screen seems to be updated in "batches", each containing about 5 lines and taking ~0.5 sec to complete, including slight yet visible delay between batches.

Running cpu profiler during navigation shows that most of the time is spent in the `redisplay` function.

Using emacs-27.2 on Linux running on 20core/64GB VM. All packages were freshly installed today.

minad:
`consult-line` generates many markers, which are registered in a buffer linked list. It may be that these markers lead to issues here. Needless to say that `consult-line` is not designed to handle such large files (use occur or consult-grep instead). However we can still try to figure out the issue here. Please replace `(point-marker)` with `(point)` and then `M-x compile-defun` to check if the problem goes away:

https://github.com/minad/consult/blob/1d5571f57977067fa9e4ac3794b0cb0fbca5dc6a/consult.el#L2788

If the problem goes indeed away with this change it may make sense to introduce a "cheap mode" where we use integers instead of markers, with the downside that recursive editing will not work and the positions will lose their validity as soon as the file is edited. This matters in particular for Embark export/snapshort. Embark could convert to markers when exporting.

minad:
[...]

PR #519, opened 2022-02-19
Experiment with cheap markers

minad:
Fix for #517

@oantolin What do you think about the change? The complexity of the change is not that high. It may help if one repeatedly executes `consult-line` on larger files. In embark-consult you could use `consult--location-marker` to access the markers.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -912,6 +912,31 @@ When no project is found and MAYBE-PROMPT is non-nil ask the user."
         (lambda (cand) (eq (get-text-property 0 'consult--type cand) consult--narrow))
         :keys types))
 
+(defmacro consult--with-location-upgrade (candidates &rest body)
+  "Upgrade location markers from CANDIDATES on window selection change.
+The markers are not upgraded when BODY has finished without a window change."
+  (declare (indent 1))
+  (let ((hook (make-symbol "hook")))
+    `(let ((,hook (make-symbol "consult--location-upgrade")))
+       (fset ,hook
+             (lambda (_)
+               (unless (eq (selected-window) (active-minibuffer-window))
+                 (remove-hook 'window-selection-change-functions ,hook)
+                 (mapc #'consult--get-location ,candidates))))
+       (unwind-protect
+           (progn
+             (add-hook 'window-selection-change-functions ,hook)
+             ,@body)
+         (remove-hook 'window-selection-change-functions ,hook)))))
+
+(defun consult--get-location (cand)
+  "Return location from CAND."
+  (let ((loc (get-text-property 0 'consult-location cand)))
+    (when (consp (car loc))
+      ;; Transform cheap marker to real marker
+      (setcar loc (set-marker (make-marker) (cdar loc) (caar loc))))
+    loc))
+
 (defun consult--lookup-member (_ candidates cand)
   "Lookup CAND in CANDIDATES list, return original element."
   (car (member cand candidates)))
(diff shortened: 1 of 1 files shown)


minad:
Okay, I cleaned this up a little bit more. To access the location data in Embark, simply use `consult--get-location` instead of `get-text-property`. Then you will get these lazily upgraded markers. I think this is a good change

oantolin:
Yes, it looks good, and fairly simple.

But I think this means that for embark-export and embark-collect I should evaluate `consult--get-location` on each candidate at collect/export time. Waiting until the user presses RET on each candidate is too late as a lot of other editing could happen in between. Now, for export this is fine, I can just do it in the exporter, but for collect I think there isn't currently a hook where I could do this. I need to think a bit how best to do it.

minad:
> But I think this means that for embark-export and embark-collect I should evaluate consult--get-location on each candidate at collect/export time. Waiting until the user presses RET on each candidate is too late as a lot of other editing could happen in between. Now, for export this is fine, I can just do it in the exporter, but for collect I think there isn't currently a hook where I could do this. I need to think a bit how best to do it.

Hmm, note that Consult also tries to upgrade markers in a window selection change hook. Not sure if this is already sufficient for embark collect?

oantolin:
> Not sure if this is already sufficient for embark collect?

I think it probably won't run, because embark-collect/export quit the minibuffer before they display the collect/export buffer, so the unwind-protect probably removes the hook before the hook can run. But I'm not sure, I'll test.

oantolin:
Oh, I just realized there is a good place to upgrade markers during collect: `embark-collect-mode-hook`!

minad:
Okay, I think from the side of Consult this is ready. We should synchronize our efforts a bit to avoid unnecessary breakage for MELPA users. Do you want me to prepare a PR for Embark or do you already have something in the works. You probably know better where to perform the upgrade?

oantolin:
I handle things in embark-consult.

oantolin:
I'm almost done. Embark collect is slightly tricky: I'm trying to only upgrade the markers when necessary, and the case where you use `embark-collect-live` is causing a bit of trouble.

minad:
I thought about this too, but I would say that any special casing is unnecessary. First you won't do worse than we already do. Second live collect will only open up after you typed a few characters or on demand. At this point the set of candidates will already be reduced such that the issue is less pronounced. Then we are already doing better than before. Therefore I would not do anything special and keep thing simple on the side of Embark.

oantolin:
OK, sounds reasonable.

oantolin:
Mmh. Choosing a consult-line candidate from the embark-collect-completions buffer is broken right now, but that's unrelated to this issue (to make sure it's unrelated I tested from my normal configuration without this branch of consult).

minad:
That's good if we also fix some other issues on the go :)

oantolin:
OK, I made a cheap-markers branch of embark. I'll fix that bug separately.

minad:
Okay, great. It looks simple on the side of Embark! I merge this then.

oantolin:
OK, I'll merge too.

oantolin:
Done.

oantolin:
And that bug I noticed is fixed now: oantolin/embark@98bdc0f553d5bf0caa2c5acf828c3a8d99a5f231. It was a simple oversight: Embark collect buffers now remove the tofu and stash the original candidate in a text property, but when I made that change I forgot to update the function called when you press RET on a candidate and it was using the tofuless candidate in the embark-completions case.

oantolin:
I forgot to say: the only reason I didn't notice that earlier is that I switched to Vertico full time!

minad:
MELPA already picked our changes up. It seems we merged just in time! :smile:

oantolin:
That was lucky.

PR #520, opened 2022-02-19
Fix #518

### LINKED ISSUE #518: Consult register actions (M-p, M-w, ...) do not work in terminal
Hi:

Recently I found that Modified keys are not working in consult properly when using emacs form the terminal (-nw).

Commands like `consult-register-save` receive `w` instead of `M-w`, `f` instead of `M-f` and so on. Before I had ivy and it worked properly; and xterm is sending the right bindings because the rest of emacs gets them.  And when I do `M-x C-h k M-w` it shows the help for `M-w`... so this seem to be something specific in Consult??? BTW I use `Vertico` in case you need to know.

Thanks in advance, Ergus

oantolin:
In terminal Emacs, given M-char, (read-key) will return 27 and set last-input-event to char. This change turns that situation into what would have been read in graphical Emacs, letting the rest of the code remain unchanged.

### DIFF
--- a/consult-register.el
+++ b/consult-register.el
@@ -230,6 +230,7 @@ This function is derived from `register-read-with-preview'."
          (prefix (car action-list))
          (action-list (cdr action-list))
          (action (car (nth 0 action-list)))
+         (key)
          (reg)
          (preview
           (lambda ()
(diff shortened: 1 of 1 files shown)


consult-register.el
```diff
@@ -252,10 +253,15 @@ This function is derived from `register-read-with-preview'."
...
-                                             'face 'minibuffer-prompt))
+	  (while (memq (setq key
+                             (read-key (propertize (caddr (assq action action-list))
+                                                   'face 'minibuffer-prompt)))
 		       help-chars)
             (funcall preview))
+          (when (and (eql key ?\e) (characterp last-input-event))
+            ;; in terminal Emacs M-letter is read as two keys, ESC and the letter,
+            ;; turn that into what would have been read in graphical Emacs
+            (setq last-input-event (logior #x8000000 last-input-event)))
```
minad, line 264: I am not comfortable with overwriting a variable which I don't own. Please introduce a separate variable here for the key decoding.

minad:
Thanks for looking into this!

consult-register.el
oantolin: Will do.

oantolin: It looks better with the new variable anyway.

oantolin: I forgot to reindent, wait a second.

oantolin: OK, should be good now.

PR #522, opened 2022-02-19
Fix #521

### LINKED ISSUE #521: `M-.` does not work as preview key (Terminal Emacs)
Hi:

When using `(setq-default consult-preview-key (kbd "M-."))` on emacs terminal (-nw) it just doesn't work. It seems like `M-.` is well captured but emacs, but not  processed by consult.

`C-h b` shows:  `M-.		ignore`

oantolin:
Use lookup-key instead of vector equality to compare keys.


minad:
Thanks!

PR #526, opened 2022-02-26
Set current error when exiting consult-compile-error

astoff:
I haven't tested thoroughly, but this seems to do the trick. But one weird problem remains: the point moves in the compilation buffer as I preview the candidates. I don't understand why, given that this is inside a `save-excursion`.

### DIFF
--- a/consult-compile.el
+++ b/consult-compile.el
@@ -74,7 +74,7 @@
             (overlay-arrow-position overlay-arrow-position))
         (save-window-excursion
           (funcall next-error-function 0)
-          (point-marker))))))
+          (cons (point-marker) marker))))))
 
 (defun consult-compile--compilation-buffers (file)
   "Return a list of compilation buffers relevant to FILE."
@@ -84,6 +84,17 @@
      (with-current-buffer buffer
        (and (compilation-buffer-internal-p)
             (file-in-directory-p file default-directory))))))
+
+(defun consult-compile--jump-state (&optional face)
+  "Like `consult--jump-state', but also sets the current compilation error.
+FACE is the cursor face."
+  (let ((state (consult--jump-state face)))
+    (pcase-lambda (`(,cand . ,error) restore)
+      (when (and error restore)
+        (with-current-buffer (marker-buffer error)
+          (setq compilation-current-error error
+	        overlay-arrow-position error)))
+      (funcall state cand restore))))
 
 ;;;###autoload
 (defun consult-compile-error ()
@@ -109,7 +120,7 @@ preview of the currently selected error."
    :group (consult--type-group consult-compile--narrow)
    :narrow (consult--type-narrow consult-compile--narrow)
    :history '(:input consult-compile--history)
-   :state (consult--jump-state 'consult-preview-error)))
+   :state (consult-compile--jump-state 'consult-preview-error)))
 
 (provide 'consult-compile)
 ;;; consult-compile.el ends here
--- a/consult-compile.el
+++ b/consult-compile.el
@@ -60,8 +60,7 @@
                                     (0 ?i)
                                     (1 ?w)
                                     (_ ?e))
-                   'consult-compile--marker (point-marker)
-                   'consult-compile--loc (compilation--message->loc msg))
+                   'consult-compile--marker (point-marker))
                   candidates))))
       (nreverse candidates))))
 
@@ -69,19 +68,13 @@
   "Lookup marker of CAND by accessing CANDIDATES list."
   (when-let ((cand (car (member cand candidates)))
              (marker (get-text-property 0 'consult-compile--marker cand))
-             (loc (get-text-property 0 'consult-compile--loc cand))
-             (buffer (marker-buffer marker))
-             (default-directory (buffer-local-value 'default-directory buffer)))
-    (consult--position-marker
-     ;; taken from compile.el
-     (apply #'compilation-find-file
-            marker
-            (caar (compilation--loc->file-struct loc))
-            (cadar (compilation--loc->file-struct loc))
-            (compilation--file-struct->formats
-             (compilation--loc->file-struct loc)))
-     (compilation--loc->line loc)
-     (compilation--loc->col loc))))
+             (buffer (marker-buffer marker)))
+    (with-current-buffer buffer
+      (let ((compilation-current-error marker)
+            (overlay-arrow-position overlay-arrow-position))
+        (save-window-excursion
+          (funcall next-error-function 0)
+          (point-marker))))))
 
 (defun consult-compile--compilation-buffers (file)
   "Return a list of compilation buffers relevant to FILE."


consult-compile.el
```diff
@@ -92,6 +85,19 @@
        (and (compilation-buffer-internal-p)
             (file-in-directory-p file default-directory))))))
 
+(defun consult-compile--jump-state (&optional face)
+  "Like `consult--jump-state', but sets the current compilation error.
+The function can be used as the `:state' argument of `consult--read'.
+FACE is the cursor face."
+  (let ((preview (consult--jump-preview face)))
```
minad, line 92: Could you wrap `consult--jump-state` here instead? This would be more natural.

minad:
Thanks! I like the simplification. I wonder why I didn't use this function right away but implemented the marker computation manually.

> But one weird problem remains: the point moves in the compilation buffer as I preview the candidates.

This should be fixed. Besides that, it looks good.

astoff:
Okay, it should be good now.

minad:
Thanks! I merged only the second commit, since the simplification modifies preview in an unintended way. This was also the reason why I used my own error marker lookup function.

See https://github.com/minad/consult/commit/defcdd2ba34411eb6f967dcd2a1df4d3316ba04d and the simplification in https://github.com/minad/consult/commit/5a3f6904178556148d4147ff873b823da290ecd1.

astoff:
I also noticed the extra "pulse highlight". I would rather try to find a way to disable it and call the `next-error-function`. The reason being that the `compilation-next-error-function` computes markers for an entire buffer the first time you visit it, so that the error positions remain correct when you edit the file (which presumably you will do if you are fixing errors!)

It wouldn't be so bad if selecting the error at line 50 always jumped to line 50, even if that info is outdated. But the current behavior of `consult-compile` is a random mix of the naive and the fancy approach.

(A secondary reason to prefer the old approach is that it was compatible with compilation-derived modes that need some custom change to the next-error-function, such as a possible `devdocs-grep` command that produces fake file names.)

astoff:
In fact, getting rid of the highlight is trivial, cf. https://github.com/astoff/consult/commit/4edebcd15f31abbb6250fec4b00fb30278e69b3c

minad:
Where does this computation of the marker positions happen? We could also do that in `consult-compile--error-candidates`, but I think I didn't do this for performance reasons. Making `consult-compile-error` work with non-compile errors which create fake errors is a non-goal here. Using the `next-error-function` had other bad implications, if I recall correctly I used that approach before but moved away from it due to unforeseeable and uncontrolled side effects. I consider `consult-compile-error` border line functionality of Consult anyway. It may be better to remove it from here and package it up separately.

astoff:
> Where does this computation of the marker positions happen?
In `compilation-next-error-function`, after this comment:
```
    ;; If loc contains no marker, no error in that file has been visited.
    ;; If the marker is invalid the buffer has been killed.
    ;; So, recalculate all markers for that file.
```

I'd be very surprised if that loop introduced any slowdown (and in any case the first preview on a given file can be slower for other reasons, e.g. it now being open yet and triggering some autoloading).

>  Making consult-compile-error work with non-compile/fake errors is a non-goal here.

That's fine and good, but if everybody uses the intended "APIs" (which in this case unfortunately are function that jump around instead of returning values, but alas), then everything will be compatible by default.

> It may be better to remove it from here and package it up separately.

I don't know, I use this command sometimes but wouldn't bother to install a package just for it.

minad:
> That's fine and good, but if everybody uses the intended "APIs" (which in this case unfortunately are function that jump around instead of returning values, but alas), then everything will be compatible by default.

Of course. But two things: First, the next-error APIs are badly designed and problematic to reuse. These APIs not supposed to be used like this. Second, I am using the compile error buffer API, not the `next-error-function` API, which sits on top of that. There was a version of `consult-compile-error`, called `consult-error` which relied only on the next-error APIs. It didn't work.

You certainly have a point that it would be better to precompute the markers. I will give that a try.

astoff:
> Second, I am using the compile error buffer API, not the next-error-function API, which sits on top of that.

I don't quite understand: `next-error-function` is a variable holding the symbol `compilation-next-error-function` in `compilation-mode` buffers (or, potentially, a remix thereof in derived modes). But if you want to call `compilation-next-error-function` directly, that's not a big deal.

> You certainly have a point that it would be better to precompute the markers. I will give that a try.

Module the refactoring you added on top of my code, this is exactly what is done in https://github.com/astoff/consult/commit/4edebcd15f31abbb6250fec4b00fb30278e69b3c.

I agree the next-error API is not nice, but it works just fine here. If there are performance issues, it's because of the implementation, not the fact that the save-excursion-and-return a marker has to be done on the Consult side.

minad:
I precompute the markers early now, this should fix the relocation issues.

> I don't quite understand: next-error-function is nothing but a variable holding the symbol compilation-next-error-function in compilation-mode buffers (or, potentially, a remix thereof in derived modes). But if you want to call compilation-next-error-function directly, that's not a big deal.

Sure. My argument was targeted at cases where the `next-error-function` is changed, e.g., the devdocs buffer. This is not supported. `consult-compile-error` is supposed to work only in compilation buffers. Therefore it is fine to use the compilation message and compilation location data structures directly as Consult does it currently. It is okay to neither call next-error-function nor compilation-next-error-function, but to use the APIs directly.

astoff:
In this last commit, aren't you visiting all the files that appear in the error list eagerly at candidate-computation time?

minad:
Hmm, right. That's the reason why I didn't do that before.

astoff:
Again, I bring up the devdocs-grep example more from a philosophical. If there is one approach that composes well with other things, it's certainly better than an approach that needs to dig deeper into the internals of another package. (My approach doesn't use any double-dashed functions or variables from Emacs, by the way.)

minad:
Yes, I agree with your overall perspective. But your argument nevertheless does not apply here, since I don't want to use the `next-error-function` or the `compilation-next-error-function`. I want to use the API on a lower level, such that I don't get the bad side effects of `compilation-next-error-function`, jumping around, pulsing, whatever. All I want is an API which converts a compilation error to a marker. This part of the code should ideally be extracted from the `compilation-next-error-function`. Since such an API does not exist, I had to write that part myself - which I did, but it seems badly, since it doesn't perform the early marker computations for all errors in a file etc.

astoff:
Okay, fine. Just to make sure you got that part — the highlight can be disabled by setting `next-error-highlight` to nil (documented behavior).

> Your approach works around side effects of a function to extract markers, which is just as bad.

It stands in my defense the fact that the jumping around it is the _documented_ behavior of the function. :stuck_out_tongue_closed_eyes:

minad:
On the other hand `compilation-find-file` also has bad side effects, asking the user, popping up buffers, so maybe my argument just doesn't hold up in practice and using `compilation-next-error-function` with worked around side effects is practically not worse. Hmm.

minad:
Well, I think it should be okay now. The code matches mostly what you proposed in your PR. Please check if it works or if you have more complaints. :-P

astoff:
All right, looks good to me, thanks!

Issue #239, opened 2021-03-01
consult buffer preview loses focus in exwm after previewing X windows


codygman:
It's probably the case that the exwm managed X window steals focus actually. 

The syntax might be wrong, but I think a workaround would be:

```
(advice-add 'exwm-reset :after 'consult-buffer-preview)
```

Except you'd want to only call exwm-reset for X buffers and not normal emacs buffers. This is imperfect, but hitting my hotkey for exwm-reset is how I get my focus out of the x window back into the buffer switch minibuffer.

minad:
Duplicate of #178 and #204. Please also take a look at https://github.com/minad/consult/wiki#do-not-preview-exwm-windows-or-tramp-buffers. Maybe you want to add your workaround to the wiki too, such that `exwm-reset` is called for X buffers. Maybe @thomasheartman would be interested in this too and could be of help?

Besides that I cannot do anything on the side of Consult. I think the issue should actually be fixed by EXWM (https://github.com/ch11ng/exwm/issues/813). Note that I am not an EXWM user myself and I cannot give full support when EXWM is combined with Consult.

codygman:
> Duplicate of #178 and #204. Please also take a look at https://github.com/minad/consult/wiki#do-not-preview-exwm-windows-or-tramp-buffers. Maybe you want to add your workaround to the wiki too, such that `exwm-reset` is called for X buffers. Maybe @thomasheartman would be interested in this too and could be of help?
> 
> Besides that I cannot do anything on the side of Consult. I think the issue should actually be fixed by EXWM ([ch11ng/exwm#813](https://github.com/ch11ng/exwm/issues/813)). Note that I am not an EXWM user myself and I cannot give full support when EXWM is combined with Consult.

Totally understandable you can't support both. 

My fault for not:

1) looking for existing issues and 
2) not consulting the wiki. 

It's kind of rare to have such a great wiki though... good job there!

Thanks for pointing me in the right direction.

minad:
No worries, it is okay to open such issues. I only don't want to rediscuss the issue again, since there has already been a very long discussion in #178 and #204. The solution we currently have is not perfect, but a step in the right direction. As soon as we get more information from the EXWM maintainers we can adjust accordingly. I still consider EXWM a fringe setup. On my system I would have great difficulties installing it since I am using i3 on qubes, which is a sandboxed, virtualized environment. I probably would also be unhappy with Emacs hanging from time to time. But I certainly like the idea!

[...]

Issue #204, opened 2021-02-04
EXWM: X window preview issues


thomasheartman:
When using EXWM and Consult's buffer preview functionality with X windows across multiple frames/monitors, the buffers aren't properly restored after switching buffers. 

This is similar to [this issue](https://github.com/minad/consult/issues/178), but doesn't seem to be the same. It is also independent of emacs-mini-frame.

While the window layout restoration works well if the buffer with the preview and the X window were on the same frame, the restoration doesn't work properly when they're on different frames, and the X window ends up being buried if not selected. For other buffers it seems to work as expected (but then again, other buffers don't have the same limitation as X buffers).

There's also an issue with the preview functionality for X buffers, where the X window 'steals focus' from the minibuffer. When this happens, the next and previous key bindings for consult don't work anymore (because the X window has the focus), but the minibuffer is still open/active. I'll see if I can reproduce this reliably and get back to you (I'm a bit short on time just now; sorry), but wanted to raise it in case it's already a known issue.

Thanks for your time and attention! Let me know if you need anything else and I'll try and respond as quickly as I can. Cheers!

minad:
This is mostly a duplicate of #178. If I understand this correctly, EXWM does not allow duplicating X buffers. This is different from how Emacs buffers usually behave. Basically this is an EXWM bug. The focus stealing is also an EXWM bug. 

However we have a solution now, since I added configurable buffer sources since then. Basically you have to replace `consult--source-buffer` (and the other problematic buffer sources) with buffer sources which filter out the X buffers or do not perform preview for them. Alternatively you can directly modify the `consult--source-buffer` and the other problematic buffer sources.

This is the current source definition:

~~~ elisp
(defvar consult--source-buffer
  `(:name     "Buffer"
    :narrow   ?b
    :category buffer
    :face     consult-buffer
    :history  buffer-name-history
    :state    ,#'consult--buffer-state
    :items
    ,(lambda ()
       (let ((filter (consult--regexp-filter consult-buffer-filter)))
         (seq-remove (lambda (x) (string-match-p filter x))
                     (consult--cached-buffer-names)))))
  "Buffer candidate source for `consult-buffer'.")
~~~

Modify it like this:

~~~ elisp
(defvar consult--source-x-buffers
  `(:name     "Buffer"
    :narrow   ?b
    :category buffer
    :face     consult-buffer
    :history  buffer-name-history
[...]

Issue #178, opened 2021-01-23
EXWM/Miniframe: Windows not restored after preview


Stebalien:
When consult is used in conjunction with selectrum and mini-frame, the window configuration is not restored after previewing then aborting.

For example,

1. Enable `mini-frame-mode`, `selectrum-mode`.
2. Run `consult-buffer`, then preview a different buffer.
3. Hit `C-g` to quit.

Expectation: I expected the buffer I was editing _before_ running `consult-buffer` to be selected and focused.

Reality: The last previewed buffer remains selected and focused, nothing is restored.

Wrapping the command in `save-window-excursion` as in https://github.com/minad/consult/pull/177 appears to fix the issue, but I'm unsure why this is only an issue with `mini-frame-mode` and not an issue otherwise.

Stebalien:
Note: everything _is_ restored if the focused buffer is not changed (e.g., with `consult-lines`). This is only an issue with commands like `consult-ripgrep` and `consult-buffer`.

minad:
Okay, I think in order to be save I should use set-buffer in `consult--preview-position`. Originally I used `save-excursion` or even `save-window-excursion` but stopped using it since I wanted a more fine grained solution.

Stebalien:
From what I can tell, `consult--preview-position` isn't used for `consult--buffer`, `consult--buffer` is (which doesn't do anything on restore.

I've been able to get _this_ to work with:

```diff
diff --git a/consult.el b/consult.el
index 9920f68..9dd1892 100644
--- a/consult.el
+++ b/consult.el
@@ -2674,6 +2674,7 @@ FACE is the face for the candidate."
 
 Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be used to display the item."
   (let* ((curr-buf (current-buffer))
+	 (curr-win (get-buffer-window curr-buf))
          (all-bufs (append (delq curr-buf (buffer-list)) (list curr-buf)))
          (buf-file-hash (let ((ht (make-hash-table :test #'equal :size (length all-bufs))))
                           (dolist (buf all-bufs ht)
@@ -2756,8 +2757,11 @@ Depending on the selected item OPEN-BUFFER, OPEN-FILE or OPEN-BOOKMARK will be u
            :preview
            (lambda (cand restore)
              (cond
-              (restore)
-              ;; In order to avoid slowness and unnecessary complexity, we
+              (restore
+	       (if (not (buffer-live-p curr-buf))
[...]

PR #537, opened 2022-03-14
Fix for EXWM buffers stealing focus during preview

chasecaleb:
Good news! I found a way to fix the EXWM focus stealing issue during buffer previews.I've tested this out pretty heavily over the last hour or so and it works reliably for me. 

For context, see:
- https://github.com/minad/consult/issues/178
- https://github.com/minad/consult/issues/204
- https://github.com/minad/consult/issues/239
- https://github.com/ch11ng/exwm/issues/813

---
 
As far as the implementation: I considered wrapping this in a conditional to only trigger for EXWM buffers, but I didn't see the point in adding complexity since this is effectively a no-op for other buffer types. I didn't notice a performance hit either, but I'm running on a pretty fast system. I'm happy to make changes if you have feedback.

I realize this is an EXWM bug and not a Consult issue, but unfortunately for the couple years EXWM has received very little love so I'm not holding my breath for a fix on their side. On the bright side, this workaround is so trivial it doesn't seem too egregious to add to Consult.

@minad How do you feel about this?

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -4032,7 +4032,11 @@ Report progress and return a list of the results"
 (defun consult--buffer-action (buffer &optional norecord)
   "Switch to BUFFER via `consult--buffer-display' function.
 If NORECORD is non-nil, do not record the buffer switch in the buffer list."
-  (funcall consult--buffer-display buffer norecord))
+  (prog1
+      (funcall consult--buffer-display buffer norecord)
+    ;; Stop EXWM buffers from hijacking focus. See https://github.com/minad/consult/issues/178.
+    (when-let ((mini (active-minibuffer-window)))
+      (select-window (active-minibuffer-window)))))
 
 (consult--define-state buffer)
 


minad:
Thank you for looking into this! I don't want to add such workarounds to Consult. It is better to fix the issue at the root, in this case in EXWM. I've heard that EXWM got a new maintainer? I suggest to use your workaround in an `:after` advice in the meantime. Furthermore I would appreciate if you add the code snippet to the Consult wiki.

chasecaleb:
Understandable, thanks for the quick reply. I'll add a user-facing snippet to the wiki tomorrow, do you want me to delete the current snippet on it that shows how to disable previews for EXWM? I don't think there's a point to keeping it since my kludge works consistently without issue, and there's still the nearly identical tramp snippet showing how to disable previews. It's your wiki though, I don't care either way.

You're right EXWM has a new maintainer, although things are still pretty quiet over there. Hopefully I'll be able to work with them and get the issue figured out though. Not your problem though (lucky you) 😄.

minad:
Please keep the old snippet and add your code next to it with some explanation.

chasecaleb:
Done.

**For any other EXWM users** who come across this:
1. See the advice version of this PR that I added [to the wiki](https://github.com/minad/consult/wiki#do-not-preview-exwm-windows-or-tramp-buffers).
2. Follow (and perhaps contribute to) the upstream EXWM issue: https://github.com/ch11ng/exwm/issues/813

And here's the advice snippet from the wiki, for the sake of posterity:
```emacs-lisp
(defun consult-exwm-preview-fix (&rest _args)
  "Kludge to stop EXWM buffers from stealing focus during Consult previews."
  (when-let ((mini (active-minibuffer-window)))
    (select-window (active-minibuffer-window))))

(advice-add
    #'consult--buffer-action :after #'consult-exwm-preview-fix)
```

minad:
I've added a customization variable `consult-preview-excluded-buffers` in https://github.com/minad/consult/commit/e1621f4e4be6c7078373c74346311e640e16338f. This variable can be used to exclude EXWM buffers from preview.

Issue #538, opened 2022-03-15
Configurable (symbol . value) bindings for file preview


jdtsmith:
I have enabled inline images on load in org-mode, which is slow when previewing image-rich org files in consult-buffer.  Right now my fix is to inhibit org startup during file preview, ala:

```elisp
(defun my/consult-no-org-preview (fun)
  (lambda (&rest r)
    (let ((org-inhibit-startup t))
      (apply fun r))))
(advice-add #'consult--temporary-files :filter-return #'my/consult-no-org-preview)
```

This also prevents `org-indent-mode` from doing its thing.  But this makes me think a user config like `consult-preview-file-variables` might be useful.  It would be an alist of symbol names and values, like `'((org-inhibit-startup . t))`, which could be let-bound in the closure returned by `consult--temporary-files` using `cl-progv` or similar chicanery.

astoff:
I just noticed that file previews will start Eglot, which is also slow. Maybe setting `delay-mode-hooks` would help for this case. But this is some shady chicanery indeed. (I don't think file previews should be on by default, TBH).

jdtsmith:
I think for that you can `(add-to-list 'consult-preview-excluded-hooks 'eglot--maybe-activate-editing-mode)`.

astoff:
I wasn't aware of this option. Well, if something that's not a built-in in ever makes it into the default value of that variable, I guess Eglot is a good candidate.

(Also, to qualify my opinion that file previews should maybe be off by default, I'm specifically talking about `consult-buffer` here. `consult-grep` is different because in that case the live preview is very much a fundamental part of the functionality, while for buffer switching it's just a nifty extra.)

jdtsmith:
Actually looking at the code, I'm not sure this (or it's lsp-mode equivalent) is actually doing anything, other than removing such hooks from `find-file-hooks` (where they don't appear). I was quite sure that lsp-mode was firing before during preview; maybe @minad fixed this another way. 

I personally like the previews with consult-buffer too.  I just want to "lightweight them" by turning off the especially heavy initialization bits (like contacting an LSP server or loading a bunch of images!).

minad:
See also #515. Maybe the excluded hooks variable should be consulted for more hooks?

jdtsmith:
I don't think that will help as much as could be hoped, because much of the initialization of a mode is done in the body of its mode function, _not_ in its mode hooks.  In things like lsp, much of the initialization is done in response to server contact, via the mode that starts the language server.

[...]

PR #542, opened 2022-03-22
Delay mode hooks when previewing files

### LINKED ISSUE #515: Defer loading major mode, or exclude certain modes on preview
Hi there, dropping a quick issue as I was thinking about a problem that I usually have when using the live-preview feature. Normally I have it off, as I believe sometimes it slows down my Emacs considerably.

An example of this is when previewing a file which has a major mode that is controlled by lsp-mode. The live-preview loads the major mode and lsp-mode, which causes the language server initialization. For heavy-handed servers, this process is slow, and I find myself waiting for the server to initialize, even if I wasn't planning on opening that file.

I have a couple suggestions:
- The problem that I see with deferring the major mode is that the preview will have no syntax highlighting.
- Maybe there could be a variable for major modes that should be taken care of differently (for previews), even if the candidate list is the same. This would probably mean having control over the candidates individually, but I don't find this to be a problem as we are already doing that when building the candidate list.

I created this issue to start a discussion, what do you think? Have you already sunk some time thinking about this topic?

astoff:
Fixes #515.

This will skip running all mode hooks, including say the activation of a language server or starting a Flyspell process. If desired, we could run some of the `delayed-mode-hooks` based on a whitelist or blacklist. But I don't think this is actually necessary.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -1124,7 +1124,8 @@ MARKER is the cursor position."
                       (prog1 nil
                         (message "File `%s' (%s) is too large for preview"
                                  name (file-size-human-readable size)))
-                 (cl-letf* (((default-value 'find-file-hook)
+                 (cl-letf* (((default-value 'delay-mode-hooks) t)
+                            ((default-value 'find-file-hook)
                              (seq-remove (lambda (x)
                                            (memq x consult-preview-excluded-hooks))
                                          (default-value 'find-file-hook)))


minad:
I am unsure. This feels like a sledgehammer method. Note that previewed buffers will stick around if you modify them. These buffers are then in a half initialized state. Wouldn't it be better to filter out only certain problematic hooks? Another possibility could be to finish the delayed initialization here by invoking `run-mode-hooks`:

https://github.com/minad/consult/blob/47c4f405efdf4692c6b7e1dd2098573db9aeae6c/consult.el#L1142

(We could also discuss to just remove file preview and other problematic previews altogether and only use explicit Embark non quitting actions for "preview". There seem to be too many issues around this preview feature. It may be more reasonable to only preview things which are already available (buffers, kill ring entries, etc.) but nothing else.)

astoff:
> Note that previewed buffers will stick around if you modify them. These buffers are then in a half initialized state.

Don't they already end up in a half-initialized state, since you skip the dir-local variables and some find-file-hooks?

How about making the preview buffers read-only, and recklessly killing them in the end?

> Lsp certainly is a problem but these cases could be configured one by one.

Then in any case you have to delay the mode hooks, then look at `delayed-mode-hooks` and pick whatever you want from there.
Personally, I'm happy to sledgehammer the problem away, but maybe some people will want to configure things.

> We could also discuss to just remove file preview

I guess the problem is not the file preview, but calling a major mode just for the sake of coloring some words? I think the preview is genuinely useful.

minad:
Hmm. I have to think about this more. Not sure if you saw the addition of `consult--with-location-upgrade`?

https://github.com/minad/consult/blob/47c4f405efdf4692c6b7e1dd2098573db9aeae6c/consult.el#L917-L932

Maybe an approach like this could work too. We start with buffer where everything is disabled. But as soon as the user moves out of the minibuffer we kill all previewed, unmodified and invisible buffers and finish the delayed initialization of the visible buffers.

astoff:
Well, I don't know what you think of the idea of #538, but that makes it totally impossible to finish a delayed initialization. (And how do you finish up adding the dir-locals and the skipped find-file-hooks?)

I'm more inclined to declare that you can't edit a _file_ preview (_buffer_ preview is fine), so make it read-only.

minad:
> I'm more inclined to declare that you can't edit a file preview (buffer preview is fine), so make it read-only.

I want to preserve recursive editing.

> Well, I don't know what you think of the idea of https://github.com/minad/consult/issues/538, but that makes it totally impossible to finish a delayed initialization. (And how do you finish up adding the dir-locals and the skipped find-file-hooks?)

I have yet to understand what the essence is of #538. Finishing dir-locals and skipped hooks are indeed an issue. A possible solution could be to kill the previewed (read-only/delayed) buffer and reopen it when moving to recursive editing. Then we could just skip everything during preview and still get a perfectly initialized buffer as soon as we edit recursively.

astoff:
> A possible solution could be to kill the previewed (read-only/delayed) buffer and reopen it when moving to recursive editing.

Yes, this sounds like a feasible approach.

astoff:
By the way, in the snippet at https://github.com/minad/consult/pull/542#issuecomment-1075611465, why do you create a symbol and fset to it? `add-hook` and `remove-hook` can handle lambdas just fine. (Of course I wouldn't add a lambda to permanently to some hook, I'm only talking about this recurse edit kind of situation.)

minad:
> add-hook and remove-hook can handle lambdas just fine.

Oh they don't. First there is a debugging issue if these hook lists are blown up with large closures. But more significantly there was a serious issue where `add-hook` and `remove-hook` led to expensive equality checks if hook priorities are used (affects Emacs 27 and older). Note that `add-hook`/`remove-hook` use `equal` to check hooks for equality. This problem occurs only if you use large closures - Consult produces very large closures in `consult--read`. Unfortunately this even affects `minibuffer-with-setup-hook` with `:append`:

https://github.com/minad/consult/blob/473e6585c516d0e7fd4c256c333713fb40e9947a/consult.el#L703-L723

Furthermore there was a memory leak issue if local hooks with priorities are used (affects both Emacs 27 and 28). For this reason, local hooks with priorities should never be used, but there are also not many reasons to use local hooks with priorities. However see here for a use case in Consult:

https://github.com/minad/consult/blob/473e6585c516d0e7fd4c256c333713fb40e9947a/consult.el#L1341-L1348

Anyway it took quite some time to figure out these issues, mitigate them and fix them upstream. Therefore I ended up with using only symbols for hooks in Consult. In some of the cases I could change the hooks back to use lambdas.

minad:
To be clear: The expensive closure equality check should be fixed in 28. This means we can use lambdas as global hooks with or without priorities. Local hooks with priorities are still broken on 28 and should not be used - not even with symbols, since this will still lead to a memory leak. These are fixed on 29. So it seems that finally in Emacs 29 the basic `add-hook` and `remove-hook` APIs do the right thing for all arguments. :laughing:

astoff:
That very good to know, since I'm adding a lambda to a hook in the lazy-highlight for query-replace which is under discussion in the bug tracker.

Another fun fact, hashing of byte-compiled closures is broken in Emacs ≤ 27. Hence this:

https://github.com/astoff/devdocs.el/blob/beaf3584e18e9359181b9a9cdc57cfce0d7724e7/devdocs.el#L106-L124

minad:
I think in upstream it doesn't matter since you can rely on the fixed APIs. 

(Unrelated but I think I will drop Emacs 26 support from my packages now, so see if people are still using that. It should be a big issue since Vertico also requires Emacs 27. Besides these hook issues and some other smaller issues, Emacs 27 was mostly robust for me so I think I will support 27 for longer. But bugs in basic APIs are really the worst :-P).

PR #546, opened 2022-04-07
Introduce new preview function protocol: setup, preview, exit, action

minad:


minad:
I apologize for the breakage! The API change was necessary to implement preview
unwinding properly for buffers. The new preview protocol is generally useful
since it gives the preview function more control. I considered implementing this
in a backward compatible way but then we would have ended up with many
duplicated preview functions or duplicated logic.

This API change affects these MELPA projects which make use of preview:

* consult-eglot and consult-yasnippet by @mohkale
* consult-lsp by @gagbo
* consult-notmuch by @jaor
* consult-org-roam by @jgru

oantolin:
Do you think it would be a good idea for embark-consult to add pre and post actions hooks that disable and reenable the preview for every action? I've been thinking about that for a while now, but the old exposed preview function didn't let me do that if I recall correctly.

minad:
I am not sure I understand https://github.com/minad/consult/commit/39118aa99a392c4654c1858c1585ffb28817f0e4 correctly. When I preview a bookmark, switch to the buffer, move the point around, go back to the minibuffer press `C-. s` then I can update the position.

oantolin:
We can discuss that issue over there to keep this place free for the people you pinged about their libraries to ask questions.

jgru:
Thanks for informing me about the breaking change, @minad. 
I modified `consult-org-roam`'s preview-function in [1fc314c101cb4e9455c8aa294d587873507b6a5e](https://github.com/jgru/consult-org-roam/commit/1fc314c101cb4e9455c8aa294d587873507b6a5e) to conform with `consult`'s API. 

Best regards, 
jgru

jaor:
thanks @minad.  i haven't looked with great care at the new protocol, so
maybe i'm missing something, but a naive attempt replacing the previous
state function:

```elisp
(defun consult-notmuch--preview (candidate _restore)
  "Open resulting CANDIDATE in ‘notmuch-show’ view, in a preview buffer."
  (consult-notmuch--close-preview)
  (when-let ((thread-id (consult-notmuch--thread-id candidate)))
    (notmuch-show thread-id nil nil nil consult-notmuch--buffer-name)))
```

with

```elisp
(defun consult-notmuch--preview (action cand)
  (cond ((eq action 'exit) (funcall (consult--temporary-files)))
        ((eq action 'preview)
         (when-let ((thread-id (consult-notmuch--thread-id cand)))
           (notmuch-show thread-id nil nil nil
                         consult-notmuch--buffer-name)))))
```

is not working for me: the state function is called with 'init and
'preview for the first selected candidate, but when i move from one
candidate to another in the completions buffer (i'm using mct), the
state function is not called again... most probably something related to
mct, or to the fact that i haven't restarted emacs... i will,
eventually.  i just wanted to ask if the new version looks correct to
you?

thanks,
jao
-- 
One will rarely err if extreme actions be ascribed to vanity, ordinary
actions to habit, and mean actions to fear. -Friedrich Nietzsche,
philosopher (1844-1900)

minad:
@jaor There was an issue. Should be fixed again in https://github.com/minad/consult/commit/e926d593d3e09d6bd1ffdf2dc55aec4c6f42abeb.

minad:
Regarding the correctness of your preview function:

* `(funcall (consult--temporary-files))` does not look right. This should be a nop.
* I suggest to take a look at `consult--file-preview` and `consult--buffer-preview` for inspiration.

jaor:
ah, on second read i think you meant this should be enough?

(defun consult-notmuch--preview (action cand)
  (when (eq action 'preview)
    (when-let ((thread-id (consult-notmuch--thread-id cand)))
      (notmuch-show thread-id nil nil nil consult-notmuch--buffer-name))))

it seems that, with the new protocol, i only need to care of the preview
action...

minad:
> it seems that, with the new protocol, i only need to care of the preview
action...

@jaor Yes if you don't need additional setup/cleanup.

jaor:
On Fri, Apr 08 2022, Daniel Mendler wrote:

>  it seems that, with the new protocol, i only need to care of the preview
>  action...
>
> @jaor Yes if you don't need additional setup/cleanup.

excellent, thanks for your help.

mohkale:
Hi. I've updated `consult-eglot` and `consult-yasnippet`. Thanks for the heads up :-).

jaor:
On Fri, Apr 08 2022, Daniel Mendler wrote:

> Regarding the correctness of your preview function:
>
> * (funcall (consult--temporary-files)) does not look right. This should be a nop.

i was looking at this:

(defun consult--file-preview ()
  "Create preview function for files."
  (let ((open (consult--temporary-files))
        (preview (consult--buffer-preview)))
    (lambda (action cand)
      (when (eq action 'exit)
        (funcall open))

so i don't follow, sorry: what shoud be a nop?  do you mean that, when
called within an active consult--read, (consult--temporary-files)
returns a no-op?

for what is worth, it seems to be working just fine with your new
consult.el...

PR #583, opened 2022-05-25
Fix warnings about single quotes in docstrings

dr-scsi:
* consult.el (consult-preview-key, consult-bookmark-narrow)
(consult--location-candidate, consult--with-preview)
(consult--async-sink, consult--read, consult--prompt): Fix
warnings about wrong usage of unescaped single quotes in
docstrings issued by Emacs 29.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -275,7 +275,7 @@ The dynamically computed arguments are appended."
   :type 'string)
 
 (defcustom consult-preview-key 'any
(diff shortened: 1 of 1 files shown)


dr-scsi:
This is a change to suppress warnings with the current Emacs from master (822350d198 in my case).

minad:
This is ugly but thanks :-P

PR #584, opened 2022-05-27
Remove a problematic single quote in docstring

tarsius:
The byte-compiler recently got more fussy about quotes.

### DIFF
--- a/consult-register.el
+++ b/consult-register.el
@@ -40,7 +40,7 @@
     (?f . "File")
     (?w . "Window"))
   "Register type names.
-Each element of the list must have the form '(char . name).")
+Each element of the list must have the form (char . name).")
 
 (cl-defun consult-register--format-value (val)
   "Format generic register VAL as string."


minad:
I think they can be escaped?

minad:
Fixed in https://github.com/minad/consult/commit/21ccad0e75db4098befc592168002b03f2d663ec. Is there a guideline if we should have a quote there or not?

tarsius:
> I think they can be escaped?

Yes. I didn't do that here because I always just write, e.g., `elements have to form (KEY . VAL)`. `'(KEY . VAL)` makes less sense to me because one could also write `(cons KEY VAL)`. Or in other words I show the form the value is supposed to have after evaluation, not how to write code that results in that when it is evaluated.

I didn't know how you do it, so I did it the way that makes more sense to me.

> Is there a guideline if we should have a quote there or not?

I am not aware of it, but generally speaking single quotes don't have to be quoted where the English language requires them but not where Lisp code requires it. `'these quotes'` need quoting though, but in such cases I use `"double quotes"` anyway.

```
`git status'
```

Also isn't allowed, due to the space. I had to fix that in a few places in magit; switching to double quotes.

tarsius:
Related: https://www.gnu.org/software/emacs/manual/html_node/emacs/Quotation-Marks.html.

minad:
Hmm but for example `(cons 'key 'value)` requires quotes for symbols. In this case we have characters and strings so they are self evaluating. Also I am never sure about symbols in docstrings. Pass 'foo as ARG when you want food.

tarsius:
> Hmm but for example `(cons 'key 'value)` requires quotes for symbols. In this case we have characters and strings so they are self evaluating and it is less of an issue.

I wasn't trying to suggest you should put `(cons 'key 'value)` in the docstring, on the contrary:

```emacs-lisp
(setq variable HERE) => RESULT
```

You may put different things HERE to produce RESULT. What I am trying to say is that you should put RESULT in the docstring, instead of one of the possible ways of writing HERE.

> I am never sure about symbols in docstrings, e.g., pass 'foo as ARG when you want foo.

I always use

```
`symbol'
```

PR #589, opened 2022-06-03
Allow consult-flymake to work across all buffers in a project

mohkale:
Now when `consult-flymake` is given a prefix argument it will source diagnostics from all buffers in the current project instead of simply the current buffer. When not in a project then consult-flymake will fallback to diagnostics just from the current buffer.

### DIFF
--- a/consult-flymake.el
+++ b/consult-flymake.el
@@ -33,15 +33,10 @@
     (?w . "Warning")
     (?n . "Note")))
 
-(defun consult-flymake--candidates ()
-  "Return Flymake errors as alist."
-  (consult--forbid-minibuffer)
-  (let* ((raw-diags (or (flymake-diagnostics)
-                        (user-error "No flymake errors (Status: %s)"
-                                    (if (seq-difference (flymake-running-backends)
-                                                        (flymake-reporting-backends))
-                                        'running 'finished))))
-         (diags
+(defun consult-flymake--candidates (diags)
+  "Return Flymake errors from DIAGS as formatted candidates.
+DIAGS should be a list of diagnostics as returned from `flymake-diagnostics'."
+  (let* ((diags
           (mapcar
            (lambda (diag)
              (let ((buffer (flymake-diagnostic-buffer diag))
@@ -60,7 +55,7 @@
                               ('flymake-error ?e)
                               ('flymake-warning ?w)
                               (_ ?n))))))))
-           raw-diags))
+           diags))
          (buffer-width (apply #'max (mapcar (lambda (x) (length (nth 0 x))) diags)))
          (line-width (apply #'max (mapcar (lambda (x) (length (number-to-string (nth 1 x)))) diags)))
          (fmt (format "%%-%ds %%-%dd %%-7s %%s" buffer-width line-width)))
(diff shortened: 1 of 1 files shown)


consult-flymake.el
```diff
@@ -75,17 +80,27 @@
...
      (sort diags
-           (pcase-lambda (`(_ _ ,t1 _ ,m1 _) `(_ _ ,t2 _ ,m2 _))
+           (pcase-lambda (`(,b1 _ ,t1 _ ,m1 _) `(,b2 _ ,t2 _ ,m2 _))
              (let ((s1 (flymake--severity t1))
                    (s2 (flymake--severity t2)))
-               (or (> s1 s2) (and (= s1 s2) (< m1 m2)))))))))
+               (or (> s1 s2)
+                   (string-lessp b1 b2)
+                   (and (= s1 s2)
+                        (< m1 m2)))))))))
```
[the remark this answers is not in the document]
mohkale: Update: Never mind. This doesn't work. :cry:. I'll test a few other combinations and see what works.

consult-flymake.el
```diff
@@ -27,16 +27,21 @@
 
 (require 'consult)
 (require 'flymake)
+(require 'project)
```
minad, line 30: Shouldn't be necessary? flymake requires project?

consult-flymake.el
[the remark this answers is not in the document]
mohkale: Couldn't figure out how to make it work in a single sort, but works with 2 sorts reasonably well.

consult-flymake.el
```diff
@@ -74,18 +69,36 @@
...
-               (or (> s1 s2) (and (= s1 s2) (< m1 m2)))))))))
+     ;; Sort first by position and then by the name of the buffer.
+     (sort
+      (sort diags
+            (pcase-lambda (`(,b1 _ ,t1 _ ,m1 _) `(,b2 _ ,t2 _ ,m2 _))
+              (let ((s1 (flymake--severity t1))
+                    (s2 (flymake--severity t2)))
+                (or (> s1 s2) (and (= s1 s2) (< m1 m2))))))
+      (pcase-lambda (`(,b1 _ _ _ _ _) `(,b2 _ _ _ _ _))
+        (string-lessp b1 b2))))))
```
minad, line 80: I am not fond of sorting twice. Please change the predicate instead.

consult-flymake.el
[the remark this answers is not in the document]
minad: :thinking:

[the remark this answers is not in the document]
minad: Just fyi, `(or (> s1 s2) (and (= s1 s2) (< m1 m2))))` is the criterion which sorts first by s and then by m....

consult-flymake.el
minad: See my comment https://github.com/minad/consult/pull/589#discussion_r889033252. It is an exercise for you. I insist on this :)

mohkale: Okay, got it in one try :tada:, ISTG I actually tried this out but then some entries with a given mark were sorted before others even though order wise they shouldn't have been. I'm guessing I just misread it (I was testing with buffer names that were too similair to each other so they were easy to mistake).

mohkale:
Working like a charm :-).

Screenshot_20220603_161849

consult-flymake.el
```diff
@@ -70,14 +70,16 @@ RAW-DIAGS should be a list of diagnostics as returned from `flymake-diagnostics'
...
+     (sort diags
+           (pcase-lambda (`(,b1 _ ,t1 _ ,m1 _) `(,b2 _ ,t2 _ ,m2 _))
+             (let ((s1 (flymake--severity t1))
+                   (s2 (flymake--severity t2)))
+               (or
+                (> s1 s2)
+                (string-greaterp b1 b2)
+                (and (= s1 s2)
+                     (string-lessp b1 b2)
+                     (< m1 m2)))))))))
```
minad, line 82: This doesn't look correct.

minad: What do you want to achieve?

Sort first by buffer name, then severity, then position?

minad: This is still wrong. I will check tomorrow. Please think about this more thoroughly.

mohkale:
> Btw, @mohkale what about the FSF copyright assignment? Do you have that?

Yep :-).

minad:
Thanks!

mohkale:
Oh, glad I got it right :sweat_smile:. That took a few attempts but nice to learn something new. I'm still not used to boolean based sorting functions but that was a nice challenge. Thanks :-).

PR #626, opened 2022-08-11
Add consult-org-contacts function

tfree87:
This is a small extension to your wonderful consult-org package.

This update adds a function `consult-org-contacts` that will search contacts for those who use the org-contacts.el library. Similar to org-agenda, org-contacts is a list of files used to store contact information with each contact being an org heading. This means we can search the contacts similar to the function `consult-org-agenda`.

The function checks to see if the library `org-contacts.el` is in the `load-path` and if the variable `org-contacts-files` is set, otherwise it throws an error. It does not check if the library is actually loaded, however, as-like me-others may have the library loading deferred.

### DIFF
--- a/consult-org.el
+++ b/consult-org.el
@@ -112,6 +112,17 @@ buffer are offered."
      :lookup #'consult--lookup-candidate)))
 
 ;;;###autoload
+(defun consult-org-contacts (&optional match)
+  "Jump to an Org contacts heading using files defined in `org-contacts-files'.
+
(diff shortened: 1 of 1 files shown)


minad:
Thanks for the proposal! Org-contacts is part of org-contrib and therefore out of scope of Consult. The command is simple enough to replicate it in the user configuration. It makes sense to add it to the Consult wiki. Feel free to add it there.

tfree87:
That sounds like a good idea.

PR #653, opened 2022-09-26
refactor location state combinators

jyp:
The intent is to improve customizability by exposing more building blocks as standalone functions.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -917,11 +917,10 @@ When no project is found and MAY-PROMPT is non-nil ask the user."
   (or (eq (selected-window) (active-minibuffer-window))
       (eq #'completion-list-mode (buffer-local-value 'major-mode (window-buffer)))))
 
-(defun consult--upgrading-location-state (candidates jump)
+(defun consult--location-upgrading-state (candidates state)
   "Location state function transformer.
-Transform the JUMP state function by upgrading the cheap location
-markers from CANDIDATES window selection change to full Emacs
-markers."
+Transform the STATE function. The cheap location markers from CANDIDATES are
+upgraded on window selection change to full Emacs markers."
   (let ((hook (make-symbol "consult--location-upgrade")))
     (fset hook
           (lambda (_)
@@ -932,13 +931,13 @@ markers."
       (pcase action
         ('setup (add-hook 'window-selection-change-functions hook))
         ('exit (remove-hook 'window-selection-change-functions hook)))
-      (funcall jump action cand))))
+      (funcall state action cand))))
 
 (defun consult--location-state (candidates)
   "Location state function.
 The cheap location markers from CANDIDATES are upgraded on window
 selection change to full Emacs markers."
-  (consult--upgrading-location-state candidates (consult--jump-state)))
+  (consult--location-upgrading-state candidates (consult--jump-state)))
 
 (defun consult--get-location (cand)
   "Return location from CAND."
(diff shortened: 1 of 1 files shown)


minad:
Did you assign copyright to the FSF for Emacs contributions? See https://github.com/minad/consult#contributions. This change is minor since it only reorders existing code, but it matters for larger contributions.

jyp:
Thanks a lot for the merge!

Yes, I have signed the copyright assignment forms in 2016.

jyp:
> Thanks, merged with minor changes. I introduced the more general consult--state-with-return. Where do you make use of the combinators?

I've made a version of consult-line which highlights the whole match.

oantolin:
How did you define "the whole match", @jyp?

jyp:
@oantolin I defined it in a completely ad-hoc way for my config, just what I need but buggy enough not to be worth sharing :) [Full disclosure: I re-interpret orderless patterns to regexps and search for that.]

A more principled way, but which requires more code, is to change consult--line-match to return both start and end of the match. Then the rest of the pipeline is relatively straightforward to change. But the change is pervasive.

minad:
Any proposals on how we could improve consult--line-match highlighting in general? The highlighting should work for all completion styles.

jyp:
@minad Here is how to do highlight the match for all completion styles.
1. Change `consult--line-match` to return both the beginning and end of the match (it already has code to compute those points, but no option can return them both)
2. In `consult--jump-preview`, change the 2nd over lay to use the end of the match rather than (1+ (point))
3. In one way or another, make sure that only one of [begin, end] of match is passed down to `consult--jump`

minad:
I will give this a try. We will see if the result is better than what we have now. I will add a third overlay for the match to the already existing whole line overlay and cursor overlay. See consult--jump-preview.

jyp:
@oantolin Speaking of this, I would if the matching end was detected by *adding* characters to the line until a match is found. This way, the end of the *first* match would be detected. I think that this behavior is preferable regardless, but in combination with my proposal it would mean that only the first match gets highlighted.

jyp:
@minad Nice :) But see also Daniel's objection and my reply.

minad:
I merged a first version in https://github.com/minad/consult/commit/1343e39fefcf8a28a7a415aa4b0a8ff7094370bf. I didn't yet take @jyp's proposal in https://github.com/minad/consult/pull/653#issuecomment-1261399271 into account. We can still improve consult--line-match later. Please let me know if everything still works for you. The change is straightforward, but still a bit intrusive since it affects multiple places of the preview pipeline.

oantolin:
I just tried it and with orderless, that definition of whole "match" can feel a little weird, since it can highlight lots of non-matching stuff and since the leftmost highlighted characters may come from matching a component that is not the first one, etc. I guess it still helps focus on the parts of the line that actually matched the input, since you at least know the first and last highlighted characters match something in the input.

Each completion styles knows how to do its own precise highlighting, would it be too slow to reuse the completion style's highlighting?

minad:
@oantolin

> Each completion styles knows how to do its own precise highlighting, would it be too slow (or otherwise problematic) to reuse the completion style's highlighting?

I think we rejected that approach early on because we don't know the completion style faces and it felt somehow "unclean".  But it shouldn't be a problem to detect faces if we first strip all faces from the selected candidate.

At least we have the machinery in place now to preview multiple matches via overlays. All we have to do is update consult--line-match to search for faces instead of doing the stepwise search.

minad:
@oantolin I implemented your suggestion. It seems to work nicely and I could share a little bit of code between consult-grep and consult-line.

oantolin:
It's great! I don't know about @jyp, but I do like this behavior better.

minad:
Omar, thanks for the feedback! I also like it, in particular with orderless, substring and basic. It's a bit odd with flex but that's okay.

Issue #616, opened 2022-07-19
Use grep-find-ignored-files in consult-grep


theottm:
Would it be possible to respect `grep-find-ignored-files` also when using `consult-grep`? Or is there a simple way to ignore them manually? It seems `M-x consul-grep #query -- [^~]` is not working for me.

aagon:
Hello Mr. Mendler.

First of all, thanks for this wonderful package.

Regarding the file and directory exclusions that should be respected by `consult-grep`, but also by `consult-ripgrep`, `consult-git-grep` and `consult-find`, I thought that it might be a good idea to use `project-ignores`(if it is bound of course) as well.

Similarly to what you did in 7c5a98a6abe5a42a9005d8c83a7ebef849e33e63, we could add in the respective command builders as exclusions all the files and directories present in the list returned by `project-ignores`. Some exclusions would probably be redundant with the ones already provided by `grep-find-ignored-files` and `grep-find-ignored-directories` or the ones already implemented by the external commands themselves, but that's not a problem.

Hence, for the current `project.el` backend, any vcs directory would be excluded (and not just `.git`, that I see you had to hardcode before 7c5a98a6abe5a42a9005d8c83a7ebef849e33e63) without us having to worry about it (let the `project-backend` worry about all that).

On top of that, if a new backend were to be defined, we would automatically inherit from all its defined exclusions.

And if you think that this amount to hardcoding too much "project logic" in the `consult-{grep,ripgrep,find,git-grep}` commands, we could always conditionate the respect of `project-ignores` to a user option.

I could try to provide a PR, if you are interested. I think it is a good step in the direction of further integration with `project.el`, which seems to be one of the directions you have already chosen for this package.

Mit freundlichen Grüssen,

Aymeric Agon-Rambosson

minad:
Hey,

I am not sure if we should also respect project-ignores. If we want that indeed, we should probably introduce a separate configuration variable which can be set either to the grep or the project ignores. Did you note that grep-find-ignore-directories is set from vc? So maybe project wouldn't add benefits on top?

Then there is the question if the other commands should respect the variables too. For ripgrep/git-grep I don't think so, since these tools come with their own ignore mechanism. Adding it for consult-find would be a good addition though, maybe only grep-find-ignore-directories?

Overall I don't have a coherent design in mind here. Proposals, prototypes and PRs are welcome!

aagon:
[...]

PR #654, opened 2022-09-27
Refactor consult--*-builder

aagon:
Like discussed in #616, a small proof of concept regarding the refactorisation and customisation of exclude arguments for consult-{grep,ripgrep,find,git-grep}.

Posted here for review by @minad.

A short remark : the `consult--build-arguments` function is such that every function inside the `consult-grep-args` must generate a list of split strings. We might want make it a little more robust. Since the function is idempotent, a simple idea to support functions producing unsplit strings is to replace `(consult--build-arguments (consult-grep-args))` with `(consult--build-arguments (consult--build-arguments (consult-grep-args)))`.

I have left `consult--grep-generate-exclude-args-project` for reference and if users want to use it, but haven't put it in the default value of `consult-grep-args`. The existence of this ready-to-use function should be advertised somewhere.

If you agree with this implementation, I can apply it to ripgrep, find and git-grep.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -236,42 +236,56 @@ See `consult--multi' for a description of the source data structure."
   "Regexp used to match file and line of grep output.")
 
 (defcustom consult-grep-args
-  "grep --null --line-buffered --color=never --ignore-case --line-number -I -r ."
+  '("grep" (consult--grep-generate-exclude-args)
+    "--null --line-buffered --color=never\
+    --ignore-case --line-number -I -r .")
   "Command line arguments for grep, see `consult-grep'.
-The dynamically computed arguments are appended."
-  :type 'string)
+The dynamically computed arguments are appended.
+
+Can be either a string, or a list of string or expressions."
+  :type '(choice string (repeat (choice string expression))))
 
 (defcustom consult-git-grep-args
   "git --no-pager grep --null --color=never --ignore-case\
    --extended-regexp --line-number -I"
   "Command line arguments for git-grep, see `consult-git-grep'.
-The dynamically computed arguments are appended."
-  :type 'string)
+The dynamically computed arguments are appended.
+
+Can be either a string, or a list of string or expressions."
+  :type '(choice string (repeat (choice string expression))))
 
 (defcustom consult-ripgrep-args
   "rg --null --line-buffered --color=never --max-columns=1000 --path-separator /\
    --smart-case --no-heading --line-number ."
   "Command line arguments for ripgrep, see `consult-ripgrep'.
-The dynamically computed arguments are appended."
-  :type 'string)
+The dynamically computed arguments are appended.
+
+Can be either a string, or a list of string or expressions."
+  :type '(choice string (repeat (choice string expression))))
 
 (defcustom consult-find-args
   "find . -not ( -wholename */.* -prune )"
   "Command line arguments for find, see `consult-find'.
-The dynamically computed arguments are appended."
-  :type 'string)
+The dynamically computed arguments are appended.
+
+Can be either a string, or a list of string or expressions."
+  :type '(choice string (repeat (choice string expression))))
 
 (defcustom consult-locate-args
   "locate --ignore-case --existing"
   "Command line arguments for locate, see `consult-locate'.
-The dynamically computed arguments are appended."
-  :type 'string)
+The dynamically computed arguments are appended.
+
+Can be either a string, or a list of string or expressions."
+  :type '(choice string (repeat (choice string expression))))
 
 (defcustom consult-man-args
   "man -k"
   "Command line arguments for man, see `consult-man'.
-The dynamically computed arguments are appended."
-  :type 'string)
+The dynamically computed arguments are appended.
+
+Can be either a string, or a list of string or expressions."
+  :type '(choice string (repeat (choice string expression))))
 
 (defcustom consult-preview-key 'any
   "Preview trigger keys, can be nil, \\='any, a single key or a list of keys."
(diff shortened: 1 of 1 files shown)


minad:
Thanks! Please adjust all the other commands too.

I don't want to call the splitter twice, even if the function is idempotent. I want to keep it simple.

I will add more comments to the code.

consult.el
```diff
@@ -236,10 +236,12 @@ See `consult--multi' for a description of the source data structure."
   "Regexp used to match file and line of grep output.")
 
 (defcustom consult-grep-args
-  "grep --null --line-buffered --color=never --ignore-case --line-number -I -r ."
+  '("grep" (consult--grep-generate-exclude-args-common)
+    "--null" "--line-buffered" "--color=never"
+    "--ignore-case" "--line-number" "-I" "-r" ".")
```
minad, line 241: You don't have to split all the strings here, since we us split-string-and-unquote.

consult.el
```diff
@@ -4405,6 +4415,25 @@ FIND-FILE is the file open function, defaulting to `find-file'."
...
+  "Produce grep exclude arguments based on `grep-find-ignored-files'."
+  (unless (boundp 'grep-find-ignored-files) (require 'grep))
+  (append
+   (mapcar (lambda (s) (concat "--exclude=" s))
+	   (bound-and-true-p grep-find-ignored-files))
+   (mapcar (lambda (s) (concat "--exclude-dir=" s))
+	   (bound-and-true-p grep-find-ignored-directories))))
+
+(defun consult--grep-generate-exclude-args-project ()
+  "Produce grep exclude arguments based on `project-ignores'."
```
minad, line 4428: I don't want to provide that function here. Users can add this code to their user configuration.

consult.el
```diff
@@ -1106,6 +1108,14 @@ MARKER is the cursor position."
   (let ((inhibit-field-text-motion t))
     (consult--region-with-cursor (line-beginning-position) (line-end-position) marker)))
 
+(defun consult--build-arguments (str)
+  "Return arg as flat list of split strings."
```
minad, line 1112: Please expand the docstring here.

minad: Do you have package-lint enabled? It should warn you about the undocumented argument.

consult.el
aagon: done

consult.el
aagon: removed. Shall I add it to the wiki ?

consult.el
aagon: Done, tell me what you think.

consult.el
```diff
@@ -236,10 +236,12 @@ See `consult--multi' for a description of the source data structure."
...
 
 (defcustom consult-grep-args
-  "grep --null --line-buffered --color=never --ignore-case --line-number -I -r ."
+  '("grep" (consult--grep-generate-exclude-args-common)
+    "--null" "--line-buffered" "--color=never"
+    "--ignore-case" "--line-number" "-I" "-r" ".")
   "Command line arguments for grep, see `consult-grep'.
 The dynamically computed arguments are appended."
-  :type 'string)
+  :type '(choice string list))
```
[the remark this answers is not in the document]
aagon: done

consult.el
```diff
@@ -236,10 +236,12 @@ See `consult--multi' for a description of the source data structure."
   "Regexp used to match file and line of grep output.")
 
 (defcustom consult-grep-args
-  "grep --null --line-buffered --color=never --ignore-case --line-number -I -r ."
+  '("grep" (consult--grep-generate-exclude-args-common)
+    "--null --line-buffered --color=never\
+    --ignore-case --line-number -I -r .")
   "Command line arguments for grep, see `consult-grep'.
```
minad, line 242: The docstring should be expanded. It should mention both possibilities - strings and lists with evaluated arguments.

minad:
Looks good. Can you please also adjust the defcustoms (types and do strings) of the other commands? find, git-grep, ripgrep, man. Then it should be ready to merge. I can do this later.

aagon:
I think I have applied all your comments. I have also adjusted consult-locate.

consult.el
```diff
@@ -236,42 +236,56 @@ See `consult--multi' for a description of the source data structure."
...
+  '("grep" (consult--grep-generate-exclude-args)
+    "--null --line-buffered --color=never\
+    --ignore-case --line-number -I -r .")
   "Command line arguments for grep, see `consult-grep'.
-The dynamically computed arguments are appended."
-  :type 'string)
+The dynamically computed arguments are appended.
+
+Can be either a string, or a list of string or expressions."
+  :type '(choice string (repeat string)))
```
minad, line 246: Sorry, what I mentioned about the type before was wrong. Should be (choice string (repeat (choice string expression))).

consult.el
```diff
@@ -4405,6 +4419,15 @@ FIND-FILE is the file open function, defaulting to `find-file'."
       (substring cand (1+ (length (get-text-property 0 'consult--grep-file cand))))
     (get-text-property 0 'consult--grep-file cand)))
 
+(defun consult--grep-generate-exclude-args-common ()
```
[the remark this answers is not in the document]
aagon: done

consult.el
aagon: done

minad:
Thanks. Did you assign copyright to the FSF for your Emacs contributions? See https://github.com/minad/consult#contributions

minad:
Merged in https://github.com/minad/consult/commit/fae2c56bdc6832af65b85a3f2111be32eb2e68b2

PR #658, opened 2022-09-30
Fix compiler warning about docstring wider than 80 characters

dr-scsi:
* consult.el (consult-highlight-match): Slightly re-write docstring.


minad:
Thx

PR #661, opened 2022-10-04
Add consult-imenu--jump support for non default goto function

dalanicolai:
Imenu allows for setting a custom `imenu-default-goto-function` (the symbol represents both a function and a variable). This commit makes consult-imenu--jump (not yet the preview functionality) support a custom goto function.

### DIFF
--- a/consult-imenu.el
+++ b/consult-imenu.el
@@ -161,7 +161,9 @@ In contrast to the builtin `imenu' jump function,
 this function can jump across buffers."
   (pcase item
     (`(,name ,pos ,fn . ,args) (apply fn name pos args))
-    (`(,_ . ,pos) (consult--jump pos))
+    (`(,name . ,pos) (if (eq imenu-default-goto-function 'imenu-default-goto-function)
+		         (consult--jump pos)
+		       (funcall imenu-default-goto-function name pos)))
     (_ (error "Unknown imenu item: %S" item))))
 
 (defun consult-imenu--narrow ()


minad:
I see. We don't support the `imenu-default-goto-function` yet. It is hard to support this fully for preview, since the goto function can perform modifications which cannot be undone generally. We could use it for the final jump as proposed here, however we have to switch to the buffer first in the case of `consult-imenu-multi`. Furthermore the downside is that we lose the recentering functionality and other functionality of `consult--jump`, which is not nice, given that `consult-imenu` behaves similarly as other commands like `consult-outline` etc. We have to think this through a little bit more.

dalanicolai:
I am not yet familiar with all those consult features (I am still using Ivy because I am still more productive with it, but I'd like to gradually switch to Vertico, as it looks to me that its design ultimately fits better with Emacs). Therfore, I'd like the [doc-tools](https://github.com/dalanicolai/doc-tools) package (which works quite fine already, but is still heavily in development) to work with Vertico also. Anyway, currently I think I can not help too much with the 'thinking process'.

However, for a first temporary solution, I would say it is fine if the preview just gets disabled when a custom jump function has been defined. And for the final jump, I guess this commit does not break any current functionality (of course you could think about it and redesign it later).

Finally, although I guess accepting the current commit, will do no harm, it, on the other hand, is not a very urgent fix (as I said I am still using Ivy myself). Just it would make the doc-tools imenu (final jump) functionality work with Vertico already...

minad:
> However, for a first temporary solution, I would say it is fine if the preview just gets disabled when a custom jump function has been defined. And for the final jump, I guess this commit does not break any current functionality (of course you could think about it and redesign it later).

Okay, let's do this for now.

dalanicolai:
Okay great! Thanks!

Then Vertico users can 'try out' doc-tools imenu functionality already also (although I did not yet 'officially' announce doc-tools development anywhere:)...

PR #665, opened 2022-10-13
consult-fd

Uthar:
I'm working on `consult-fd`.

`fd` (https://github.com/sharkdp/fd) is a "faster `find`".


minad:
Rejected, see https://github.com/minad/consult/issues/6#issuecomment-1278252044

minad:
I've finally added `consult-fd` in 5960ad787d8d1051d9c3140219d021496731e14c, which uses the new `--and` option of `fd`.

PR #669, opened 2022-10-15
Enable setting an index var from consult-mode-histories

jdtsmith:
`consult-mode-histories` can be a cons or 3 element list, the last member of which is the name of an index variable to update with the index position of the selected history element on the history list or ring.  Note that the _first_ such matching element is taken.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -135,12 +135,18 @@ This applies to asynchronous commands, e.g., `consult-grep'."
   :type '(alist :key-type symbol :value-type plist))
 
 (defcustom consult-mode-histories
-  '((eshell-mode . eshell-history-ring)
-    (comint-mode . comint-input-ring)
-    (term-mode   . term-input-ring))
-  "Alist of (mode . history) pairs of mode histories.
-The histories can be rings or lists."
-  :type '(alist :key-type symbol :value-type symbol))
+  '((eshell-mode eshell-history-ring eshell-history-index)
+    (comint-mode comint-input-ring   comint-input-ring-index)
+    (term-mode   term-input-ring     term-input-ring-index))
+  "Alist of mode histories as (mode . history) or (mode history index).
+The histories can be rings or lists.  INDEX, if provided, is a
+variable to set to the index of the selection within the ring or
+list."
+  :type '(alist :key-type symbol
+		:value-type (choice (symbol :tag "List or Ring Name")
+				    (group :tag "Include Index"
+					   (symbol :tag "List/Ring")
+					   (symbol :tag "Index Var")))))
 
 (defcustom consult-themes nil
   "List of themes (symbols or regexps) to be presented for selection.
@@ -3724,35 +3730,40 @@ This command can act as a drop-in replacement for `repeat-complex-command'."
 (defun consult--current-history (&optional history)
   "Return the normalized HISTORY or the history relevant to the current buffer.
 
-If the minibuffer is active, returns the minibuffer history,
-otherwise the history corresponding to the mode is returned.
-There is a special case for `repeat-complex-command',
-for which the command history is used."
-  (cond
-   (history)
-   ;; If pressing "C-x M-:", i.e., `repeat-complex-command',
-   ;; we are instead querying the `command-history' and get a full s-expression.
-   ;; Alternatively you might want to use `consult-complex-command',
-   ;; which can also be bound to "C-x M-:"!
-   ((eq last-command 'repeat-complex-command)
-    (setq history (mapcar #'prin1-to-string command-history)))
-   ;; In the minibuffer we use the current minibuffer history,
-   ;; which can be configured by setting `minibuffer-history-variable'.
-   ((minibufferp)
-    (if (eq minibuffer-history-variable t)
-        (user-error "Minibuffer history is disabled for `%s'" this-command)
-      (setq history (mapcar #'consult--tofu-hide (symbol-value minibuffer-history-variable)))))
-   ;; Otherwise we use a mode-specific history, see `consult-mode-histories'.
-   (t (when-let (found
-                 (or (seq-find (lambda (ring)
-                                 (and (derived-mode-p (car ring))
-                                      (boundp (cdr ring))))
-                               consult-mode-histories)
-                     (user-error
-                      "No history configured for `%s', see `consult-mode-histories'"
-                      major-mode)))
-        (setq history (symbol-value (cdr found))))))
-  (consult--remove-dups (if (ring-p history) (ring-elements history) history)))
+Returns the cons (HISTORY . INDEX).  If the minibuffer is active,
+HISTORY is the minibuffer history, otherwise the history
+corresponding to the mode.  There is a special case for
+`repeat-complex-command', for which the command history is used.
+INDEX is the name of the index variable to update, if any."
+  (let (index)
+    (cond
+     (history)
+     ;; If pressing "C-x M-:", i.e., `repeat-complex-command',
+     ;; we are instead querying the `command-history' and get a full s-expression.
+     ;; Alternatively you might want to use `consult-complex-command',
+     ;; which can also be bound to "C-x M-:"!
+     ((eq last-command 'repeat-complex-command)
+      (setq history (mapcar #'prin1-to-string command-history)))
+     ;; In the minibuffer we use the current minibuffer history,
+     ;; which can be configured by setting `minibuffer-history-variable'.
+     ((minibufferp)
+      (if (eq minibuffer-history-variable t)
+	  (user-error "Minibuffer history is disabled for `%s'" this-command)
+	(setq history (mapcar #'consult--tofu-hide (symbol-value minibuffer-history-variable)))))
+     ;; Otherwise we use a mode-specific history, see `consult-mode-histories'.
+     (t (when-let (found
+		   (or (seq-find (lambda (ring)
+				   (and (derived-mode-p (car ring))
+					(boundp (let (c (cdr ring)) (if (consp c) (car c) c)))))
+				 consult-mode-histories)
+		       (user-error
+			"No history configured for `%s', see `consult-mode-histories'"
+			major-mode)))
+	  (if (consp (cdr found))
+	      (setq history (symbol-value (cadr found))
+		    index (caddr found))
+	    (setq history (symbol-value (cdr found)))))))
+    (cons history index)))
 
 ;; This command has been adopted from https://github.com/oantolin/completing-history/.
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


jdtsmith:
Tested on comint only so far.  Now that I play with this, I wonder if we should also consider another member of the list (4 elements total): a "saved" index variable (e.g. `comint-save-input-ring-index`).  Together, these 2 index vars could be used to specify the _starting index_ within the ring.  Consult could then add an index prefix to each history command and wrap around, just like `consult-line` does (coloring the index red when it _wraps_).  With this `consult-history` would allow a quick "browse nearby history" when you have already just pulled from history, either before or after sending it.

minad:
On 10/15/22 20:40, JD Smith wrote:
> Tested on comint only so far. Now that I play with this, I wonder if we
> should also consider another member of the list (4 elements total): a
> "saved" index variable (e.g. |comint-save-input-ring-index|). Together,
> these 2 index vars could be used to specify the /starting index/ within
> the ring. Consult could then add an index prefix to each history command
> and wrap around, just like |consult-line| does (coloring the index red
> when it /wraps/). With this |consult-history| would allow a quick
> "browse nearby history" when you have already just pulled from history,
> either before or after sending it.

I don't get it. What is the purpose of comint-save-input-ring-index? Why
is there not only single pointer to the last selected element?

Regarding rotating ala consult-line - we could use an index to rotate
the list to the current position before passing the list to
consult--read. However I am a bit worried that this would lead to a
worse UI overall since the history would rotate all the time as soon as
you select an element. Also I don't think we should add an index as
prefix, since the number isn't meaningful, in contrast to line numbers.

minad:
I think I prefer to leave it as is for now and avoid the rotation feature you suggested since this may lead to a lot of confusion. Believe it or not, but even for consult-line, where it makes sense, is easy to understand and is useful, it led to confusion.

minad:
Sorry, I had to revert again. There were too many issues. I will comment in the code.

consult.el
```diff
@@ -3724,35 +3730,40 @@ This command can act as a drop-in replacement for `repeat-complex-command'."
...
-                      "No history configured for `%s', see `consult-mode-histories'"
-                      major-mode)))
-        (setq history (symbol-value (cdr found))))))
-  (consult--remove-dups (if (ring-p history) (ring-elements history) history)))
+Returns the cons (HISTORY . INDEX).  If the minibuffer is active,
+HISTORY is the minibuffer history, otherwise the history
+corresponding to the mode.  There is a special case for
+`repeat-complex-command', for which the command history is used.
+INDEX is the name of the index variable to update, if any."
+  (let (index)
```
minad, line 3738: Remove the let, don't write in an imperative style. Instead each cond should return a pair directly.

consult.el
```diff
@@ -3761,25 +3772,32 @@ for which the command history is used."
...
+		      ('face-name-history 'face)
+		      ('read-envvar-name-history 'environment-variable)
+		      ('bookmark-history 'bookmark)
+		      ('file-name-history 'file)))
+	       :sort nil
+	       :state (consult--insertion-preview (point) (point))))))
     (when (minibufferp)
-      (delete-minibuffer-contents))
+      (delete-minibuffer-contents)) 
+    (if index-var
```
minad, line 3797: when instead of if

consult.el
```diff
@@ -3761,25 +3772,32 @@ for which the command history is used."
...
-                       ('file-name-history 'file)))
-                :sort nil
-                :state (consult--insertion-preview (point) (point))))))
+  (pcase-let*
+      ((`(,history . ,index-var)
+	(or (consult--current-history history)
+	    (user-error "History is empty")))
+       (str (consult--local-let ((enable-recursive-minibuffers t))
+	      (consult--read
+	       (consult--remove-dups (if (ring-p history) (ring-elements history) history))
```
minad, line 3781: Instead of returning a ring, always return a list.

consult.el
```diff
@@ -3761,25 +3772,32 @@ for which the command history is used."
...
+		      ('file-name-history 'file)))
+	       :sort nil
+	       :state (consult--insertion-preview (point) (point))))))
     (when (minibufferp)
-      (delete-minibuffer-contents))
+      (delete-minibuffer-contents)) 
+    (if index-var
+	(set index-var (if (ring-p history)
+			   (ring-member history str)
+			 (seq-position history str))))
```
minad, line 3800: We can always use seq-position here.

consult.el
```diff
@@ -3724,35 +3730,40 @@ This command can act as a drop-in replacement for `repeat-complex-command'."
...
+     ;; which can be configured by setting `minibuffer-history-variable'.
+     ((minibufferp)
+      (if (eq minibuffer-history-variable t)
+	  (user-error "Minibuffer history is disabled for `%s'" this-command)
+	(setq history (mapcar #'consult--tofu-hide (symbol-value minibuffer-history-variable)))))
+     ;; Otherwise we use a mode-specific history, see `consult-mode-histories'.
+     (t (when-let (found
+		   (or (seq-find (lambda (ring)
+				   (and (derived-mode-p (car ring))
+					(boundp (let (c (cdr ring)) (if (consp c) (car c) c)))))
```
minad, line 3757: I got some compilation warnings for this line. Do you use package-lint?

minad:
I committed a fixed version, see https://github.com/minad/consult/commit/cb5aea3e5e3d3b434705b1662df8cce316460264. Please check if it works for you.

jdtsmith:
> Consult could then add an index prefix to each history command and wrap around, just like |consult-line| does (coloring the index red when it /wraps/). With this |consult-history| would allow a quick "browse nearby history" when you have already just pulled from history, either before or after sending it.

> I don't get it. What is the purpose of comint-save-input-ring-index? Why is there not only single pointer to the last selected element? 

Because comint wants to tell the difference between "I've pulled something from history to the current command line" (non-save version set) and "I just sent a command that I pulled from history but haven't yet pulled anything else" (save version set, non-save version nil).  

> Regarding rotating ala consult-line - we could use an index to rotate the list to the current position before passing the list to consult--read. However I am a bit worried that this would lead to a worse UI overall since the history would rotate all the time as soon as you select an element. 

OK can think about this more. 

> Also I don't think we should add an index as prefix, since the number isn't meaningful, in contrast to line numbers.

Interesting.  Comint mentions the history number quite often. For example, when using any of the history-search, or next from history.

jdtsmith:
> Believe it or not, but even for consult-line, where it makes sense, is easy to understand and is useful, it led to confusion.

I don't believe it!  This feature is excellent, since it orients you "where you are".  I honestly often get lost in terms of "where I am in history" with comint.  Comints helpfully tells you index all the time (including every time I up arrow), but the mapping from index to saved command is never apparent.  That's why I thought adding index would be useful, but if people can't get behind _line numbers_... geez.

consult.el
jdtsmith: Oops looks like I forgot to double wrap the let vars...  Do you run a linter on save?  Never set that up.

minad:
> > Also I don't think we should add an index as prefix, since the number isn't meaningful, in contrast to line numbers.
> 
> Interesting. Comint mentions the history number quite often. For example, when using any of the history-search, or next from history.

What I mean is that the history changes often, while line numbers are more stable over time. This means the history number is much less meaningful to the user. I am not saying that the history number is never useful, but just not often enough for consult-history and not so much in a completion context. Iirc this discussion has also come up a few times - completion/selection is "random access" and it doesn't play so well together with moving step by step through a history. On top we also remove duplicates.

> I don't believe it! This feature is excellent, since it orients you "where you are". I honestly often get lost in terms of "where I am in history" with comint. Comints helpfully tells you index all the time (including every time I up arrow), but the mapping from index to saved command is never apparent. That's why I thought adding index would be useful, but if people can't get behind line numbers... geez.

You misunderstood. I meant that people got confused by the wrap around in consult-line. That's all. If we add wrap around to consult-history it may make things more confusing. Adding numbers would be fine, but I argue that they are not sufficiently useful here.

> One other important point: usually both of these index variables are nil. After entering any new command of any kind, they are both nil. Only when you i) are in the midst of recalling a command from history to the prompt line or ii) have just sent such a recalled history command are they non-nil. This is where the wraparound thing would happen. I mention this because it seems you are thinking the history list would be almost always wrapped, but in fact it would almost never be.

Okay, fair enough. So it will be reset every time you enter a new command such that the wrap around would happen only directly after a history search? Anyway, I think what we have now is fine.

consult.el
minad: I have the linter always on via flymake. In every Elisp buffer, also in my init.el. It is quite helpful.

minad: What I should probably set up also is some CI for github to check PRs.

jdtsmith:
> So it will be reset every time you enter a new command such that the wrap around would happen only directly after a history search?

Exactly.  After every new command is sent, the index is nil, and the saved index is the last index value (also usually nil).  I've spent too much time reading comint sources.

minad:
Haha. Regarding the number information of the comint history, I've also observed before that comint is telling me the number and I recall that I found it somehow too much. Do you really use this information? It probably depends on your workflow. I am more used to a quick and dirty approach, searching for the command I want via completion or backward search instead of carefully juggling with the history and recalling commands by position.

jdtsmith:
Yeah it's a bit TMI; but for searches I do like it because at a glance it tells me how far I'm reaching back into history (I wonder whether consult-history should mention the index if the index var is set?).  I think moderately more useful is seeing the local context when you are "in the midst of pulling history".  The idea being that if you pulled something from history at the current or last prompt, you are likely to be more interested in commands _around_ that one.   But as you say, search-to-narrow and scroll-through-history function quite differently.  

BTW, `C-c C-l`, which shows an `*Input History*` buffer, does neither (show index, or indicate current "position" via a highlight and/or placing window point).  So probably this means it wouldn't be that useful, would be confusing, or at least nobody thought of it.

minad:
You remind me of an important point I wanted to bring up yesterday but forgot - completion doesn't preserve context! It may even make a mess by reordering (flex sorting). But then I thought it could be interesting to write a context-preserving completion style, like Orderless but some items before and after a match are kept and maybe greyed out. Would be an interesting experiment at least.

jdtsmith:
Very interesting.  So more like a Rolodex, where search is used just to spin through it, or like grep with context.  

I bet just making this some kind of `:with-context 3` type option of orderless wouldn't be too difficult.  And you can presumably still select these grayed-out candidates, if your search is just serving to get you "near" to what you actually wanted.  Would be great for the quite common case of "What was the command I entered after I scp'd that file over?".  Any selection list backed by a time-tagged history file could benefit from this, though I'd probably want to toggle it on case by case since it adds some verbosity.

minad:
There are a few problems though - if you add context before, the first candidate won't be selected in Vertico for example. 

Adding support to Orderless won't be directly possible since we have to add an additional postprocessing step to add the context candidates. The fast filtering happens using all-completions and I think it would be good to preserve that. 

I am thinking more of a "wrapper completion style", which takes another completion style as argument and adds context in an additional O(n) pass.

jdtsmith:
Perhaps you could use tofus to encode index.  Then orderless with its all-completions would effectively report the index of all candidates. The style wrapper could remove these index-tofus, "dilate" them by some number of context neighbors (taking care of overlapping neighborhoods) and style those specially (grayed).  I guess it would also have to enforce a by-index sort order. 

BTW, I just checked and was surprised to find all-completions preserves text properties from its input collection.  At what point in the completion chain do those go missing, necessitating tofus?

minad:
> BTW, I just checked and was surprised to find all-completions preserves
> text properties from its input collection. At what point in the
> completion chain do those go missing, necessitating tofus?

It is not so surprising if you consider how consult-line looks for
example. The properties get lost right in the end when the minibuffer
contents are returned after exiting the minibuffer. The point is that
the minibuffer may hold a fully propertized candidate, but it could as
well hold a plain string, which was manually typed out by the user if it
is still a valid candidate. This is the price we pay for *completion*
(think prefix/tab completion), in contrast to *selection* which is what
we do in Vertico with Orderless. *Completion* leads to intermediate
partial candidate strings in the minibuffer.

jdtsmith:
Aha.  I suppose tofus also help defeat unwanted de-duplication independent of the "where to store candidate metadata" issue.  But what to make of:

> If the variable minibuffer-allow-text-properties is non-nil, then the string that is returned includes whatever text properties were present in the minibuffer. Otherwise all the text properties are stripped when the value is returned.

minad:
> Aha. I suppose tofus also help defeat unwanted de-duplication
> independent of the "where to store candidate metadata" issue. 
Yes.

> But what to make of:
> 
>     If the variable minibuffer-allow-text-properties is non-nil, then
>     the string that is returned includes whatever text properties were
>     present in the minibuffer. Otherwise all the text properties are
>     stripped when the value is returned.

This is a recent addition on 27 or 28, I don't recall. The problem with
this addition is that it is not really supported by completing-read. If
minibuffer-allow-text-properties is non-nil and completing-read is
called, the minibuffer content could be returned propertized.

But this would still create the issue, which I mentioned before: If the
user typed out the candidate manually, how could one distinguish
duplicates then? Where to take the properties from? The completion UI
must support selection in order to select from otherwise equal
candidates. Also the UI must not delete duplicates. See
delete-consecutive-duplicates in the default completion UI and Vertico.

For Vertico one could establish the contract that if
minibuffer-allow-text-properties is non-nil:

1) Matching candidates are returned propertized (REQUIRE-MATCH non-nil)
2) No (consecutive) duplicates are deleted

But this contract wouldn't be supported by other UIs for a while, in
particular not by default completion which completes step by step and is
not a selecting UI by construction. It could be made into one by always
showing the *Completions* buffer if there are multiple duplicates to
select from.

Furthermore it would require changing the contract with Embark which
currently always ensures that candidates are stripped of text properties.

All in all this would lead to more special casing and we could still not
get rid of the tofus. Anyway, is there an actual problem with the tofus
or do you just rethink this design again?

PR #670, opened 2022-10-15
Fix for previewing on not current window

aki2o:
`consult--jump-preview` could not work with the feature that manages `display-buffer-function` like [e2wm.el](https://github.com/kiwanami/emacs-window-manager).
That's because `consult--buffer-action` don't have a consideration for the case that `current-buffer` won't be switched.

So, I tried to fix this trouble and it looks to work well with this fix.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -1357,22 +1357,36 @@ See `isearch-open-necessary-overlays' and `isearch-open-overlay-temporary'."
                 restore))))
     restore))
 
-(defun consult--jump-1 (pos)
-  "Go to POS and recenter."
-  (if (and (markerp pos) (not (marker-buffer pos)))
-      ;; Only print a message, no error in order to not mess
-      ;; with the minibuffer update hook.
-      (message "Buffer is dead")
-    ;; Switch to buffer if it is not visible
-    (when (and (markerp pos) (not (eq (current-buffer) (marker-buffer pos))))
-      (consult--buffer-action (marker-buffer pos) 'norecord))
-    ;; Widen if we cannot jump to the position (idea from flycheck-jump-to-error)
-    (unless (= (goto-char pos) (point))
-      (widen)
-      (goto-char pos))))
+(defmacro consult--with-jump (pos &rest body)
+  "Go to POS."
+  (declare (indent 1))
+  `(if (and (markerp ,pos) (not (marker-buffer ,pos)))
+       ;; Only print a message, no error in order to not mess
+       ;; with the minibuffer update hook.
+       (message "Buffer is dead")
+     (let* ((goto-other (and (markerp ,pos) (not (eq (current-buffer) (marker-buffer ,pos)))))
+            (w (cond (goto-other
+                      (consult--buffer-action (marker-buffer ,pos) 'norecord)
+                      (get-buffer-window (marker-buffer ,pos)))
+                     (t
+                      (get-buffer-window (current-buffer)))))
+            (f (lambda ()
+                 (unless (= (goto-char ,pos) (point))
+                   (widen) ;; Widen if we cannot jump to the position (idea from flycheck-jump-to-error)
+                   (goto-char ,pos))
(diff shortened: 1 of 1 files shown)


consult.el
```diff
@@ -469,7 +469,7 @@ called from the corresponding command. Note that the options depend on
 the private `consult--read' API and should not be considered as stable
 as the public API.")
 
-(defvar consult--buffer-display #'switch-to-buffer
+(defvar consult--buffer-display #'display-buffer
```
aki2o, line 472: I'm using e2wm.el and it must support `switch-to-buffer`.
However, it happens to collapse window layout in my environment.

Anyway, I believe `display-buffer` is more appropriate for this package role. So, I hope this fix be included

consult.el
```diff
@@ -1410,10 +1423,10 @@ The function can be used as the `:state' argument of `consult--read'."
...
                 (narrow-to-region saved-min saved-max)
                 (goto-char saved-pos)))
-            ;; Handle positions with overlay information
-            (consult--jump-1 (or (car-safe cand) cand))
-            (setq invisible (consult--invisible-open-temporarily)
-                  overlays
+          ;; Handle positions with overlay information
+          (consult--with-jump (or (car-safe cand) cand)
+            (setq consult--preview-previous-invisibles (consult--invisible-open-temporarily)
+                  consult--preview-previous-overlays
```
aki2o, line 1429: I feel `invisible` should be defined for not local but global because `void variable` error happened.

consult.el
```diff
@@ -1357,22 +1357,36 @@ See `isearch-open-necessary-overlays' and `isearch-open-overlay-temporary'."
...
+           (with-selected-window w (funcall f))
+         (funcall f)))))
+
+(defvar consult--preview-previous-invisibles nil)
+(defvar consult--preview-previous-overlays nil)
 
 (defun consult--jump (pos)
   "Push current position to mark ring, go to POS and recenter."
+  (mapc #'funcall consult--preview-previous-invisibles)
+  (mapc #'delete-overlay consult--preview-previous-overlays)
```
aki2o, line 1389: it looks like the variables need to be cleared here as well.
And I guess there is a timing that's more appropriate to do this procedure but I couldn't find it.

consult.el
```diff
@@ -1357,22 +1357,36 @@ See `isearch-open-necessary-overlays' and `isearch-open-overlay-temporary'."
...
+                     (t
+                      (get-buffer-window (current-buffer)))))
+            (f (lambda ()
+                 (unless (= (goto-char ,pos) (point))
+                   (widen) ;; Widen if we cannot jump to the position (idea from flycheck-jump-to-error)
+                   (goto-char ,pos))
+                 ,@body
+                 (run-hooks 'consult-after-jump-hook))))
+       (if (window-live-p w)
+           (with-selected-window w (funcall f))
```
aki2o, line 1380: Here is taking care of the case like

* `current-buffer` is not the target buffer
* the window of the target buffer is not selected window

consult.el
minad: No, switch-to-buffer is the better choice here. You can set switch-to-buffer-obey-display-actions or change the variable in your configuration. Please change back.

consult.el
minad: Yes, this is wrong. Please revert the change to the overlays variable.

consult.el
```diff
@@ -1357,22 +1357,36 @@ See `isearch-open-necessary-overlays' and `isearch-open-overlay-temporary'."
...
+         (funcall f)))))
+
+(defvar consult--preview-previous-invisibles nil)
+(defvar consult--preview-previous-overlays nil)
```
minad, line 1384: Remove these variables.

minad:
I close this PR in favor of #671. We should first fix consult--buffer-preview to handle display-buffer properly before trying to fix consult--jump-preview.

PR #671, opened 2022-10-16
Ensure that buffer preview supports general display-buffer

minad:
The problem is that set-window-configuration only supports the dont-set-miniwindow option since Emacs 28!

Experimenting with these test buffers:

~~~ elisp
(setq display-buffer-alist
      `(("\\`\\*no-window\\*\\'"
         (display-buffer-no-window))
        ("\\`\\*at-bottom\\*\\'"
         (display-buffer-at-bottom))
        ("\\`\\*in-side-window\\*\\'"
         (display-buffer-in-side-window))
        ("\\`\\*pop-up-window\\*\\'"
         (display-buffer-pop-up-window))))
~~~

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -4125,21 +4125,15 @@ Report progress and return a list of the results"
 
 (defun consult--buffer-preview ()
   "Buffer preview function."
-  (let ((orig-buf (current-buffer)) other-win)
+  (let (wc)
     (lambda (action cand)
       (when (eq action 'preview)
-        (when (and (eq consult--buffer-display #'switch-to-buffer-other-window)
-                   (not other-win))
-          (switch-to-buffer-other-window orig-buf)
-          (setq other-win (selected-window)))
-        (let ((win (or other-win (selected-window))))
-          (when (window-live-p win)
-            (with-selected-window win
-              (cond
-               ((and cand (get-buffer cand))
-                (switch-to-buffer cand 'norecord))
-               ((buffer-live-p orig-buf)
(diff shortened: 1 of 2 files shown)


consult.el
```diff
@@ -4125,21 +4125,15 @@ Report progress and return a list of the results"
...
-        (let ((win (or other-win (selected-window))))
-          (when (window-live-p win)
-            (with-selected-window win
-              (cond
-               ((and cand (get-buffer cand))
-                (switch-to-buffer cand 'norecord))
-               ((buffer-live-p orig-buf)
-                (switch-to-buffer orig-buf 'norecord))))))))))
+        (when wc
+          (set-window-configuration wc 'no-frame 'no-miniwindow)
```
minad, line 4132: @aki2o The problem is here that we don't have that functionality available on Emacs 27!

minad:
@aki2o I fixed consult--jump in https://github.com/minad/consult/pull/671/commits/b5033d244172298e274c16bcf0a72e2a04434e69. Please read the code and let me know what you think about it. Does it work in your use case? Unfortunately we cannot merge this for quite some time due to the limitations of set-window-configuration, which doesn't take the no-miniwindow argument before Emacs 28. Do you have an idea on how one could shim that functionality on Emacs 27?

minad:
Another problem is that the file preview is broken now. Trying to generalize the preview is like Pandora's box :(

minad:
Close this for now, since it is unclear how the issues in https://github.com/minad/consult/pull/671#issuecomment-1279885026 can be resolved in full generality.

Issue #668, opened 2022-10-15
consult-yank-pop: Support yank-from-kill-ring-rotate


minad:
I have the same kind of problem. I'm used to have the `kill-ring` cycle when using `yank-pop` and can't get this behaviour with `consult-yank-pop`

When using `yank-pop` the value of `kill-ring-yank-pointer` changes. Example, I just killed 1, 2 and 3.

`kill-ring`:

```elisp
(#("3" 0 1
   (fontified t))
 #("2" 0 1
   (fontified t))
 #("1" 0 1
   (fontified t)))
```

`kill-ring-yank-pointer`

```elisp
(#("3" 0 1
   (fontified t))
 #("2" 0 1
   (fontified t))
 #("1" 0 1
   (fontified t)))
```

If I `yank-pop` 3 times it will yank `1` and the value of `kill-ring-yank-pointer` changes:

```elisp
(#("1" 0 1 (fontified t)))
```

When starting from the same initial setup (1, 2 and 3 have been killed) with consult, I can hit `consult-yank-pop` once and select from its menu to chose the value I want to yank. After selecting `1`, the value of `kill-ring-yank-pointer` hasn't changed and if I use `consult-yank-pop` after that the default value will be `3` and not `1`.

It looks like this cycling can be opted in or out:

```elisp
(defcustom yank-from-kill-ring-rotate t
  "Whether using `yank-from-kill-ring' should rotate `kill-ring-yank-pointer'.
If non-nil, the kill ring is rotated after selecting previously killed text."
  :type 'boolean
  :group 'killing
  :version "28.1")
```

But this doesn't seem to be used in consult

_Originally posted by @mattiasdrp in https://github.com/minad/consult/issues/553#issuecomment-1239125862_

mattiasdrp:
I'm sorry, couldn't find the time to work on it. Do you plan on making a PR for it?

minad:
No, but @jdtsmith's recent issue #667 reminded me of this and I took a look if you made some progress. I don't plan to work on this myself, but I will keep it open for a while to see if there is some interest.

mattiasdrp:
I still plan to work on it but I can't provide a time where it'll be done :-)

PR #681, opened 2022-11-21
Move yanked string to top of kill ring after yanking

YorkZ:
Firstly, thank you so much for this amazing project.

Please review this PR. I bind `M-y` to `consult-yank-pop`, and this change made my life a lot easier because previously when I needed to yank the same content again and again from the kill ring, I had to pick it from the kill ring far away from the top, not to mention that the content would eventually be removed from the kill ring altogether.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -3569,6 +3569,8 @@ version supports preview of the selected string."
     (push-mark)
     (insert-for-yank string)
     (setq this-command 'yank)
+    (setq kill-ring (delete string kill-ring))
+    (kill-new string)
     (when (consp arg)
       ;; Swap point and mark like in `yank'.
       (goto-char (prog1 (mark t)


minad:
Thanks. I am not sure about this. We should probably make it configurable. How does this relate to #668 (kill ring rotation)? Ideally we implement a unified solution for #668 which rotates the kill ring (if enabled via an option).

YorkZ:
I'm not sure I understand exactly what #668 means, but I would guess it says the same thing. I think #668 proposed a customizable variable but no implementation yet right? So if you decide to add the customization, the implementation would be similar to this PR right?

minad:
No, I think it would be different. There is some index used to rotate the ring.

YorkZ:
I don't know how `consult` works in this regard. So deleting the entry from the kill ring and killing the content again to bring it to the top doesn't rotate the part of kill ring?

minad:
No, obviously this won't lead to a rotation.

minad:
Close in favor of #682

YorkZ:
OK, thanks @mattiasdrp. Your PR hasn't been merged yet right?

mattiasdrp:
Not yet, I'm starting my journey through fixing issues in emacs packages and there're a lot of things I don't do right :D

PR #682, opened 2022-11-23
consult-yank-(pop/replace) follows yank-kill-from-rotate

### LINKED ISSUE #668: consult-yank-pop: Support yank-from-kill-ring-rotate

mattiasdrp:
If `yank-from-kill-ring-rotate` is set to t, `consult-yank-(pop/replace)` will put the last yanked value at the top of the candidates by rotating the kill-ring

If the kill-ring is `'(1 2 3 4)` and 3 was the last yanked value, kill-ring-yank-pointer is `'(3 4)`, by appending `(butlast '(1 2 3 4) (length '(3 4)))` to `'(3 4)` we have a new kill-ring that is `'(3 4 1 2)`

This fixes #668 and provides a better solution than #681 (this solution has the default of modifying kill-ring in place)

@minad Sorry for the delay, had some issues and couldn't work on this until now. Thanks for your patience :-)

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -3541,7 +3541,10 @@ If no MODES are specified, use currently active major and minor modes."
   (consult--lookup-member
    (consult--read
     (consult--remove-dups
-     (or kill-ring (user-error "Kill ring is empty")))
+     (or (append
+          kill-ring-yank-pointer
+          (butlast kill-ring (length kill-ring-yank-pointer)))
+         (user-error "Kill ring is empty")))
     :prompt "Yank from kill-ring: "
     :history t ;; disable history
     :sort nil
@@ -3569,6 +3572,11 @@ version supports preview of the selected string."
     (push-mark)
     (insert-for-yank string)
     (setq this-command 'yank)
+    (when yank-from-kill-ring-rotate
+      (let ((pos (seq-position kill-ring string)))
+        (if pos
+            (setq kill-ring-yank-pointer (nthcdr pos kill-ring))
+          (kill-new string))))
     (when (consp arg)
       ;; Swap point and mark like in `yank'.
       (goto-char (prog1 (mark t)


consult.el
```diff
@@ -3541,7 +3541,10 @@ If no MODES are specified, use currently active major and minor modes."
   (consult--lookup-member
    (consult--read
     (consult--remove-dups
-     (or kill-ring (user-error "Kill ring is empty")))
+     (or (append
+          kill-ring-yank-pointer
+          (butlast kill-ring (length kill-ring-yank-pointer)))
+         (user-error "Kill ring is empty")))
```
minad, line 3547: Shouldn't this bei guarded too?

consult.el
```diff
@@ -3569,6 +3572,11 @@ version supports preview of the selected string."
     (push-mark)
     (insert-for-yank string)
     (setq this-command 'yank)
+    (when yank-from-kill-ring-rotate
```
minad, line 3575: yank-from-kill-ring-rotate probably doesn't exist on Emacs 27? I think we have to.introduce a variable consult-yank-rotate.

mattiasdrp: You're right, I'll fix it

minad: On 28 the variable could be a defalias.

mattiasdrp: I'm not sure how to do this though

consult.el
mattiasdrp: Oh yes, you're right. Done

consult.el
oantolin: Maybe it's simpler to have the consult-specific variable have default value `(bound-and-true-p yank-from-kill-ring-rotate)`.

consult.el
```diff
@@ -3530,6 +3530,13 @@ If no MODES are specified, use currently active major and minor modes."
 
 ;;;;; Command: consult-yank
 
+(defcustom consult-yank-from-kill-ring-rotate
+  (if (symbolp yank-from-kill-ring-rotate) yank-from-kill-ring-rotate t)
```
oantolin, line 3534: I'd suggest `(bound-and-true-p yank-from-kill-ring-rotate)` as the default value.

consult.el
mattiasdrp: This means that this feature will not work if emacs version is lower than 28

oantolin: Oh, I got confused I thought the rotation behavior was new in Emacs 28, but what's new is that you can turn it off? In that case, definitely go with your default and not my suggestion.

consult.el
oantolin: But I still have a suggestion: use `boundp` instead of `symbolp`.

mattiasdrp:
Thanks! :-)

minad:
Thanks! Merged with minor changes. Btw, the change was below 15 lines, but for the future - do you have assigned copyright to the FSF for your Emacs contributions?

mattiasdrp:
I don't have any assigned copyright with the FSF. Is it the one that requires you to give your full identity?

minad:
> Is it the one that requires you to give your full identity?

As far as I understand, not publicly, but the FSF must have the ability to associate your identity and your mail address.

mattiasdrp:
I'll look into it then ;-)

PR #691, opened 2022-12-03
obey org-fold-show-context-detail when expanding context

Stebalien:
This makes `consult-org-heading` obey the `org-fold-show-context-detail` setting. Unfortunately, there's no way (as far as I can tell) to distinguish between `consult-org-agenda` and `consult-org-heading` inside this function, so I'm picking `org-goto` and hoping for the best.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -1343,9 +1343,9 @@ ORIG is the original function, HOOKS the arguments."
 (defun consult--invisible-open-permanently ()
   "Open overlays which hide the current line.
 See `isearch-open-necessary-overlays' and `isearch-open-overlay-temporary'."
-  (if (and (derived-mode-p #'org-mode) (fboundp 'org-fold-show-set-visibility))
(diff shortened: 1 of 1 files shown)


Stebalien:
The main downside of this patch is that in the default configuration, it'll `cosnult-org-agenda` function use `ancestors` instead of `local`.

It's also unclear if the defaults are sane, so I'm not going to push for this patch. Unfortunately, the default `ancestors` won't actually show the entry being jumped to.

Stebalien:
Ok, no. This may be slightly more correct, maybe, but it provides a strictly worse user experience out of the box.

minad:
Alternatives:

- Add a `consult-org-fold-context` variable, set to `local` by default.
- Lookup `consult` in `org-fold-show-context-detail` and use 'local if the key doesn't exist.
- Use `isearch` as context

PR #699, opened 2022-12-11
Highlight all matches of each regexp in consult-grep

astoff:
As discussed in https://github.com/oantolin/embark/pull/568.


minad:
Thanks

PR #727, opened 2023-01-25
Draft for consult-info

minad:
Missing features:

- Match highlighting in the candidates list
- Match highlighting during preview
- Better candidate formatting
- Performance is bad, too bad to add this to Consult in the current form.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -4,6 +4,7 @@
 
 * Development
 
+- Add =consult-info= command (#634, #727).
 - =consult-buffer=: Always select the first candidate when narrowing (#714).
 - Drop obsolete =consult-apropos=. Alternative: =describe-symbol= in combination
   with =embark-export=.
--- a/README.org
+++ b/README.org
@@ -334,6 +334,16 @@ their descriptions.
 - =consult-org-agenda=: Jump to an agenda heading. Supports
   narrowing by heading level, priority and TODO state, as well as
   live preview and recursive editing.
+** Help
+:properties:
+:description: Searching through help
+:end:
+
+#+findex: consult-info
+#+findex: consult-man
+- =consult-man=: Find Unix man page, via Unix =apropos= or =man -k=. =consult-man= opens
+  the selected man page using the Emacs =man= command.
+- =consult-info=: Full text search through info pages.
 
 ** Miscellaneous
 :properties:
@@ -342,11 +352,8 @@ their descriptions.
 
 #+findex: consult-completion-in-region
 #+findex: consult-theme
-#+findex: consult-man
 #+findex: consult-preview-at-point
 #+findex: consult-preview-at-point-mode
-- =consult-man=: Find Unix man page, via Unix =apropos= or =man -k=. =consult-man= opens
-  the selected man page using the Emacs =man= command.
 - =consult-theme=: Select a theme and disable all currently enabled themes.
   Supports live preview of the theme while scrolling through the candidates.
 - =consult-preview-at-point= and =consult-preview-at-point-mode=: Command and minor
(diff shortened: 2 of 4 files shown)


hmelman:
Worked ok for me (Emacs 28.2 macport).  One thing for me is that I use GUI emacs and have the following to open info in its' own frame, which I think is not uncommon.  Normally when invoking info from my main frame, a new frame opens with the info buffer.  When doing consult-info a new frame also opens, but point is in the info buffer and not in that frame's minibuffer which would be more convenient since it's in the middle of consult input. Doing C-x o to switch to the minibuffer and everything worked fine.  Not sure if you could guarantee this working in your code or if it would just be some recommendation for people doing what I'm doing.

```elisp
(add-to-list 'display-buffer-alist
             '("\\*info\\*" . ((display-buffer-reuse-window display-buffer-pop-up-frame)
                               (pop-up-frame-parameters (left . 300))
                               (inhibit-same-window . t)
                               (reusable-frames . t))))
```

minad:
@hmelman Thanks, the issue with opening in frames should definitely be fixed.

@alphapapa This command is based on your initial prototype and uses `Info-search`. I must admit that I am not happy with the performance. It seems Helm uses its own search function and parses the info page itself. Is that correct? Maybe we could do the same here to speed up the searching and only use info itself to display/preview the location. Could you try `consult-info` and compare it with Helm (behavior and performance)? Thanks!

hmelman:
I think `consult-info-emacs` should use a variable for the info file names, kinda like `Info-file-list-for-emacs`.

minad:
@hmelman I don't think that's necessary. This command is just a "sample command". I would recommend that you create your own commands based on `consult-info` which allows you to search the info pages you are most interested in.

hmelman:
@minad I'm just thinking that emacs ships with many info files.  I wouldn't expect them all to be an "info-emacs" command but I would expect gnus, calc, ediff, and eshell to be in one with emacs and cl-lib to be in one with elisp.  I'm sure people will disagree on their set to include.  Providing a simple defcustom with even the current list of just "emacs" and "elisp" makes it easier for people to add their own without having to bother creating `my-consult-info-emacs` to not conflict or overwrite the "sample command".  This allows customization via customize as opposed to having to write code.

minad:
> I wouldn't expect them all to be an "info-emacs" command but I would expect gnus, calc, ediff, and eshell to be in one with emacs and cl-lib to be in one with elisp.

I have different expectations. I think at least of gnus and calc as applications which sit on top of Emacs. I don't understand why they are part of Emacs in the first place. They are certainly not for technical reasons.

Maybe I will just remove the sample command instead of wasting time discussing this. But adding 
a defcustom is not a good idea, this is just added code for nothing. It does not add any value. the user can then as well write their own commands in three lines of code.

hmelman:
Fair enough (though I like the command).  The sample would be good in the README.

jdtsmith:
Really great!  Zero interface to learn since it borrows from the rest of consult.  

The one small thing I noticed is that some of the info manuals in the initial list are not files but directories, which causes an error.    Also, often I want to search a few manuals at once (e.g. what the user Emacs manual and Elisp manual have to say about processes).  Leaving aside the interface, is this technically possible, given the current design?

minad:
I don't see directories in my Info manual list on Emacs 29.

jdtsmith:
Looks like a bug in Emacs 28 (including files with "no suffix") that got squashed.  From `Info-read-node-name-2`:

Emacs 28:

```elisp
	  ;; If the file name has no suffix or a standard suffix,
	  ;; include it.
	  (and (or (null (file-name-extension file))
		   (string-match suffix file))
	       ;; But exclude subfiles of split Info files.
	       (not (string-match "-[0-9]+\\'" file))
	       ;; And exclude backup files.
	       (not (string-match "~\\'" file))
	       (push (if string-dir (concat string-dir file) file) names))

```

and Emacs 29:

```elisp
          ;; If the file name has a standard suffix,
          ;; include it (without the suffix).
          (when (and (string-match suffix file)
                     ;; But exclude subfiles of split Info files.
                     (not (string-match "\\.info-[0-9]+" file))
                     ;; And exclude backup files.
                     (not (string-match "~\\'" file)))
            (push (substring file 0 (match-beginning 0))
                  names)))))
```

PR #745, opened 2023-02-18
consult-isearch-history: use cursor-in-echo-area

astoff:
I wasn't aware of `cursor-in-echo-area` when I wrote this function, but it's a much less drastic measure than `inhibit-redisplay` (which makes Emacs temporarily unresponsive to the user).

I'm a bit puzzled by the comments explaining the source of flickering. I don't see any flickering in the mode line on my Emacs 29 with pgtk, even if I disable both let-bindings. The only flickering I can see is in the cursor.

(Also, for the record, when you drop support for Emacs 28 you could rely on `minibuffer-lazy-highlight-setup`. But it's not clear it will simplify `consult-isearch-history` too much.)

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -4111,8 +4111,8 @@ This replaces the current search string if Isearch is active, and
 starts a new Isearch session otherwise."
   (interactive)
   (consult--forbid-minibuffer)
-  (let* ((isearch-message-function 'ignore) ;; Avoid flicker in echo area
-         (inhibit-redisplay t)              ;; Avoid flicker in mode line
+  (let* ((isearch-message-function #'ignore)
+         (cursor-in-echo-area t) ;; Avoid cursor flickering
          (candidates (consult--isearch-history-candidates)))
     (unless isearch-mode (isearch-mode t))
     (with-isearch-suspended


minad:
> I wasn't aware of cursor-in-echo-area when I wrote this function, but it's a much less drastic measure than inhibit-redisplay (which makes Emacs temporarily unresponsive to the user).

Seems like a special variable only introduced for Isearch itself, given that Isearch uses the echo area for input.

> (Also, for the record, when you drop support for Emacs 28 you could rely on minibuffer-lazy-highlight-setup. But it's not clear it will simplify consult-isearch-history too much.)

Dropping Emacs 27 support won't happen any time soon, since supporting old Emacs versions became more effortless thanks to the Compat library. We could add `minibuffer-lazy-highlight-setup` to compat-29.el. I don't think it will offer advantages over the current Consult preview mechanism? I you want to use some new API in packages feel free to propose an addition to Compat.

astoff:
> Dropping Emacs 27 and 28 support won't happen any time soon

I meant this for some long-term TODO list you might keep :-)

> I don't think it will offer advantages over the current Consult preview mechanism?

Right, some of the hard work in the built-in mechanism overlaps with consult's preview.

>  I you want to use some new API in packages feel free to propose an addition to Compat.

I don't think this is so useful, since it's just a visual thing.

minad:
> I meant this for some long-term TODO list you might keep :-)

Without Compat I would have been more inclined to maintain such a list. Dropping support really makes only sense if significant bugs are fixed in newer Emacs versions or if Elisp changes somehow fundamentally, such that a clear cut is absolutely necessary. For example the move to lexical scoping was such a change. Maybe I am overly enthusiastic about Compat, but I think it really helps in making Emacs a truly stable platform for package developers which at the same time is not stuck in the past.

PR #749, opened 2023-02-22
Add consult support for multi-path and wildcard argument handling

rswgnu:
This adds support for searching over multiple directories or files sets for all of the consult grep commands when a prefix argument is given to the command.  Space-separated shell glob-type expressions may be given to produce the path set sent to the grep command; expansion is done with file-expand-wildcards.   See the CHANGELOG.org for detailed updates.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -4,16 +4,13 @@
 
 * Development
 
-Multi-path handling contributed by Robert Weiner (@rswgnu):
-- Add support for grep and find over multiple directories and files.
-- Change =consult--directory-prompt=.  With a prefix arg sent as =dir= to one
-  of the consult search commands:
-    if it is a single C-u, prompt for a single directory to recursively search;
-    with any other prefix arg, prompt for a space-separated string of positive
-    number of directories or files.
-- Rename =consult--git-grep-builder= to =consult--git-grep-make-builder=
-  and return lambda so can pass in =search-path-list=.
-
+- BREAKING: Remove the "." argument from  =consult-grep-args= and
+  =consult-ripgrep-args=, since directories or files to search are appended by the
+  command line builder. Take this change into account, when you use a customized
+  version of those variables.
+- =consult-grep=: Add support for grep and find over multiple files or directory.
+  If the prefix argument DIR is a single C-u, prompt for a directories or files
+  to search recursively via =completing-read-multiple=.
 - =consult-buffer= and =consult-isearch-history=: Align annotations dynamically
   depending on candidate width, instead of computing the alignment beforehand.
 - Add the full path as =help-echo= property to abbreviated directory paths and
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -4,6 +4,16 @@
 
 * Development
 
+Multi-path handling contributed by Robert Weiner (@rswgnu):
+- Add support for grep and find over multiple directories and files.
+- Change =consult--directory-prompt=.  With a prefix arg sent as =dir= to one
+  of the consult search commands:
+    if it is a single C-u, prompt for a single directory to recursively search;
+    with any other prefix arg, prompt for a space-separated string of positive
+    number of directories or files.
+- Rename =consult--git-grep-builder= to =consult--git-grep-make-builder=
+  and return lambda so can pass in =search-path-list=.
+
 - =consult-buffer= and =consult-isearch-history=: Align annotations dynamically
   depending on candidate width, instead of computing the alignment beforehand.
 - Add the full path as =help-echo= property to abbreviated directory paths and
--- a/README.org
+++ b/README.org
@@ -256,9 +256,8 @@ their descriptions.
   completion style like orderless. =consult-grep= supports preview. If the
   =consult-project-function= returns non-nil, =consult-grep= searches the
   current project directory. Otherwise the =default-directory= is searched. If
-  =consult-grep= is invoked with prefix argument =C-u M-s g=, you can specify the
-  directory manually.  With any other prefix argument, =C-u C-u M-s g=, you can
-  enter any positive number of space-separated directories or files to search over.
+  =consult-grep= is invoked with prefix argument =C-u M-s g=, you can specify one or
+  multiple files or directories manually.
 - =consult-find=, =consult-locate=: Find file by matching the path against a regexp.
   Like for =consult-grep=, either the project root or the current directory is the
   root directory for the search. The input string is treated similarly to
@@ -1198,6 +1197,7 @@ Code contributions:
 - [[https://github.com/aagon][Aymeric Agon-Rambosson]]
 - [[https://github.com/geolessel][Geoffrey Lessel]]
 - [[https://github.com/piotrkwiecinski][Piotr Kwiecinski]]
+- [[https://github.com/rswgnu][Robert Weiner]]
 
 Advice and useful discussions:
 - [[https://github.com/clemera/][Clemens Radermacher]]
--- a/README.org
+++ b/README.org
@@ -257,12 +257,14 @@ their descriptions.
   =consult-project-function= returns non-nil, =consult-grep= searches the
   current project directory. Otherwise the =default-directory= is searched. If
   =consult-grep= is invoked with prefix argument =C-u M-s g=, you can specify the
-  directory manually.
+  directory manually.  With any other prefix argument, =C-u C-u M-s g=, you can
+  enter any positive number of space-separated directories or files to search over.
 - =consult-find=, =consult-locate=: Find file by matching the path against a regexp.
   Like for =consult-grep=, either the project root or the current directory is the
   root directory for the search. The input string is treated similarly to
   =consult-grep=, where the first part is passed to find, and the second part is
-  used for Emacs filtering.
+  used for Emacs filtering.  Prefix arguments to =consult-find= work just like those
+  for the consult grep commands.
 
 ** Compilation
 :properties:
--- a/consult.el
+++ b/consult.el
@@ -451,7 +451,7 @@ Used by `consult-completion-in-region', `consult-yank' and `consult-history'.")
      :foreground "#333"))
   "Face used for thin line separators in `consult-register-window'.")
 
-;;;; History variables
+;;;; Input history variables
 
 (defvar consult--keep-lines-history nil)
 (defvar consult--grep-history nil)
@@ -580,6 +580,26 @@ Turn ARG into a list, and for each element either:
                 (split-string-and-unquote x)
               (ensure-list (eval x 'lexical))))
           (ensure-list arg)))
+
+(defun consult--build-args-with-paths (arg search-path-list)
+  "Return ARG and SEARCH-PATH-LIST joined as a flat list of split strings.
+
+Turn ARG into a list, and for each element either:
+- split it if it a string.
+- eval it if it is an expression.
+
+If SEARCH-PATH-LIST is non-nil and not equal to (\".\"),
+splice its paths in at the end of the arg list."
+  (let ((args (consult--build-args arg)))
+    (if (and search-path-list
+             (not (equal search-path-list '("."))))
+        (if (member "." args)
+            ;; Replace occurrences of "." with `search-path-list' in `args'
+            (flatten-list
+             (mapcar (lambda (elt) (if (equal elt ".") search-path-list elt))
+                     args))
+          (nconc args search-path-list))
+      args)))
 
 (defun consult--command-split (str)
   "Return command argument and options list given input STR."
@@ -723,7 +743,7 @@ The line beginning/ending BEG/END is bound in BODY."
         adir))))
 
 (defun consult--directory-prompt (prompt dir)
-  "Return prompt and directory.
+  "Return a list of latest PROMPT, expanded DIR and a list of paths to search.
 
 PROMPT is the prompt prefix.  The directory
 is appended to the prompt prefix.  For projects
@@ -731,27 +751,63 @@ only the project name is shown.  The `default-directory'
 is not shown.  Other directories are abbreviated and
 only the last two path components are shown.
 
-If DIR is a string, it is returned.
-If DIR is a true value, the user is asked.
-Then the `consult-project-function' is tried.
-Otherwise the `default-directory' is returned."
-  (let* ((dir
+If DIR is null or equal to `default-directory', local
+`search-path-list' is set to nil.  Otherwise, when DIR
+is a string, it is split into space-separated path
+patterns and a flattened list of the results is saved
+in `search-path-list'.
+
+The last element of the return value is either the
+expanded value of DIR if it is a single directory, the
+current project root directory if any from a call to
+`consult-project-function' or else `default-directory'."
+  (let* ((abbrev-dir ".")
+         (possible-prefix-arg dir)
+         (dir
           (cond
-           ((stringp dir) dir)
+           ((or (stringp dir) (and (listp dir) (stringp (car dir))))
+            dir)
            (dir
             ;; Preserve this-command across `read-directory-name' call,
             ;; such that `consult-customize' continues to work.
             (let ((this-command this-command))
-              (read-directory-name "Directory: " nil nil t)))
+              (pcase dir
+                ;; Single C-u prefix arg, prompt for a single directory
+                ('(4) (read-directory-name "Directory: " nil nil t))
+                ;; Any other prefix arg, prompt for any number of existing
+                ;; directories or files
+                (_    (let ((crm-separator " "))
+			(completing-read-multiple "Search Locations (default = .): "
+						  #'completion-file-name-table nil t nil nil abbrev-dir))))))
            (t (or (consult--project-root) default-directory))))
-         (edir (file-name-as-directory (expand-file-name dir)))
+         (edir (file-name-as-directory
+                (expand-file-name (if (and (stringp dir)
+                                           (file-directory-p dir))
+                                      dir
+                                    abbrev-dir))))
          ;; Bind default-directory in order to find the project
-         (pdir (let ((default-directory edir)) (consult--project-root))))
-    (cons
-     (format "%s (%s): " prompt
-             (if (equal edir pdir)
-                 (concat "Project " (consult--project-name pdir))
-               (consult--abbreviate-directory dir)))
+         (pdir (let ((default-directory edir)) (consult--project-root)))
+         (search-path-list
+          (if (equal dir default-directory)
+              (list abbrev-dir)
+            (when (stringp dir)
+              ;; Convert space-separated paths to a list
+              (split-string dir)))))
+    (list
+     (format "%s %s: " prompt
+             (if (and possible-prefix-arg (not (stringp possible-prefix-arg))
+                      (not (equal possible-prefix-arg '(4))))
+                 (if (listp dir)
+                     (mapcar #'consult--abbreviate-directory dir)
+                   (format "(%s)" (consult--abbreviate-directory dir)))
+               (format "(%s)"
+		       (if (equal edir pdir)
+			   (concat "Project " (consult--project-name pdir))
+			 (consult--abbreviate-directory
+			  (if (and (stringp dir) (file-directory-p dir))
+                              dir
+                            abbrev-dir))))))
+     search-path-list
      edir)))
 
 (defun consult--default-project-function (may-prompt)
@@ -2141,6 +2197,7 @@ PROPS are optional properties passed to `make-process'."
              (when args
                (funcall async 'indicator 'running)
                (consult--async-log "consult--async-process started %S\n" args)
+               ;; `args' is a list of the command-line args run by consult for this command
                (setq count 0
                      proc-buf (generate-new-buffer " *consult-async-stderr*")
                      proc (apply #'make-process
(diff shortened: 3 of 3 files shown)


minad:
Thanks. I have to try this out. In the meantime I will add some superficial comments to the code.

consult.el
```diff
@@ -451,7 +451,17 @@ Used by `consult-completion-in-region', `consult-yank' and `consult-history'.")
...
-;;;; History variables
+;;;; Command argument variables (public)
+
+(defvar consult-last-command-args nil
+  "List of args used in prior consult command.
+Automatically set in `consult--async-process'.")
+
+(defvar consult-this-command-args nil
+  "List of args used in current consult command.
+Automatically set in `consult--async-process'.")
```
minad, line 462: These global variables should not be necessary. Please remove them.

consult.el
```diff
@@ -465,6 +475,11 @@ Used by `consult-completion-in-region', `consult-yank' and `consult-history'.")
 
 ;;;; Internal variables
 
+(defvar consult--search-path-list nil
+  "List of paths passed to consult search commands.
+If any directories are included in the list, the first element
+must be a directory.")
+
```
minad, line 482: Same here. No additional global variables.

consult.el
```diff
@@ -546,6 +561,11 @@ We use invalid characters outside the Unicode range.")
 
 ;;;; Miscellaneous helper functions
 
+(defun consult--replace-in-list (from to list)
+  "Replace occurrences of FROM with TO in LIST."
+  (mapcar (lambda (elt) (if (equal elt from) to elt))
+	  list))
```
minad, line 567: Inline this helper for now. If we need it more than once we can still extract it.

consult.el
```diff
@@ -574,12 +594,24 @@ We use invalid characters outside the Unicode range.")
...
+				     (ensure-list arg)))))
+    (if (and consult--search-path-list
+	     (not (equal consult--search-path-list '("."))))
+	(if (member "." args)
+	    (progn (unless (and consult--search-path-list
+				(file-directory-p (car consult--search-path-list)))
+		     (setq args (delete "-r" args)))
+		   (flatten-list (consult--replace-in-list "." consult--search-path-list args)))
+	  (nconc args consult--search-path-list))
+      args)))
```
minad, line 614: `consult--build-args` is not the correct place to add this code, since this function should be generic for all commands. It should rather go to the different command builders, `consult--grep-make-builder` etc. These builders must be reworked anyway to avoid the global variables.

consult.el
```diff
@@ -723,35 +755,55 @@ The line beginning/ending BEG/END is bound in BODY."
...
           (cond
            ((stringp dir) dir)
            (dir
             ;; Preserve this-command across `read-directory-name' call,
             ;; such that `consult-customize' continues to work.
             (let ((this-command this-command))
-              (read-directory-name "Directory: " nil nil t)))
+              ;; (read-directory-name "Directory: " nil nil t)
+	      (read-string "Search Locations (default = .): "
+			   nil nil abbrev-dir)))
```
minad, line 787: This change is not acceptable, since you introduce a regression for existing behavior. `read-string` is worse than `read-directory-name`. Do you have a better suggestion on how we could switch to reading multiple directories? One can use `completing-read-multiple` with a file completion table and a directory predicate, but this is generally a bit brittle.

consult.el
rswgnu: You are right, this was a separate feature.  I thought it would be useful to be able to quickly manipulate the actual command-line arguments used as I found it hard to see this information when initially working with the code.  I patterned it after this-command and last-command in Emacs as it can serve a similar function.

rswgnu:
> EDIT: After a first look I am happy with this addition. It needs some stylistic improvements though. I want to avoid introducing additional global variables. Instead we should adjust the command builders and the grep commands to take multiple directory arguments. For example `consult-grep` currently takes a DIR argument, which could be generalized to take a list, which is then expanded. Then each of the command builders which accepts multiple directories must be adjusted, e.g., `consult--grep-make-builder`.

I used the global variable to avoid repeating the code in multiple builder commands.  That seemed the cleanest to me.  I think the multiple paths can work for any builder command that calls it, though some of the other commands could use some more testing.  Do you like the idea of centralizing the change?  Do you know of a command for which it is not appropriate.  I am fine if you want to integrate the code and then adjust to your style/standards though.

minad:
No, I will not accept the change with the global variables, since it is not a clean addition. The idea of "centralizing the change" is wrong, since of course there are commands for which it does not work correctly. The points I made are not up to debate. Sorry for the strictness, but this is the only way I can ensure high quality and long-term maintainability.

rswgnu:
I get it.  I will work on the updates you have requested when I get some more free time.  Thanks for the clear feedback.

minad:
Thanks, Robert! No need to rush this addition. We will add it when it is ready. :)

rswgnu:
On Wed, Feb 22, 2023 at 1:40 AM Daniel Mendler ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In consult.el
> <https://github.com/minad/consult/pull/749#discussion_r1113894359>:
>
> >            (cond
>             ((stringp dir) dir)
>             (dir
>              ;; Preserve this-command across `read-directory-name' call,
>              ;; such that `consult-customize' continues to work.
>              (let ((this-command this-command))
> -              (read-directory-name "Directory: " nil nil t)))
> +              ;; (read-directory-name "Directory: " nil nil t)
> +	      (read-string "Search Locations (default = .): "
> +			   nil nil abbrev-dir)))
>
> This change is not acceptable, since you introduce a regression for
> existing behavior. read-string is worse than read-directory-name. Do you
> have a better suggestion on how we could switch to reading multiple
> directories? One can use completing-read-multiple with a file completion
> table and a directory predicate, but this is generally a bit brittle.
>
> —
>
Hi Daniel:

I have implemented all of your other suggested changes except this one.  I
see your point, but the issue is that the new code allows for multiple
arguments in the string, none of which need to be directories.  Any of the
arguments can be wildcarded file sets for use with grep or find commands.
For example, the equivalent of "grep my-regex ~/code/*.el *.c" or "find
*.el -name *info* -print".

What if we left a C-u prefix argument to commands so it still prompted for
a single directory but if you used any other prefix argument, it allowed
for reading the multiple, possibly non-directory arguments using
read-string?  If we find an acceptable strategy, I can finish this up
tomorrow.

minad:
On 2/23/23 08:32, Robert Weiner wrote:
> I have implemented all of your other suggested changes except this one. I
> see your point, but the issue is that the new code allows for multiple
> arguments in the string, none of which need to be directories. Any of the
> arguments can be wildcarded file sets for use with grep or find commands.
> For example, the equivalent of "grep my-regex ~/code/*.el *.c" or "find
> *.el -name *info* -print".

Thanks, Robert! That's great. We have to think which approach will
result in the best UI. I see multiple possibilities:

1. Use completing-read-multiple.
2. Use C-u for a single directory and C-u C-u for multiple wildcards.
3. Make it configurable, introduce a consult-read-directories-function
and two implementations.
4. Bind some command in the minibuffer-local-map for read-directory
which allows you to switch to read-string to enter wildcards.

I am really not sure...

> What if we left a C-u prefix argument to commands so it still prompted for
> a single directory but if you used any other prefix argument, it allowed
> for reading the multiple, possibly non-directory arguments using
> read-string? If we find an acceptable strategy, I can finish this up
> tomorrow.

Yes, that would be solution 2 from above. I suggest you go with that
solution for now. When I try it out I will check if that works well
enough. We will probably have to iterate again anyway and document
everything in the end. As I wrote, no rush. I am hopeful that we end up
with something reasonable.

rswgnu:
Good ideas.  2, 3, and 4 all sound useful to me.  Will make an update and then you can try it out as you said.  -- BobOn Feb 23, 2023, at 3:21 AM, Daniel Mendler ***@***.***> wrote:﻿
On 2/23/23 08:32, Robert Weiner wrote:
> I have implemented all of your other suggested changes except this one. I
> see your point, but the issue is that the new code allows for multiple
> arguments in the string, none of which need to be directories. Any of the
> arguments can be wildcarded file sets for use with grep or find commands.
> For example, the equivalent of "grep my-regex ~/code/*.el *.c" or "find
> *.el -name *info* -print".

Thanks, Robert! That's great. We have to think which approach will
result in the best UI. I see multiple possibilities:

1. Use completing-read-multiple.
2. Use C-u for a single directory and C-u C-u for multiple wildcards.
3. Make it configurable, introduce a consult-read-directories-function
and two implementations.
4. Bind some command in the minibuffer-local-map for read-directory
which allows you to switch to read-string to enter wildcards.

I am really not sure...

> What if we left a C-u prefix argument to commands so it still prompted for
> a single directory but if you used any other prefix argument, it allowed
> for reading the multiple, possibly non-directory arguments using
> read-string? If we find an acceptable strategy, I can finish this up
> tomorrow.

Yes, that would be solution 2 from above. I suggest you go with that
solution for now. When I try it out I will check if that works well
enough. We will probably have to iterate again anyway and document
everything in the end. As I wrote, no rush. I am hopeful that we end up
with something reasonable.

rswgnu:
Added my multi-path updates addressing your prior feedback.

consult.el
```diff
@@ -576,10 +576,37 @@ Turn ARG into a list, and for each element either:
...
+- split it if it a string.
+- eval it if it is an expression.
+
+If SEARCH-PATH-LIST is non-nil and not equal to '(\".\"),
+splice its paths in at the end of the arg list."
+  (let ((args (copy-sequence (mapcan (lambda (x)
+				       (if (stringp x)
+					   (split-string-and-unquote x)
+					 (ensure-list (eval x 'lexical))))
+				     (ensure-list arg)))))
```
minad, line 597: The copy-sequence is unnecessary? Also you can reuse `consult--build-args` here.

consult.el
```diff
@@ -4763,38 +4829,52 @@ the directory to search in.  By default the project directory is used
...
 Otherwise the `default-directory' is searched."
   (interactive "P")
-  (consult--grep "Grep" (consult--grep-make-builder) dir initial))
+  ;; Must call `consult--directory-prompt' to set local `search-path-list'
+  ;; prior to calling `consult--grep-make-builder' where the path list is used.
+  (let* ((prompt-paths-dir (consult--directory-prompt "Grep" dir))
+	 (prompt (nth 0 prompt-paths-dir))
+	 (search-path-list (nth 1 prompt-paths-dir)))
+    (setq dir (nth 2 prompt-paths-dir))
+    (consult--grep prompt (consult--grep-make-builder search-path-list) dir initial)))
```
minad, line 4838: It seems you replicated this code three times, for grep, git-grep and ripgrep. Instead you can move it into consult--grep and change consult--grep to take a MAKE-BUILDER argument instead of a BUILDER argument.

Furthermore I think we should reuse the DIR argument here. DIR can either be a single string which is then used as default-directory and searched or a list of strings for the wildcard handling.

consult.el
```diff
@@ -4680,13 +4747,12 @@ Take the variables `grep-find-ignored-directories' and
 BUILDER is the command line builder function.
 PROMPT is the prompt string.
 INITIAL is inital input."
-  (let* ((prompt-dir (consult--directory-prompt prompt dir))
-         (default-directory (cdr prompt-dir)))
+  (let* ((default-directory dir))
```
minad, line 4750: consult--directory-prompt should still be called here, see my comment https://github.com/minad/consult/pull/749/files#r1116595189.

consult.el
```diff
@@ -576,10 +576,37 @@ Turn ARG into a list, and for each element either:
...
+				       (if (stringp x)
+					   (split-string-and-unquote x)
+					 (ensure-list (eval x 'lexical))))
+				     (ensure-list arg)))))
+    (if (and search-path-list
+	     (not (equal search-path-list '("."))))
+	(if (member "." args)
+	    (progn (unless (and search-path-list
+				(file-directory-p (car search-path-list)))
+		     (setq args (delete "-r" args)))
```
minad, line 603: Why remove the -r argument here? I think you can always specify -r even if you only pass a list of files. The check if the first argument is a directory is brittle and will break if the user specifies the directories/wildcards in a different order.

CHANGELOG.org
```diff
@@ -4,6 +4,16 @@
 
 * Development
 
+Multi-path and wildcard handling contributed by Robert Weiner (@rswgnu):
+- Add support for grep and find over multiple directories and wildcarded paths.
+- Change =consult--directory-prompt=.  With a prefix arg sent as =dir= to one
+  of the consult search commands:
+    if it is a single C-u, prompt for a single directory to recursively search;
+    with any other prefix arg, prompt for a space-separated string of any
+    number of directories followed by file paths optionally with wildcards.
```
minad, line 13: I don't like the restriction that directories have to come before a list of wildcards. This restriction seems unnecessary and arbitrary. See also https://github.com/minad/consult/pull/749/files#r1116602923

README.org
```diff
@@ -257,12 +257,15 @@ their descriptions.
   =consult-project-function= returns non-nil, =consult-grep= searches the
   current project directory. Otherwise the =default-directory= is searched. If
   =consult-grep= is invoked with prefix argument =C-u M-s g=, you can specify the
-  directory manually.
+  directory manually.  With any other prefix argument, =C-u C-u M-s g=, you can
+  enter any number (including 0) space-separated directories followed by any number
+  of wildcarded paths, like "~/src/*.c" to search over.
```
minad, line 262: Also here, we should get rid of the restriction.

minad:
@rswgnu Thanks for the changes! I added a few comments. Please check them out.

consult.el
```diff
@@ -576,10 +576,37 @@ Turn ARG into a list, and for each element either:
...
 - eval it if it is an expression."
   (mapcan (lambda (x)
-            (if (stringp x)
-                (split-string-and-unquote x)
-              (ensure-list (eval x 'lexical))))
-          (ensure-list arg)))
+	    (if (stringp x)
+		(split-string-and-unquote x)
+	      (ensure-list (eval x 'lexical))))
+	  (ensure-list arg)))
```
minad, line 582: (Minor issue - can you try to avoid such whitespace changes please?)

consult.el
```diff
@@ -723,35 +750,74 @@ The line beginning/ending BEG/END is bound in BODY."
...
+		    (if (listp dir)
+			dir
+		      (format "(%s)" dir)))
+		   ((equal edir pdir)
+                    (format "(Project %s)" (consult--project-name pdir)))
+		   (t (format "(%s)"
+			      (consult--abbreviate-directory
+			       (if (and (stringp dir) (file-directory-p dir))
+				   dir
+				 abbrev-dir))))))
```
minad, line 819: I must admit I don't understand what you are doing here. What do you want to achieve? How should the prompt look for directory search, project search and wildcard search? I propose the following:

- `Grep (Project consult): `
- `Grep (some/abbreviated/dir): `
- `Grep (first/abbreviated/unexpanded/wildcard, …): `

rswgnu:
Thanks, will work on these.On Feb 24, 2023, at 2:49 AM, Daniel Mendler ***@***.***> wrote:﻿
@minad commented on this pull request.



In consult.el:
> +	       (mapcar (lambda (pat) (or (file-expand-wildcards pat) pat))
+		       (split-string dir)))))))
+    (list
+     (format "%s %s: " prompt
+	     (cond ((and possible-prefix-arg (not (stringp possible-prefix-arg))
+			 (not (eq possible-prefix-arg '(4))))
+		    (if (listp dir)
+			dir
+		      (format "(%s)" dir)))
+		   ((equal edir pdir)
+                    (format "(Project %s)" (consult--project-name pdir)))
+		   (t (format "(%s)"
+			      (consult--abbreviate-directory
+			       (if (and (stringp dir) (file-directory-p dir))
+				   dir
+				 abbrev-dir))))))

I must admit I don't understand what you are doing here. What do you want to achieve? How should the prompt look for directory search, project search and wildcard search? I propose the following:

Grep (Project consult): 
Grep (some/abbreviated/dir): 
Grep (first/abbreviated/unexpanded/wildcard, …):

consult.el
rswgnu: Next push will resolve these two.

consult.el
rswgnu: You are right.  Next push will stop removing the -r.

consult.el
rswgnu: Should be resolved in next push.

consult.el
rswgnu: This should be resolved with next push.

consult.el
rswgnu: Added a .dir-locals.el but won't commit it.

consult.el
rswgnu: I was looking for and getting pretty much what you are showing except that the I did not deal with abbreviating all the paths.  I'll look at it.

rswgnu: Did my best but the conditional logic here requires some complexity.

CHANGELOG.org
rswgnu: I think this was just related to removal of the the -r flag, which we have now done away with, so this restriction should no longer hold after the next push.

README.org
rswgnu: Yes, handled.

rswgnu:
Latest update to address all of your issues has been pushed.

minad:
Okay, I tried this. I have identified a problem. Since we expand the wildcards the whole procedure can become quite inefficient if we end up with a large list of files. It is not a good idea to pass thousands of command line arguments to the grep search process. Therefore we should only allow a list of directories to search. Is this also sufficient for your use case? In order to restrict the file extensions to search one can pass wildcards directly. See for example the `--glob` argument of ripgrep.

consult.el
minad: I fiddled a bit around with this and it is possible to reduce the conditional logic significantly. But this does not matter much. We should first discuss the wildcard expansion issue, I mentioned in https://github.com/minad/consult/pull/749#issuecomment-1445043576.

minad:
Another small nit: Can you please enable [package-lint](https://github.com/purcell/package-lint) in your Emacs configuration such that you introduce no new warnings? There are a few new warnings here due to long docstrings and missing argument documentation.

rswgnu:
> Okay, I tried this. I have identified a problem. Since we expand the wildcards the whole procedure can become quite inefficient if we end up with a large list of files. It is not a good idea to pass thousands of command line arguments to the grep search process. This problem may not matter much for most of the use cases, but the Consult commands are also meant to scale well and handle the edge cases. Therefore we should only allow a list of directories to search. Then we could even use `completing-read-multiple` to read the directories. Is this also sufficient for your use case? In order to restrict the file extensions to search one can pass wildcards directly. See for example the `--glob` argument of ripgrep.

I had thought about this and whether it was worth it to use xargs to pass any wildcard args but thought the complexity therein would not be worth it.  The user should have an idea of the fileset he is looking for and not go overboard, after all he could just as well try to enter hundreds of directories too.  If there is an error in the worst case, now it will be displayed.  If you wanted, we could display the number of search locations after expansion to give another hint as to what is going on.

My original use case does require a list of files and as you have noted being able to pass in a list of files when called non-interactively is a useful feature.  I am adding a single function to Hyperbole's HyRolo to interface it to consult.  HyRolo uses  a single variable, hyrolo-file-list, that is a list of files to search over and using consult-grep on that allows for interactive full-text search.

rswgnu:
> Another small nit: Can you please enable [package-lint](https://github.com/purcell/package-lint) in your Emacs configuration such that you introduce no new warnings? There are a few new warnings here due to long docstrings and missing argument documentation.

Will do.

rswgnu:
The missing argument documentation seems to be from your code.

I thought I had eliminated doc lines greater than 80 characters and am not
seeing that error with the linter in Emacs 27.1 but I will look again.

minad:
> The missing argument documentation seems to be from your code.

My code has a few (unfixable) warnings but please check that you don't add new ones.

> The user should have an idea of the fileset he is looking for and not go overboard, after all he could just as well try to enter hundreds of directories too.

This is not true if we remove the wildcard expansion. Manually specifying hundreds of files or directories doesn't happen accidentally. I only worry about the wildcard expansion and the resulting potential explosion. We should remove that and just allow an unexpanded list of multiple directories/files.

> My original use case does require a list of files and as you have noted being able to pass in a list of files when called non-interactively is a useful feature. I am adding a single function to Hyperbole's HyRolo to interface it to consult.

This use case should still work even if we don't expand. It will still be possible to search multiple files and directories.

rswgnu:
I would really like to be able to utilize wildcards.  The utility of doing
this in directories with a limited number of files greatly outweighs the
rare edge case usage where someone adds a directory with thousands of files.

But it is your package and if you are set on that not being a part of the
change, I will remove it.  Just think on it a bit and let me know your
decision.  Thanks.

rswgnu:
To limit usage of wildcards further, we could utilize the user config
setting of 'find-file-wildcards' and only do expansion if that were
non-nil, i.e. someone specifically says they want it.

minad:
I would prefer if we remove it for now, going more slowly and being conservative by only allowing files/directories. My problem with expanded wildcards is that they are the wrong tool. If you really want to search across files of only a specific type you either have to combine find and grep via xargs or use the --glob argument of ripgrep. This would be the technically better solution but of course much more complicated. I don't want to add anything to Consult where I have the nagging feeling that the approach is not the right one. This will only create headaches down the road.

rswgnu:
Ok.

rswgnu:
Updates made.  Wildcards were removed, linter warnings removed and I added use of completing-read-multiple so you still get completion when use the C-u C-u prefix to prompt for multiple elements.

rswgnu:
Now that we have removed wildcards and allow for just directories and files with completion enabled, could we move this capability from C-u C-u to just a single C-u prefix argument so more users see it and can make use of it?

oantolin:
Doesn't a single C-u prompt for a single directory already? If C-u C-u no longer does wildcards, then the two cases should probably be merged and out under a single C-u.

minad:
@oantolin @rswgnu Yes, that's a nice and simple generalization.

oantolin:
As for globs, isn't the current ability to pass commandline arguments enough? I've used GNU grep's `--include` and `--exclude` and ripgrep's `--glob` and `--type` arguments. Those seem pretty convenient. I don't need a special UI for them, I'm OK with typing the arguments at the main `consult-grep` or `consult-ripgrep` prompt.

minad:
@oantolin

> As for globs, isn't the current ability to pass commandline arguments enough? I've used GNU grep's --include and --exclude and ripgrep's --glob and --type arguments. Those seem pretty convenient. I don't need a special UI for them, I'm OK with typing the arguments at the main consult-grep or consult-ripgrep prompt.

Exactly, this is sufficient and more efficient. Therefore I didn't want to add the wildcards via the prior completion step.

minad:
Refactored and merged. Please test if it works for you as expected.

rswgnu:
Great.  Will give it a go and let you know how it goes.On Feb 25, 2023, at 5:21 PM, Daniel Mendler ***@***.***> wrote:﻿
Refactored and merged. Please test if it works for you as expected.

aaronjensen:
@minad my primary use case for this was a single directory. I used vertico to select that directory. Very often, I was going back to a parent directory and would just hit backspace a few times. Now I get this:

CleanShot 2023-02-27 at 22 18 04@2x

I don't imagine there is a way to configure vertico to work with backspace for multi directory selection, so is my best bet to invoke completing read myself to select a directory and pass that to `consult-ripgrep`?

minad:
@aaronjensen Please check if https://github.com/minad/consult/commit/bd0fce89fffba3b2ad803f4a0de799a01efb71f5 leads to a better behavior for you.

aaronjensen:
Ah, yes, that works, thank you @minad. FWIW, it makes backspace behave funny in vertico if you try to multi-select -- a backspace will delete past the `,`.

minad:
@aaronjensen We should probably adapt the vertico-directory commands to take the completion boundaries into account.

PR #753, opened 2023-02-26
Grep multi path - small doc improvements

rswgnu:


rswgnu:
I made some small doc improvements you can see at:  https://github.com/minad/consult/pull/753/files

minad:
I added your doc fixes, but didn't merge since your PR could not be merged cleanly. In the future, please rebase that we end up with a single commit for such minor changes. 

The prompt improvement is questionable. This is a notable usability flaw of `completing-read-multiple` (CRM) that the user is not informed that they are in a CRM session and they are not informed about the separator. See https://github.com/minad/vertico#configuration for an advice to fix this.

A better fix would be to report upstream that CRM must be improved. It would be better if `crm-separator` could be set to a pair:

~~~ elisp
(setq crm-separator '("\\s-*,\\s-*" . "comma"))
~~~

Then CRM could display the prompt "Foo (separated by comma): " when invoked with `(completing-read-multiple "Foo: " table)`.

rswgnu:
But what is the harm in telling the user now in one place where we know what the separator is?  It will prevent some confusion.

I had wanted to allow space as at least one of the separators but will see if not doing so allows for easy use of filenames with spaces as Windows users often have.

minad:
Well, we don't know the separator. The user can also modify crm-separator. Space is not a good idea because of windows file names and because of Orderless.

rswgnu:
Good point.  It is a simple variable and not a user option, however.  What
if we add a let that sets it to the known crm-default-separator value just
for this use case so we are sure and behavior matches the doc?

minad:
No, I think the current behavior is better. Users who adjust crm-separator will know.

PR #773, opened 2023-03-24
Add list of faces used to highlight previews.

muirdm:
This allows for compatibility with orderless faces. For example, `(setq consult-preview-match-faces orderless-match-faces)` will make preview highlighting faces match that of your orderless faces.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -380,6 +380,12 @@ Used for example in `consult-grep'.")
   '((t :inherit isearch))
   "Face used for match previews, e.g., in `consult-line'.")
 
+(defcustom consult-preview-match-faces
+  [consult-preview-match]
+  "Faces used for preview matches.
+This allows compatibility with `orderless-match-faces', if desired."
+  :type '(vector face))
+
 (defface consult-preview-cursor
   '((t :inherit cursor))
   "Face used for cursor previews and marks, e.g., in `consult-mark'.")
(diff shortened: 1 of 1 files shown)


muirdm:
I realized a better fix may be to propagate the faces from the candidate rather than only the ranges that have a face. Then we wouldn't need extra config. Thoughts?

minad:
Admittedly I am not fond of this change. Currently matches within the buffer are styled like Isearch matches and stand out, which is not necessarily the case for the Orderless faces. The design is intentionally as is.

muirdm:
Note that this commit doesn't change the default behavior.

My orderless faces stand out, so having the preview styled consistently is a strict improvement for me. I like the visual feedback of which part of the pattern is matching where.

minad:
> Note that this commit doesn't change the default behavior.

Sure, this can be said about every change guarded by a configuration variable.

> My orderless faces stand out, so having the preview styled consistently is a strict improvement for me. I like the visual feedback of which part of the pattern is matching where.

I understand your point, but I want to avoid adding configuration variables all over the place which imho only lead to marginal improvements. I have to keep the maintainability of my packages in mind. There is simply no clear line on where to stop when adding new configuration variables. I wouldn't mind if we deconstruct the function `consult--jump-preview` function, e.g., extracting a new function `consult--add-match-overlays` which you could then advise in your configuration.

muirdm:
I will revisit after I'm done reconciling my current workflow with consult.

Issue #281, opened 2021-04-20
[Question] How can I use `consult-line` together with `visual-line-mode`?


apc:
Is there a way to have `consult-line` play nice with `visual-line-mode`? 

This is what I get using `consult-line` with `global-visual-line` turned on: 

<img width="1278" alt="image" src="https://user-images.githubusercontent.com/350504/115449170-7f96b580-a1e8-11eb-8c96-6ad90dad66c9.png">

Compare to the result of using `swiper`, where the previews in the minibuffer don't show the beginning of the line, but rather the context where the match occurs:

<img width="1277" alt="image" src="https://user-images.githubusercontent.com/350504/115449255-9937fd00-a1e8-11eb-8d54-42aab6116879.png">

Perhaps there's an alternative to `counsel-line` that works much like `swiper` does?

minad:
This is actually a feature request. Consult does not handle visual lines as of now. There are two possibilities.

1. We could handle this on the side of Consult by splitting logical lines into visual lines. I am not sure how this works internally - if it is easily possible to move across all visual lines of a buffer.

* consult-line (split logical lines in visual lines)
* consult-mark (show only visual line around the mark)
* consult-global-mark (show only visual line around the mark)
* consult-outline (show only visual line of the heading which appears at the beginning of the logical line)

2. Another way to handle the issue would be to ensure on the side of the completion UI that the current match position is visible. Selectrum for example already does something like this for multi-line candidate strings, but not (yet?) for long single lines.  

Disadvantage of approach 1 is that the candidates are less meaningful, we are not showing something "logical" anymore. The same issue applies to `consult-yank-pop` for example which also uses very long candidate strings. In that case one could also split up the kill ring candidates for a better display in visual substrings of the logical kill ring string. Disadvantage of the second approach is that every completion UI must do such special casing then for long candidates as Selectrum does for multi-line candidates.

minad:
I just checked swiper and it uses either a flag or a predicate to check if visual line movement should be used. As I suspected movement over visual lines is slow, so it is not actually easily possible to use option 1 unless one accepts that the command will be unacceptably slow for large buffers. One could use a predicate like swiper but this is not a particularly compelling solution.

~~~ elisp
(defvar swiper-use-visual-line-p
  (lambda (n-lines)
    (and visual-line-mode
         ;; super-slow otherwise
         (< (buffer-size) 20000)
         (< n-lines 400)))
  "A predicate that decides whether `line-move' or `forward-line' is used.
[...]

PR #774, opened 2023-03-25
Consult line preview all matches

muirdm:
Consult

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -182,6 +182,10 @@ This setting affects the command `consult-line' and the `consult-grep' variants.
 See also `display-line-numbers-widen'."
   :type 'boolean)
 
+(defcustom consult-line-preview-all-matches nil
+  "Highlight matches in other lines."
+  :type 'boolean)
+
 (defcustom consult-goto-line-numbers t
   "Show line numbers for `consult-goto-line'."
   :type 'boolean)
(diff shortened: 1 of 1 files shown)


consult.el
```diff
@@ -3259,14 +3278,41 @@ IGNORED-FACES are ignored when determining the match position."
...
+    (setq candidates (list (car (member selected candidates)))))
+
+  (let ((candidates (consult--completion-filter
+                     input
+                     (mapcar (lambda (cand)
+                               (propertize
+                                (substring-no-properties cand)
+                                'consult-location (get-text-property 0 'consult-location cand)))
+                               candidates)
+                     'consult-location 'highlight)))
```
minad, line 3297: Thanks for this!

A quick comment about performance - I am unlikely to accept this patch in this form since an entire list of completion candidates and all the candidate strings (!) are allocated here on every key press for preview. This will be too costly for the Emacs GC.

Also note that I am not happy to add this special casing only for consult-line/consult-outline. There are also commands like consult-grep, consult-line-multi and consult-info which would need a similar treatment.

muirdm: Yes, this function is implemented in an awkward spot. Having to strip properties before passing candidates through the matcher is problematic. I wanted to first cheaply filter to visible candidates, but that can't happen here for multiple reasons.

The whole approach of passing the lines through the matcher (again) to find the highlight ranges seems very roundabout. Even the current implementation is `O(total candidates)` which conceptually doesn't seem necessary.

Would something like this be possible?
1. Somehow hook the minibuffer completion matching to record the match ranges as an explicit string property.
1. Pass this function a list of already filtered instead of all candidates (optional - could use above match range metadata to efficiently filter).
1. Propagate match ranges without having to do any real work.

minad: > Even the current implementation is O(total candidates) which conceptually doesn't seem necessary.

You mean the candidate lookup? While that's true, it is not that relevant considering the constants. Allocating all the strings is a different story.

> The whole approach of passing the lines through the matcher (again) to find the highlight ranges seems very roundabout.

It is the best we came up with while sticking to the exposed APIs.

> Would something like this be possible?

Yes, with some significant hacking, since you can do almost everything with advices.

I think I didn't emphasize this enough - I am happy with the status quo. It nicely highlights the current line I am looking at. I don't miss the additional highlights. Note that consult-line also has other issues like #281, which are difficult to fix if one doesn't want to hack everything deeply. If you are absolutely relying on preserving the Swiper behavior, why not continue using it?

If it is possible to figure out a generalization which fits nicely into the given framework I am happy to consider it of course. There are two factors at play here - one side is the UI (which you want to improve) and one side is the implementation, which I want to keep simple and which should preserve API boundaries. The goal is to avoid entangling the UI and Consult.

muirdm: > Somehow hook the minibuffer completion matching to record the match ranges as an explicit string property.

Probably can't do this in general, but looks like I can do it easily enough w/ orderless.

muirdm: >  If you are absolutely relying on preserving the Swiper behavior, why not continue using it?

It's a good question. I was never fully satisfied w/ Swiper, even after a bunch of hacking. In general I believe with a few tweaks a line based approach like swiper or consult-line can cover 95% of my use of isearch with an overall improved UX.

PR #813, opened 2023-06-11
Add user option `consult-kmacro-runner`.

okamsn:
Hi, Minad.

I am trying out the package Meow, which can play back kmacros at multiple locations using its ["Beacon" state](https://github.com/meow-edit/meow/blob/master/TUTORIAL.org#beacon). Meow uses its own commands for running the last defined kmacro at the "beacons", so it doesn't automatically work with `consult-kmacro`.

This PR adds a way to pass the kmacro selected by `consult-kmacro` to another function that would run the kmacro. While I intend to use this with Meow, it feels reasonable to assume that other packages, maybe Multiple Cursors, would also benefit.

Would you be open to adding something like this?

The function receives two arguments: the chosen keyboard macro (as a function) and the prefix argument of `consult-kmacro`.

### DIFF
--- a/consult-kmacro.el
+++ b/consult-kmacro.el
@@ -30,6 +30,14 @@
 
 (defvar consult-kmacro--history nil)
 
+(defcustom consult-kmacro-runner nil
+  "A function used to run the keyboard macro chosen by `consult-kmacro'.
+
+This function receives two arguments: the macro (as a function)
+and the prefix argument of `consult-kmacro'."
+  :group 'consult
+  :type '(choice function (const nil)))
+
 (defun consult-kmacro--candidates ()
   "Return alist of kmacros and indices."
   (thread-last
@@ -66,7 +74,10 @@
   "Run a chosen keyboard macro.
 
 With prefix ARG, run the macro that many times.
-Macros containing mouse clicks are omitted."
+Macros containing mouse clicks are omitted.
+
+When `consult-kmacro-runner' is non-nil, it is used
+to run the chosen keyboard macro."
   (interactive "p")
   (let ((km (consult--read
              (or (consult-kmacro--candidates)
@@ -82,6 +93,7 @@ Macros containing mouse clicks are omitted."
              :lookup #'consult--lookup-candidate)))
     (unless km (user-error "No kmacro selected"))
     (funcall
+     (or consult-kmacro-runner #'funcall)
      ;; Kmacros are lambdas (oclosures) on Emacs 29
      (if (eval-when-compile (> emacs-major-version 28))
          km


minad:
Is it not possible that we just use the standard mechanisms, where Meow can hook in? Can this be implemented in Meow instead, in a way which is independent of Consult?

okamsn:
I don't think that it can be done independent of Consult, because `consult-kmacro` is the one selecting the macro and running it as a function. As far as I know, the information is not available outside of the command. I found `kbd-macro-termination-hook`, but it is not aware of the kmacro that was run. It seems that `executing-kbd-macro` is nil by the time the hook is run. I agree that this kind of hook would be better than a configuration option.

What if `consult-kmacro` returned the kmacro function as it's return value? That would be sufficient for advice using `:around`.

```el
(defun consult-kmacro (arg)
  "Run a chosen keyboard macro.

With prefix ARG, run the macro that many times.
Macros containing mouse clicks are omitted."
  (interactive "p")
  (let ((km (consult--read
             (or (consult-kmacro--candidates)
                 (user-error "No keyboard macros defined"))
             :prompt "Keyboard macro: "
             :category 'consult-kmacro
             :require-match t
             :sort nil
             :history 'consult-kmacro--history
             :annotate
             (lambda (cand)
               (get-text-property 0 'consult-kmacro--annotation cand))
             :lookup #'consult--lookup-candidate)))
    (unless km (user-error "No kmacro selected"))
    ;; Kmacros are lambdas (oclosures) on Emacs 29
    (let ((fn (if (eval-when-compile (> emacs-major-version 28))
                  km
                (kmacro-lambda-form km))))
      (funcall fn arg)
      ;; Return value for any surrounding advice.
      fn)))
```

minad:
Can you please explain why Meow needs to execute some code after macro expansion? How does such an advice look? How does Meow execute this post execution code if kmacros are evaluated using the standard built-in commands?

okamsn:
Currently, Meow uses its own commands to run the last defined kmacro. These
commands in turn use the default Emacs commands. For example, this is what Meow
recommends binding to `F4`:

```elisp
(defun meow-end-or-call-kmacro ()
  "End kmacro recording or call macro.

This command is a replacement for built-in `kmacro-end-or-call-macro'."
  (interactive)
  (cond
   ((and meow--keypad-this-command defining-kbd-macro)
    (message "Can't end kmacro with KEYPAD command"))
   ((eq meow--beacon-defining-kbd-macro 'record)
    (setq meow--beacon-defining-kbd-macro nil)
    (meow-beacon-end-and-apply-kmacro))
   ((or (meow-normal-mode-p)
        (meow-motion-mode-p))
    (call-interactively #'kmacro-end-or-call-macro))
   (t
    (message "Can only end or call kmacro in NORMAL or MOTION state."))))
```

When using the Beacon state (its version of multiple cursors), it runs

```elisp
(defun meow-beacon-end-and-apply-kmacro ()
  "End or apply kmacro."
  (interactive)
  (call-interactively #'kmacro-end-macro)
  (meow--beacon-apply-kmacros))
```

in which `meow--beacon-apply-kmacros` reduces to calling `kmacro-call-macro` at
each beacon (but not at point).

In some cases, when using beacons and switching to Meow's Insert state, Meow
automatically starts recording a kmacro and plays it back at each beacon
automatically once the Insert state is exited.

To play back a kmacro, which isn't the last one defined, at the beacons, one
first has to rotate the kmacro ring.  As you are aware, this is less convenient
than `consult-kmacro`.

The advice I am imagining would be similar to `meow-end-or-call-kmacro`, and
would look something like this:

``` emacs-lisp
(defun my-meow-advice-consult-kmacro (fn &optional arg)
  (when (eq meow--current-state 'beacon)
    (meow--switch-state 'normal)
    ;; Call kmacro at cursor, store the function form,
    ;; then run a wrapping command at each beacon.
    (let ((kmacro (funcall fn arg)))
      (meow--beacon-apply-command
       (lambda () (interactive) (funcall kmacro arg))))
    (meow--switch-state 'beacon)))

(advice-add 'consult-kmacro :around #'my-meow-advice-consult-kmacro)
```

I don't know of a way for Meow to get the kmacro when it is run for code that
executes after. When I tried using `kbd-macro-termination-hook`,
`executing-kbd-macro` was nil.

Does this answer the question?

minad:
I think you can already solve your issue by providing a `:state` function for `consult-kmacro`. Does this work for you?

~~~ elisp
(consult-customize
 consult-kmacro
 :state (lambda (action cand)
          (when (and (eq action 'return) cand)
            (run-at-time 0 nil (lambda ()
                                 ;; Run the macro Meow style...
                                 (funcall cand)))
            (signal 'quit nil))))
~~~

okamsn:
Thanks! This is what I am using:

```elisp
(consult-customize
 consult-kmacro
 :state (lambda (action cand)
          (when (and (eq action 'return)
                     cand
                     (meow-beacon-mode-p))
            (run-at-time
             0 nil
             (lambda ()
               (with-undo-amalgamate
                 ;; Following `meow-beacon-apply-kmacro'.
                 (meow--switch-state 'normal)
                 (let ((arg current-prefix-arg))
                   (funcall cand arg)
                   (meow--beacon-apply-command
                    (lambda () (interactive) (funcall cand arg))))
                 (meow--switch-state 'beacon))))
            (signal 'quit nil))))
```

PR #825, opened 2023-07-30
consult-register--describe: Accept buffer as a buffer object

EugeneNeuron:
* consult-register.el (consult-register--describe): Accept buffer as a buffer object.

`(cdr val)` can be a string or buffer object, since `cl-defmethod register-val-jump-to ((val cons) delete)` in register.el calls:
```
((eq (car val) 'buffer)
    (switch-to-buffer (cdr val)))
```
and `cl-defmethod register-val-describe ((val cons) verbose)` calls:
```
((eq (car val) 'buffer)
    (princ "the buffer ")
    (prin1 (cdr val))
    (princ "."))
```

### DIFF
--- a/consult-register.el
+++ b/consult-register.el
@@ -90,7 +90,7 @@ Each element of the list must have the form (char . name).")
 
 (cl-defmethod consult-register--describe ((val (head buffer)))
   "Describe buffer register VAL."
-  (list (propertize (cdr val) 'face 'consult-buffer)
+  (list (propertize (format "%s" (cdr val)) 'face 'consult-buffer)
         'consult--type ?f 'multi-category `(buffer . ,(cdr val))))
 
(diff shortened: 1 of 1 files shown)


minad:
Hi! It seems storing buffer objects in the `register-alist` contradicts the documentation. See https://github.com/emacs-mirror/emacs/blob/1f3995f65a065a28e108653128b31a2fb7eeb01c/lisp/register.el#L72, https://www.gnu.org/software/emacs/manual/html_node/elisp/Registers.html and https://www.gnu.org/software/emacs/manual/html_node/emacs/File-and-Buffer-Registers.html. It seems buffer objects are only supported by accident. How does it happen that you end up with buffer objects in your `register-alist`? As far as I understand `buffer` registers are not created by register saving commands, both `buffer` and `file` registers are mainly meant to be manually configured to get some quick access keys.

PR #826, opened 2023-08-02
consult--find-file-temporarily-1 : improve binary detection heuristic

aagon:
Hi Daniel,

To decide whether to preview a file, various consult commands use `consult--find-file-temporarily-1`. This function contains a heuristic to decide whether the file to be previewed is binary.

The heuristic is the following : if the null character ("\0") is present at all in the file, then the file is deemed binary and the preview is cancelled.

This heuristic is too violent, and presents too many false positives. I met such a false positive when trying `consult-ripgrep` on a directory that contained multiple info files (the directory info of the latest emacs release, to be specific, that contains some very large info files). Those files are very much text, with some rare null characters here and there. It would be nice to be able to preview those when using `consult-ripgrep`.

In fact, for real binary files, like `/usr/bin/bash`, those null characters are going to be there from the first bytes. It is therefore possible to limit the search to the first bytes (I chose 40 arbitrarily), which will remove false positives without increasing false negatives, improving the accuracy of the heuristic. Another benefit is that the search is now constant in the size of the file.

With this patch, I am now able to preview correctly text files that were previously incorrectly deemed binary. Feel free to use this.

Best,

Aymeric

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -1270,7 +1270,7 @@ ORIG is the original function, HOOKS the arguments."
           (with-current-buffer buf
             (if (save-excursion
                   (goto-char (point-min))
-                  (search-forward "\0" nil 'noerror))
+                  (search-forward "\0" 40 'noerror))
                 (progn
                   (kill-buffer buf)
                   (format "Binary file `%s' not previewed literally" name))


minad:
Hi Aymeric! 

Thanks for the suggestion. It would be great if null bytes generally appear early in "binary" files, but for example this is not the case with file formats like PDF. Therefore I want to keep the existing heuristic. Note that the linear behavior is not relevant, since we already created the buffer in linear time.

For files larger than `consult-preview-raw-size`, Consult has basically already given up on doing a proper preview. If you don't want this you can set the variable to a larger value. 

Nevertheless it would be nice to improve preview and make it more flexible. I recently talked with @protesilaos about creating a separate filepreview backend library, which provides fine-grained preview handlers. There can be handlers like image thumbnailers or a handler which extracts a subset of a large file, obsoleting the `consult-preview-raw-size` mechansim. Such a package could be useful for Consult, Dired-Preview, Dirvish, etc. Developing such a more sophisticated set of preview handlers is out of scope of Consult and can potentially benefit other packages, so a separate library is the way to go.

aagon:
> It would be great if null bytes generally appear early in "binary" files, but for example this is not the case with file formats like PDF.

Yes, I did not think about that. I saw "binary", I assumed ELF, my mistake.

> For files larger than `consult-preview-raw-size`, Consult has basically already given up on doing a proper preview. If you don't want this you can set the variable to a larger value.

I should have done this.

>Nevertheless it would be nice to improve preview and make it more flexible. I recently talked with @protesilaos about creating a separate filepreview backend library, which provides fine-grained preview handlers. There can be handlers like image thumbnailers or a handler which extracts a subset of a large file, obsoleting the consult-preview-raw-size mechansim. Such a package could be useful for Consult, Dired-Preview, Dirvish, etc. Developing such a more sophisticated set of preview handlers is out of scope of Consult and can potentially benefit other packages, so a separate library is the way to go.

I assume that this future library will need some way of distinguishing file types to decide what preview handler to use. When I was searching for a better heuristic than yours earlier, I was surprised by how nontrivial it is to decide whether a file is binary or not. What you're proposing is even more specific than this, akin to what something like `file(1)` does. Did you plan on using it ? Just for information, on my earlier example, it was incapable to recognise `elisp.info` as text, instead classifying it as "data", probably just for the odd null character in the middle of the file, just like your heuristic. However `isutf8(1)`, from moreutils, was capable of recognising `elisp.info` as text.

Anyway, sorry for the bad PR.

Best,

Aymeric

minad:
No worries, the PR addresses a real problem.

> I assume that this future library will need some way of distinguishing file types to decide what preview handler to use.

The idea is to use a heuristic based on the file name or the built-in magic mechanism, i.e., `magic-mode-alist`. Unfortunately I am a bit too busy these days to get started on such a `filepreview` library, but may have more time in two or three months.

PR #836, opened 2023-09-01
Allow custom sort function. Add one based on path commonality for xref

ovy:
When working with large repos, both tag-based xref and git can generate loads of matches that aren't at all relevant. Add a display-sort-function, `consult-sort-path-commonality` that looks at how many path elements are in common between each candidate and the last buffer with a filename, something we can do pretty fast.

Pass `:sort` configs that are not `t` as display/cycle sort functions to allow consult-customization.

Use the new function for xref where it's a good match without any downsides IMHO.

#### Algorithm/speed considerations

Other than code simplicity, another factor considered is that this can run on tens of thousands of files, and repeatedly for git-grep [async]. Many clever things won't work well at scale, e.g.  longest contiguous substring or making paths absolute. Instead, we rely on the regex engine and count the matches on any of the reference components without even extracting substrings. E.g. this is 5x faster than even a similarly naive path-components-hash-count. It works well enough in practice for focusing xref or grep in a large project to the closest related files, to the point that xref even when tag-based becomes magically good as opposed to usually needing refinement.  [I could swear the old tags had something similar, but I can't find this anymore].

I expect this sort will be used for both relative and absolute path producing commands, so the code makes no assumptions about how relative the paths are and relative to what.

As for speed, it sorts 65K path matches from a intentionally vague search -- for commonality with a "current path" 15 directories deep -- in 0.4 seconds on my Macbook. This is IMHO fast enough.

Why it's not the default for git-grep and locate:  in async commands that do take a while, sorting can be slightly disconcerting as new paths can pop at the top of the tree when the finder gets close to current. While those are often the most interesting results, it might not be to everyone's taste. locate is better than git-grep in this regard just because it's faster. Anyway, changing the defaults or customizing is simple enough. I do use it for both myself.

One reason the sort is stable, by commonality only, as opposed to "then by alpha", is to minimize the jumping when we're not actually finding any results with common path elements. Easy enough to presort with a secondary criterion if needed.

I hereby and hereafter assign the copyright to this chunk of code to the FSF.

### DIFF
--- a/README.org
+++ b/README.org
@@ -980,7 +980,7 @@ Useful options are:
 - =:default= set the default value
 - =:history= set the history variable symbol
 - =:add-history= add items to the future history, for example symbol at point
-- =:sort= enable or disable sorting
+- =:sort= enable (t) or disable (nil) sorting, or specify a custom =display-sort-function=
 - =:group= set to nil to disable candidate grouping and titles.
 - =:inherit-input-method= set to non-nil to inherit the input method.
 
--- a/consult-xref.el
+++ b/consult-xref.el
@@ -102,7 +102,7 @@ FETCHER and ALIST arguments."
             :prompt "Go to xref: "
             :history 'consult-xref--history
             :require-match t
-            :sort nil
+            :sort #'consult-sort-path-commonality
             :category 'consult-xref
             :group #'consult--prefix-group
             :state
--- a/consult.el
+++ b/consult.el
@@ -2537,8 +2537,9 @@ PREVIEW-KEY are the preview keys."
                                 . ,(apply-partially #'consult--read-affixate annotate))
                                (annotation-function
                                 . ,(apply-partially #'consult--read-annotate annotate))))
-                         ,@(unless sort '((cycle-sort-function . identity)
-                                          (display-sort-function . identity)))))
+                         ,@(unless (eq sort t)
+                             `((cycle-sort-function . ,(or sort 'identity))
+                               (display-sort-function . ,(or sort 'identity))))))
              (consult--annotate-align-width 0)
              (result
               (consult--with-preview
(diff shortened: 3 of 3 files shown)


minad:
Hi,

thanks for your suggestion! I am not sure if we should add an opinionated sorting method on top of the default xref sorting. Did you consider proposing your sort method upstream to the xref maintainer instead as an option?

> Pass :sort configs that are not t as display/cycle sort functions to allow consult-customization.

I've avoided this generalization so far since it is better/faster to specify `:sort nil` and pass the presorted list of candidates to `consult--read`. This way one avoids resorting the list of candidates repeatedly in the completion UI. Note that completion UIs first filter and then sort.

However if one still wants to override the sort function one can do that by configuring the completion UI. For example via the variable `vertico-sort-override-function` in Vertico.

ovy:
Hi Daniel and thanks for the consideration.

> I am not sure if we should add an opinionated sorting method on top of the default xref sorting. Did you consider proposing your sort method upstream to the xref maintainer instead as an option?

I also need it for consult-git-grep and consult-locate.  So I would also need to talk to the maintainer of those 😉      And come up with customization mechanisms for each.  This seems like a relatively general concept that benefits from a general customize like we have here.

I can also trim the PR to just passing through a sort function and keep the sorting function for myself ... bit of a pity.

> 
> > Pass :sort configs that are not t as display/cycle sort functions to allow consult-customization.
> 
> I've avoided this generalization so far since it is better/faster to specify `:sort nil` and pass the presorted list of candidates to `consult--read`. This way one avoids resorting the list of candidates repeatedly in the completion UI. Note that completion UIs first filter and then sort.

That may be an option for precomputed-list commands, again with significant effort command-by-command. I don't see it as an option for consult async commands which don't have a stable intermediate API -- I would pretty much need to keep variants of each. And they're complicated...

I did not realize that you disfavored display-sorts for this reason. The concept seems central in Vertico -- and frankly the history-aware sort is the one reason I was finally able to tolerate a list-style M-x, which would in previous attempts present me with a list of irrelevant 2C commands.  For what it's worth, I tried postfiltering on my 65K path testcase and it's still usable even when the postfilter intentionally doesn't filter out much. Once it does, you don't notice any sorting delay at all.

Maybe the best thing to do is to have a customize-presort for at least the async commands?   From a performance perspective, the fastest thing would actually be to have a bucketing function and keep the buckets open, then just add to them to sort incrementally.

> 
> However if one still wants to override the sort function one can do that by configuring the completion UI. For example via the variable `vertico-sort-override-function` in Vertico.

Sure, but there is no one-liner like consult-customize, I have to wrap or advice everything.

In summary, can I reduce to just passing through a display sort?   The sort config key and the completion metadata are both right there and they just need to meet.

Thanks again for your time  🙏

minad:
My point is that the display-sort-function is not a feature I want to use or expose in Consult. I am quite critical about it, but of course there exist a few reasons for it's existence: 1. Sorting after filtering is faster if most candidates have already been filtered out (this is an assumption of the default ui, but not of vertico-style uis), 2. Sorting the candidate list (produced by completion-all-completions) works for arbitrary completion tables (hashes, obarrays, functions, ...), 3. Late sorting can be more fancy, depending on input for example like fuzzy sorting.

However in scenarios where we have the candidates already available like in consult-xref, we should sort the candidates once before passing them to completing-read/consult--read. Ideally the same should be done for asynchronous completion with bucketing as you say.

We already have vertico-sort-function or vertico-sort-override-function which can be set relatively conveniently via vertico-multiform-mode. This mechanism is UI-specific, but more general than offering customizability via consult-customize only for Consult commands. Therefore I don't think exposing display-sort-function via Consult is advantageous. One could argue to expose it just out of principle and for consistency with completing-read, but my worry is that it will lead to inefficient code - sorting once is just better.

One could discuss adding some alternative presort function to Consult, but then the design is unclear because of the different completion table types.

ovy:
Thanks for the feedback (sorry, was away for a couple days).

I'll try to make it work with multiform then, and maybe later take a look at what async bucketsorting would entail.

minad:
> I'll try to make it work with multiform then, ...

Yes, multiform and `vertico-sort-override-function` should be as good (or even better, more general) as adding customizability of post-sorting here.

> ...and maybe later take a look at what async bucketsorting would entail.

Sure go ahead with experiments, regarding async or regarding configurable pre-sorting per command. For async sorting I see the problem that candidates pop up unexpectedly which may be so confusing that the addition wouldn't be perceived as an improvement. For pre-sorting other commands there is the aforementioned problem that it somehow has to work for different completion table types.

I just want to avoid to create false expectations. I'd consider such functionality on the prototype/experimental level and it is not clear if we really add something like this here in the end. I hope that's understandable.

PR #916, opened 2024-01-12
consult-register: Set register-command-info (Emacs 30)

slotThe:
Emacs recently changed how register jumping/loading is handled. The variable `register-use-preview` may be used to adjust that behaviour; it's read and acted upon by (e.g.) `register-read-with-preview`. That function (which we call inside of `consult-register-load`) checks which command called it and gets some information about that via the `register-command-info` defmethod. If (say) the command indicates that it just wants to jump to some register, an additional confirmation is skipped when `register-use-preview` is set to an appropriate value.

Thus, to completely restore how Emacs used to handle register, we need `consult-register-load` to signal that all it does it jump to the register.

---

N.b.: I didn't change the changelog yet, as I don't know if this qualifies as something that should be in there.

### DIFF
--- a/consult-register.el
+++ b/consult-register.el
@@ -233,6 +233,15 @@ for the meaning of prefix ARG."
      (unless (string-search "access aborted" (error-message-string err))
        (insert-register reg (not arg))))))
 
+;; When `register-read-with-preview' is called, and `register-use-preview' is
+;; set to nil or `never', it checks whether the command calling it just wants
+;; to jump to a register, and skips an additional confirmation in that case.
+;; This restores the register behaviour of earlier versions of Emacs.
+(when (>= emacs-major-version 30)
+  (cl-defmethod register-command-info ((_command (eql consult-register-load)))
+    "Register command info for `consult-register-load'."
+    (register-command-info #'jump-to-register)))
+
 (defun consult-register--action (action-list)
   "Read register key and execute action from ACTION-LIST.
 


minad:
Thanks for this. I noticed the debates regarding the register changes. I am not sure if everything is settled regarding that matter?

For now I won't merge the change. I will do so after I had time to test the new register behavior and the interaction with Consult. This will likely happen when the emacs-30 branch is created. Do you have an idea when that might happen? Then I will also bump Compat to compat-30 and finalize Emacs 30 support in my packages.

slotThe:
> Thanks for this. I noticed the debates regarding the register changes. I am not sure if everything is settled regarding that matter?

Yeah it was quite a heated discussion, but people seem to have cooled off a bit now and [this commit](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=67e16d37e9c83fea9f67d144eeac27a83d52c949) is—to my knowledge—the end of that story. There is actually another option to `register-use-preview` now (which I missed at first): `traditional`. That one does not require downstream packages to change and essentially restores the old Emacs 29 functionality. However, the `register-command-info` is still used for other values of `register-use-preview`. Some of these, like a value of `nil` (how descriptive), add some eye-candy and so I think downstream packages should still adjust and support that use-case.

> For now I won't merge the change. I will do so after I had time to test the new register behavior and the interaction with Consult. This will likely happen when the emacs-30 branch is created. Do you have an idea when that might happen? Then I will also bump Compat to compat-30 and finalize Emacs 30 support in my packages.

Okay, fair enough, thanks for considering! I have no idea when emacs-30 is going to be cut, but I'd imagine that it'll take a few more months.

minad:
I've just migrated to Emacs 30. Is this PR still needed? I don't see any change in behavior with `register-use-preview=nil`.

slotThe:
> Is this PR still needed?

I believe so. Without this change, even setting `register-use-preview=nil`, one has to give Emacs an additional confirmation when using `consult-register-load`.
E.g., given a register on `a`, with this change (and before Emacs 30) one would just have to execute `consult-register-load a` to jump to the register, while on Emacs 30 without this change it's now `consult-register-load a RET`.

minad:
It seems to me that the change is unnecessary. One can simply set `register-use-preview` to `traditional` (the default) and everything keeps working as is. What is the advantage of setting `register-use-preview` to `never` or `nil`?

minad:
Well, I don't agree and I see why the change has been controversial. There seem to be zero advantages if you set `register-use-preview=nil` over traditional mode with a delay. The traditional behavior cannot be legacy and is here to stay, since otherwise you cannot use a non-zero delay. I think traditional+delay is the best behavior, similar to which-key if you  hesitate for too long pressing the next key. Also note that Consult offers both commands for immediate exit (consult-register-load) and with navigation via completing-read (consult-register), so even `register-use-preview=t` is not adding anything new.

minad:
Okay, I close this for now. I will reconsider, in the case I discover that there are actual advantages of the non-traditional register UI. To be clear, the non-traditional UI works if used as intended (preview=t). The only settings which are affected are preview=never and preview=nil, but as far as I understand they are some kind of half-way compromise between preview=traditional and preview=t.

PR #970, opened 2024-03-13
Fix register preview for position register with nil buffer

nick4f42:
The consult register preview would error when a register entry was `#<marker in no buffer>`, which could happen if you killed the buffer a marker pointed to.

### DIFF
--- a/consult-register.el
+++ b/consult-register.el
@@ -65,16 +65,19 @@ Each element of the list must have the form (char . name).")
 
 (cl-defmethod consult-register--describe ((val marker))
   "Describe marker register VAL."
-  (with-current-buffer (marker-buffer val)
-    (save-excursion
-      (without-restriction
-        (goto-char val)
-        (let* ((line (line-number-at-pos))
-               (str (propertize (consult--line-with-mark val)
-                                'consult-location (cons val line))))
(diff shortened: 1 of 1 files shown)


minad:
Hi, thanks for this!

> ...which could happen if you killed the buffer a marker pointed to.

When does this happen? Marker registers are usually swapped out with file-query registers when the corresponding buffer is killed. See:

https://github.com/emacs-mirror/emacs/blob/30bc867aecc59265b6e315acf459f8d79c423bca/lisp/register.el#L623-L624

nick4f42:
If I create a non-file-visiting buffer "foo", create a mark register in it, then kill "foo", I see the issue. I didn't realize there was a hook to update the register when the buffer is killed, but it looks like it only handles file-visiting buffers:

https://github.com/emacs-mirror/emacs/blob/30bc867aecc59265b6e315acf459f8d79c423bca/lisp/register.el#L730

Maybe Emacs should delete the register if there's no file, but I still think consult should handle the nil buffer case to be safe.

minad:
Okay, thanks. Fixed in b48ff6bf0527baeb6bfd07c6da9d303ff0b79c3d. I decided to filter out such invalid registers early on, since they can be considered empty.

nick4f42:
Thanks!

Issue #978, opened 2024-03-22
Handling buffer renames while switching buffers


Stebalien:
If a buffer is renamed after running `consult-buffer`, attempting to switch to it will open a new blank buffer with the original name. This is especially problematic in my EXWM setup, where previewing a buffer can change the buffer's name (e.g. previewing a slack window clears the unread messages count from the title).

Would it be possible to somehow record the original buffer name to buffer mapping?

PR #980, opened 2024-03-22
Use (cons 'buffer . buffer-object) as multi-category for candidates

minad:
This fixes the problem where buffers are renamed during completion. Renaming can happen during buffer name uniquify or due to EXWM renaming. See the issues #978 and #979.

Consult buffer preview magically continues to work since `get-buffer` accepts both string and buffer arguments. I haven't tested Embark yet.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -2,6 +2,11 @@
 #+author: Daniel Mendler
 #+language: en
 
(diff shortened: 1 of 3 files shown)


minad:
This needs testing and coordination with Embark before it can be merged.

PR #1021, opened 2024-05-21
Add file preview to `consult-find`

ErnestKz:
This is such that triggering the `:preview-key` of a `consult-find` will preview the file, as it does for `consult-recent-file`

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -4978,6 +4978,7 @@ INITIAL is initial input."
      :file-handler t) ;; allow tramp
    :prompt prompt
    :sort nil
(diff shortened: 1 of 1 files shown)


minad:
No, such a preview is usually undesired. consult-find searches by name while preview shows content, in contrast to consult-grep. But there is no problem - you can use consult-customize to set a state function for consult-find.

ErnestKz:
Ah! Now I see.  `consult-customize` modifies the settings that are passed to `consult--read`. What a nice design. Thank you

Issue #710, opened 2023-01-07
consult-theme but for fonts?


nrvale0:
Before I even ask for this let me say that I'm looking at whether I could write it and send it in as PR. But my elisp is ... well ... I'm getting there slowly. ;)

It would be super helpful if consult had a function like consult-theme but for previewing fonts instead of previewing themes.

minad:
I am not sure if that would be a good addition to Consult. I think one could instead add a Marginalia annotator, which displays a font sample as annotation. This annotator could then be used for the commands `describe-font`, `set-face-font`, `set-frame-font` and maybe other commands.

nrvale0:
There are a handful of packages out there which allow limited preview of fonts but, like some of the earlier attempts at previewing themes, they don't give quite the same of understanding what to expect that one gets from =consult-theme= auto-preview using open buffers.

I'll spin up a consult ecosystem package to try to hit my use-case and, if I'm successful, open a new Issue to discuss whether it should remain a separate package or be mainlined back into consult itself.

minad:
I understand that real preview would be different from annotators. But still, I think the command would not provide sufficient value to add it here. It is not that I am switching fonts ten times a day, in contrast to other Consult commands.

PR #1044, opened 2024-06-19
Add function consult-font to configure a font using font-family-list

erickgnavar:
Allow to select a new font and also support preview during scrolling

### DIFF
--- a/README.org
+++ b/README.org
@@ -379,10 +379,13 @@ their descriptions.
 
 #+findex: consult-completion-in-region
 #+findex: consult-theme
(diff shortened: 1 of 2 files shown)


erickgnavar:
I know it was addressed in #710 but it could be helpful for someone to have previews of different fonts using `consult`

minad:
Thanks for the contribution, but I've rejected this feature in #710. Consult is already quite stuffed with features and I don't think it is reasonable to add a rarely used command like `consult-font`. Feel free to add the command to the wiki.

PR #1045, opened 2024-06-20
add consult-rg-dir for rg search under directory, with symbol under cursor

liuzhishan:
Hi, I found that I often need to search under a directory, with the symbol under cursor. It's a very common situation. But after some digging, I didn't find a function to do the work. So I add a new function `consult-rg-dir` to do the job:

```
;;;###autoload
(defun consult-rg-dir ()
  "Search with `rg` for files in prompted DIR with symbol under cursor."
  (interactive)
  (consult-ripgrep (read-directory-name "rg directory: ") (thing-at-point 'symbol)))
```

It will prompt for directory, which can be chosen by user, and then do `rg` search using symbol under the cursor.

Hope It will be helpful.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -4960,6 +4960,12 @@ See `consult-grep' for details."
   (interactive "P")
   (consult--grep "Ripgrep" #'consult--ripgrep-make-builder dir initial))
 
+;;;###autoload
+(defun consult-rg-dir ()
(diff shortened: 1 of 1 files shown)


minad:
You can use Embark instead to run `consult-ripgrep` on the symbol at point. Press `C-. M-s r` to do that.

liuzhishan:
Thanks for the advice. I'll thy this command.

PR #1069, opened 2024-08-17
consult-project-grep added, fixed consult--with-project

HarmVanDerVegt:
Summary: Used to using the project bindings from anywhere, wanted to use consult-grep in the same manner as project-find-regexp, prompting for a project root if not in one.

This PR consists of two changes, one adding consult-project-grep and one fixing the consult--with-project macro.

Issue with consult--with-project macro:

The macro errors when `consult-project-function` is needed in the body. Steps to reproduce:

1. Have consult installed.
2. Have an available project.
3. `emacs -q`
4. Evaluate `(require 'package)`, `(package-initialize)`, `(require 'consult)`
5. Open a scratch buffer
6. Paste the following function:
`(defun consult-project-grep ()
  (interactive)
  (consult--with-project
   (consult--grep "Grep" #'consult--grep-make-builder nil nil)))
`
7. Evaluate the buffer.
8. Execute `consult-project-grep` and select a valid project.
9. You should encounter the following error: `if: Symbol’s value as variable is void: depth`.

If `depth` is provided, it will fail on `root`. 

The provided fix allows `consult-project-grep` to be called. This can be tested by including the modified `consult--with-project` in the scratch buffer an evaluating it again.

I've checked the code, `consult--with-project` is used in one place, for `consult-project-buffer`. I've run this function after applying the change, it all still works as expected.

Annoyance: `C-x p p` -> `g Find regexp` does not call `consult-project-grep`, leading to inconsistent behavior. Unsure how to tackle this.

Future plans: I´d like to add `consul-project-find` to act in the same way, prompting for a project root if note in one already, this goes beyond my current capabilities, sadly.

This is my first PR for an Emacs package (or PR outside of work in general), if I've missed something, please let me know. I have not yet signed the copyright over to the FSF yet, since I wanted to see first if this PR is wanted.

### DIFF
--- a/README.org
+++ b/README.org
@@ -240,6 +240,7 @@ their descriptions.
 #+cindex: locate
 
 #+findex: consult-grep
+#+findex: consult-project-grep
 #+findex: consult-ripgrep
 #+findex: consult-git-grep
 #+findex: consult-find
@@ -262,6 +263,7 @@ their descriptions.
   [[#project-support][project directory]] if a project is found. Otherwise the =default-directory= is
   searched. If =consult-grep= is invoked with prefix argument =C-u M-s g=, you can
   specify one or more comma-separated files and directories manually.
+- =consult-project-grep=: Variant of =consult-grep=, prompting for a project if not in one.
 - =consult-find=, =consult-fd=, =consult-locate=: Find file by matching the path
   against a regexp. Like for =consult-grep=, either the project root or the
   current directory is the root directory for the search. The input string is
@@ -779,6 +781,7 @@ configuration examples.
          ("C-x t b" . consult-buffer-other-tab)    ;; orig. switch-to-buffer-other-tab
          ("C-x r b" . consult-bookmark)            ;; orig. bookmark-jump
          ("C-x p b" . consult-project-buffer)      ;; orig. project-switch-to-buffer
+         ("C-x p g" . consult-project-grep)	     ;; orig. project-find-regexp
          ;; Custom M-# bindings for fast register access
          ("M-#" . consult-register-load)
          ("M-'" . consult-register-store)          ;; orig. abbrev-prefix-mark (unrelated)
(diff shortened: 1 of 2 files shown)


minad:
Hi! 

Thanks for your contribution. 

Your fix to `consult--with-project` doesn't seem to be necessary in my tests. The fix should be mostly equivalent, so I don't see how it makes a difference. Maybe you evaluated without lexical binding? Could you please check this again and debug this carefully?

Regarding `consult-project-grep`, I think the idea makes sense, but it won't scale well, if we add project-specific functions to Consult for every command operating on a directory, leading to lots of boilerplate.

Instead of adding many project commands, I think one should use `project-any-command`. Did you try this? This will scale and compose better, e.g., `C-x p o M-s g`. Note that one can also use the prefix argument `C-u M-s g` to trigger directory selection, but not project selection as you prefer. Maybe we could also support `C-u C-u M-s g` (note the double prefix) and in that case trigger project selection? What do you think? The resulting UI would be the same, since one could define a keyboard command `C-x p g => C-u C-u M-s g`. What do you think?

HarmVanDerVegt:
Hey, thanks for your reply!

I'll retest `consult--with-project` on a locally modified and built version of Consult.

Agreed with the scaling comment, this might be better to let the user customize for themselves. 

Thanks for letting me know about `project-any-command`, I did now know about it yet. After trying it out, this seems very handy. Now I just need to figure out why my current build of Emacs doesn´t have it.

I really like the double prefix idea. It would be a great way to add the functionality without providing more functions. Iĺl see if I can cobble together a way to make the double prefix do what I want and make another PR.

We could update other directory command to work in the same way then.

minad:
The change turned out to be extremely simple: https://github.com/minad/consult/commit/28d60a222784515d2982be4f3aa287a789c84244

PR #1130, opened 2024-11-14
docs: add 'Hacking' section

vermiculus:
While many examples exist in the current codebase, the basic process for using a new asynchronous completion source can get lost in the sophistication of examples like `consult-rg` and friends.

Introduce a 'Hacking' section to the README that describes the bare minimum required to do this -- including live preview.

---

Address discussion #1123.


README.org
```diff
@@ -1112,6 +1112,82 @@ Please provide the necessary important information with your bug report:
...
+of a =STATE= function as defined by =consult--with-preview=. The
+=consult--define-state= macro can assist by defining a function suitable to
+provide to =consult--read='s =:state= argument.
+
+First, define a 'preview' function with a name using the format
+=consult--%s-preview=. When called with no arguments, it should return a suitable
+'state' function:
+
+#+begin_src emacs-lisp
+  (defun consult--testibus-preview ()
```
vermiculus, line 1156: What I don't necessarily like about my example is that it requires defining symbols in the `consult` namespace. What are your thoughts here?

minad: In order to avoid the namespacing problem, do not use `consult--define-state`. For a hacking section it is better to be more explicit anyway, avoiding both `consult--define-state` and `consult--state-with-return`.

minad:
Thanks, that's nice!

README.org
```diff
@@ -1112,6 +1112,86 @@ Please provide the necessary important information with your bug report:
...
+#+begin_src emacs-lisp
+  (consult--define-state testibus)
+#+end_src
+
+Finally, we can use the function generated by =consult--define-state= in
+=consult--read=:
+
+#+begin_src emacs-lisp
+  (consult--read
+   (consult--async-command (lambda (input) (list "testibus" (s-trim input))))
```
minad, line 1186: Could you also provide an example not relying on `consult--async-command`? See `consult--dynamic-collection`. Also an example based directly on the underlying async machinery (without `consult--async-command` and without `consult--dynamic-collection`) would be helpful for understanding.

README.org
vermiculus: I think I've done this correctly – at least, 'it works on my machine'. Let me know if I've missed the mark here.

README.org
vermiculus: I've added an example with `consult--dynamic-collection`.

As for a more DIY approach – I'm not sure I'm the best person to write that. To be honest, I only started messing with Consult internals yesterday :-) I'm not sure how the async machinery actually works myself.

minad:
Thanks. Merged and slightly revised in https://github.com/minad/consult/commit/1e9153dfd1d9d89a5b89c9706d405d58c03c1437.

PR #1131, opened 2024-11-14
consult-line: Enforce global completion style settings

minad:
See https://github.com/minad/vertico/discussions/533 for related discussion.

Ensure that the global completion style settings are used for `consult-line`, `consult-focus-lines` and `consult-keep-lines` filtering. This override is necessary since users may want to override the settings buffer-locally for in-buffer completion, e.g., via Corfu.

cc @jdtsmith

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -653,11 +653,18 @@ Turn ARG into a list, and for each element either:
 CATEGORY is the completion category, used to find the completion style via
 `completion-category-defaults' and `completion-category-overrides'.
 HIGHLIGHT must be non-nil if the resulting strings should be highlighted."
-  ;; completion-all-completions returns an improper list
-  ;; where the last link is not necessarily nil.
-  (nconc (completion-all-completions pattern cands nil (length pattern)
-                                     `(metadata (category . ,category)))
-         nil))
+  ;; Ensure that the global completion style settings are used for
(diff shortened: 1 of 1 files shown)


jdtsmith:
I presume this works for normal completion style variations, but the variable that is affecting me (local vs. global) is `orderless-style-dispatchers` (i.e. a variable to _configure_ a particular style).  Perhaps it would work to have a list of affected variables to `let`-bind to their default values, which could be appended to?

oantolin:
In the special case of orderless you could handle this by creating a completion style for each orderless configuration you use. Orderless come with a macro to create new completion styles based on orderless with particular settings. (I forget what it's called, but I'm sure you can easily find it).

jdtsmith:
Yep, `orderless-define-completion-style`.  You thought of everything.

minad:
See also this comment: https://github.com/minad/corfu/commit/53a5f116254500fe1883a4179b0154f219ef59b2#commitcomment-149101194

There are two options: `orderless-define-completion-style` and the new Emacs 31 let-bindings which can be specified directly in `completion-styles` or the other variables. The reason why we've added `orderless-define-completion-style` is that it wasn't possible to adjust Orderless variables for different categories. Creating new styles was the only possibility, and this is also helping us here for buffer-local Corfu completion style settings.

PR #1164, opened 2025-01-14
consult--async-split: do not insert initial split char if the minibuffer already starts with it

kassick:
When running vertico-repeat-previous, vertico will clear the minibuffer and then insert the text that was saved in history (e.g. "#some-text").

Later, when consult--async-split tries to insert the split char, the restored session ends up with "##some-text" as input -- which does not search for anything.

This commit fixes the issue by checking the minibuffer contents before inserting the initial split char.


minad:
Thanks! I pushed a slightly different fix.

PR #1165, opened 2025-01-14
Use the base-size returned by completion-all-completions correctly

catern:
Using completion-boundaries here is incorrect and doesn't work in practice for completion tables with non-trivial boundaries.  The correct thing to do is to use the base-size returned as the cdr of the last element of completion-all-completions.


minad:
Indeed. Thanks.

PR #1173, opened 2025-01-31
consult-man: use 'manual-program rather than assuming "man"

freesteph:
From the commit:

```
For the reasons listed on the original swiper issue[1], on macOS
`gman` is preferred to `man`. Emacs actually provides the
`manual-program` variable to customize that, so observe the
abstraction in consult-man as well.

[1]: https://github.com/abo-abo/swiper/issues/2836
```

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -317,7 +317,7 @@ Can be either a string, or a list of strings or expressions."
   :type '(choice string (repeat (choice string sexp))))
 
 (defcustom consult-man-args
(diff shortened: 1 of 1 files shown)


minad:
The variable `manual-program` is not available in consult.el since it is defined in man.el. Therefore we cannot use this. I suggest to override `consult-man-args` manually.

freesteph:
oh my bad, thank you.

PR #1210, opened 2025-03-31
Mark the consult-buffer command with :category 'consult-buffer

errge:
So the end-user can costumize styles (e.g. in vertico-multiform).


minad:
The consult-buffer category is `multi-category`. This is needed by Embark and Marginalia.

PR #1227, opened 2025-05-21
add MODE-INVERT keyarg to consult--buffer-query

mdsib:
Hi! I've been using this code for a couple of years to filter out `erc-mode` buffers from the rest of my buffers. I didn't see a way of achieving this without coming up with a list of every mode besides `erc-mode`, which doesn't seem practical. 

That being said, it seems kind of clunky to me to do it in this way. Let me know if this sort of feature would be welcome.

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -4683,7 +4683,7 @@ to search and is passed to `consult--buffer-query'."
                    (t "")))
           buffers)))
 
-(cl-defun consult--buffer-query (&key sort directory mode as predicate (filter t)
+(cl-defun consult--buffer-query (&key sort directory mode mode-invert as predicate (filter t)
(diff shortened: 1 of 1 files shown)


minad:
You can pass a `:predicate` argument which checks the mode. For now I don't see a reason to add `:mode-invert` (or an alternative `:excluded-modes`), but I will keep an eye on this. If it comes up again or often we can reconsider.

mdsib:
Oh yeah, looks like I missed that arg 😅 
I agree, no reason to add it since it's uncommon and can be worked around with:
```
:predicate (lambda (buf)
             (not (provided-mode-derived-p (buffer-local-value 'major-mode buf)
                                           'erc-mode)))

```

PR #1248, opened 2025-08-08
Add consult-font command

shaneikennedy:
Allows for previewing fonts before applying, similar to consult-theme

Discussion started here https://github.com/minad/consult/discussions/1247


shaneikennedy:
https://github.com/user-attachments/assets/08c29393-3e02-4602-9922-daa15b49295e

PR #1306, opened 2026-04-07
Fix quoting of backslash in regex

aikrahguzar:
This came up https://codeberg.org/rahguzar/consult-hoogle/issues/32.

Before this change:
```emacs-lisp
(consult--command-split "a \\-> b")
```
results in `"a \\-> b"`. After this change it results in `a -> b` which I think is the intended behavior.

I basically converted the current regexp to sexp form using `xr` and saw that it isn't matching on the backslash. The change is the result of micro-expansion of `(rx (group (or bos " ")) "\\-")`

### DIFF
--- a/consult.el
+++ b/consult.el
@@ -674,7 +674,7 @@ Turn ARG into a list, and for each element either:
         (setq opts (substring str (1- (match-end 0)))
               str (substring str 0 (match-beginning 0))))
       ;; Replace backslash-escaped dashes
-      (setq str (replace-regexp-in-string "\\(\\`\\| \\)\\-" "\\1-" str))
+      (setq str (replace-regexp-in-string "\\(\\`\\| \\)\\\\-" "\\1-" str))
       ;; Options end with double dash
       (when (string-match "\\(\\`\\| \\)--\\(?: \\|\\'\\)" opts)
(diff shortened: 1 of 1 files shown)


minad:
Thanks. But I think you should use a custom splitter in consult-hoogle. The command splitter really only makes sense for Unix-style commands, and not for other asynchronous commands.

aikrahguzar:
> Thanks. But I think you should use a custom splitter in consult-hoogle. The command splitter really only makes sense for Unix-style commands, and not for other asynchronous commands.

I agree a custom splitter would be better. However, I am not sure what would be a good experience. My feeling is that passing flags to hoogle is rare, so perhaps no splitting is better than the current default. However, I am not a very sophisticated hoogle user so I don't trust myself on this.

minad:
Not a user either but it seems to me that the default splitter is harmful if it treats all dashes as options, since one often writes `->` when hoogling.

aikrahguzar:
> Not a user either but it seems to me that the default splitter is harmful if it treats all dashes as options, since one often writes `->` when hoogling.

I checked the options that hoogle has and none of them seem useful dynamically so I think I will remove the splitting altogether. https://codeberg.org/rahguzar/consult-hoogle/issues/32#issuecomment-12919671

PR #1307, opened 2026-04-11
consult-imenu: Support Emacs 31 nested entries

gessen:
With Emacs 31, Eglot, and the following C++ code:

```cpp
struct foo {
  struct bar {
    struct faz {};
  };
};
```

Imenu correctly lists all nested symbols, allowing navigation to `foo`, `bar`, and `faz`. However, `consult-imenu` does not handle this properly and only exposes the innermost symbol (`faz`). This behavior appears to be language-agnostic and can be reproduced with other programming languages as well. Imenu supports nested symbols via the `imenu-region` text property, which Eglot may now provide.

In addition, Eglot may annotate its Imenu output with the `imenu-kind` property. `consult-imenu` could use this information to improve symbol categorization.

### DIFF
--- a/consult-imenu.el
+++ b/consult-imenu.el
@@ -75,6 +75,19 @@ ARGS are the arguments to the special item function."
          (list pos #'consult-imenu--switch-buffer (current-buffer)
                imenu-default-goto-function)))))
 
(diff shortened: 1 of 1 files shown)


gessen:
@minad ping

PR #1308, opened 2026-04-11
Initialize buffers on final selection via `consult-after-final-jump-hook`

alberti42:
This PR fixes an issue where buffers opened through Consult's preview mechanism remain partially initialized after they are finally selected. This prevents core Emacs features like `recentf` from correctly tracking files opened via commands like `consult-ripgrep`.

### The Technical Problem: "Half-Initialized" Buffers

1.  **Selective Hook Suppression**: To maintain high performance during live previews, Consult suppresses most `find-file-hook` members that aren't explicitly allowed in `consult-preview-allowed-hooks`.
2.  **Buffer Reuse**: When a user confirms a selection that was being previewed, Consult switches to the already-open preview buffer instead of opening a fresh one.
3.  **Missing Initialization**: Because the buffer is already active, standard Emacs hooks like `find-file-hook` do not re-run upon selection. This means side effects like updating the `recentf` history, refreshing VCS state, or running full mode-specific logic are never triggered for that file.

This creates an inconsistency where `consult-find` (which typically opens a new buffer) initializes correctly, but `consult-ripgrep` or `consult-line` (which often reuse preview buffers) do not.

### The Proposed Solution

This PR introduces a canonical mechanism to handle the transition from a "previewed" state to a "selected" state.

1.  **`consult-after-final-jump-hook`**: A new hook that fires exactly once when a candidate is finally selected in `consult--jump`. This is distinct from `consult-after-jump-hook`, which runs on every preview movement.
2.  **`consult-run-suppressed-hooks`**: A utility function provided for users who want to fully initialize their buffers upon selection. It identifies and executes only the `find-file-hook` members that were blocked during the preview phase.

### Usage Example

To opt-in to full buffer initialization, users can add this to their configuration:

```elisp
(add-hook 'consult-after-final-jump-hook #'consult-run-suppressed-hooks)
```

Alternatively, if a user only wants specific features like `recentf`, they can add them individually:

```elisp
(add-hook 'consult-after-final-jump-hook #'recentf-track-opened-file)
```

### Implementation Details

- Added `consult-after-final-jump-hook` (defcustom, hook).
- Added `consult-run-suppressed-hooks` utility function.
- Updated `consult--jump` to trigger the new hook.
- Updated docstrings for `consult-after-jump-hook` and `consult-after-final-jump-hook`.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -2,6 +2,13 @@
 #+author: Daniel Mendler
 #+language: en
 
+* Development
+
+- Add =consult-after-final-jump-hook=, which is executed when a candidate is
+  finally selected. The convenience function =consult-run-suppressed-hooks=
+  can be added to this hook to ensure that suppressed =find-file-hook=
+  members (e.g., =recentf=, =vc-refresh-state=) are executed upon arrival.
+
 * Version 3.4 (2026-03-09)
 
 - Bugfixes.
--- a/consult.el
+++ b/consult.el
@@ -170,12 +170,33 @@ nil shows all `custom-available-themes'."
   :type '(repeat (choice symbol regexp)))
 
 (defcustom consult-after-jump-hook (list #'recenter)
-  "Function called after jumping to a location.
+  "Functions called after jumping to a location.
+
+This hook is executed after every jump, including both preview
+movements during completion and the final jump after selection.
 
 Commonly used functions for this hook are `recenter' and
 `reposition-window'.  You may want to add a function which pulses the
-current line, e.g., `pulse-momentary-highlight-one-line'.  The hook
-called during preview and for the jump after selection."
+current line, e.g., `pulse-momentary-highlight-one-line'.
+
+See also `consult-after-final-jump-hook' for a hook that is executed
+only once after the final selection."
+  :type 'hook)
+
+(defcustom consult-after-final-jump-hook nil
+  "Functions called after the final jump to a location.
+
+This hook is called only once when the user officially selects a
+candidate, in contrast to `consult-after-jump-hook' which also
+fires during every preview movement.
+
+One can add functions such as `recentf-track-opened-file' or
+`vc-refresh-state' to this hook to ensure that standard Emacs hooks
+are executed upon selection.  For convenience, the utility function
+`consult-run-suppressed-hooks' is provided to automatically execute
+all hooks that were blocked during preview.  Adding this function to
+`consult-after-final-jump-hook' ensures the buffer is fully
+initialized, as when normally opening a file."
   :type 'hook)
 
 (defcustom consult-line-start-from-top nil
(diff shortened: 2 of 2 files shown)


minad:
Thanks for the suggestions. I see your point about the final jump hook - this might have come up before.

> When a user confirms a selection that was being previewed, Consult switches to the already-open preview buffer instead of opening a fresh one.

This does not sound right. Consult recreates a completely new buffer. Preview buffers have the `Preview:` prefix.

alberti42:
Thanks for the first feedback already!

>  When a user confirms a selection that was being previewed, Consult switches to the already-open preview buffer instead of opening a fresh one.

Yes, thank you for the correction. I was not sure about it. So, this statement of mine is wrong, but the PR code is still valid.

From what I understand, you put in an effort to recreate by design the same look as in the preview buffer when opening the entirely new buffer (different from the preview buffer). With this PR, I did not want to change this design but to offer the user a hook for calling, for example, `'recentf-track-opened-file`. This is something that most users, if they use recentf, would want to have.

Of course, power users already today have the option to "advise" `'consult--jump` to trigger recentf. This is how I was doing it, but I thought that this creates an unnecessary friction, and we should prefer to use canonical hooks rather than "advising" on internal functions.

In addition, I provided `'consult-run-suppressed-hooks` as an auxiliary function that the user may decide to use to trigger all other hooks that were suppressed. I don't have a strong opinion about this auxiliary function; it is a nice-to-have, but it could also be shown in the README.md as an example if you prefer to remove it from the main code.

minad:
> So, this statement of mine is wrong, but the PR code is still valid.

Not entirely. When the file is finally opened by selection (after preview), all the find file hooks are executed. So the example with the recent-track-opened-file function does not seem valid, since the function should already be called (EDIT: I confirmed that this works correctly in my setup via trace-function and by checking the recentf-list afterwards). The part of the code which is okay is the addition of the new final jump hook, however I wonder about realistic use cases for the hook.

alberti42:
Thanks.

> When the file is finally opened by selection (after preview), all the find file hooks are executed. 

I need to investigate why this does not happen on my setup.

Did you check with `'consult-ripgrep`? Because I think other paths do trigger all hooks, like you are saying. This would be the case of `'consult-find`. But the ripgrep path does not trigger the other hooks—this was for me the main problem.

minad:
Yes, of course. consult-ripgrep also triggers the hook. Let's close this for now. Feel free to continue with the discussion, and let me know about your investigations, in case there is still a need for an addition.

alberti42:
Thank you for your consideration and rapid feedback. I got it all wrong, and the problem was elsewhere in my setup. Thanks to your feedback, I could solve it now.

minad:
You're welcome. In case you discover unconfirmed issues, or if you're unsure, please open a discussion instead. Looking at a PR costs more time, and I have to spread it across many of my projects.

alberti42:
> Looking at a PR costs more time, ...

Yes, I totally understand it, and I feel embarrassed for having taken your time on a non-issue. Thank you very much for your gracious handling of it.

PR #1313, opened 2026-05-02
Add consult-org-tag

kisaragi-hiu:
This removes a tag if the chosen tag is already on the heading. This also has support for adding or removing tags from the agenda.

[螢幕錄製_20260503_001552.webm](https://github.com/user-attachments/assets/a7ec58a2-7170-40bf-adca-1945ec15e99b)

There are some minor differences from Counsel, most notably this requires at least Org>=9.0 since I'm using org-current-tag-alist directly without adding fallbacks like Counsel does. This variable was added in 2016 (Org commit 4743d43dd8) and shipped in 9.0. Counsel predated that, but this port doesn't, so I don't think that's necessary.

I used consult--read since it appears no function in this project other than consult--read itself uses completing-read directly.

---

This is more or less a port of `counsel-org-tag`. License-wise since both Consult and Counsel are FSF-assigned (part of GNU Emacs) this should be fine.

I have assigned copyright to the FSF before (for my PRs to yaml.el when it was added to GNU ELPA).

### DIFF
--- a/README.org
+++ b/README.org
@@ -343,6 +343,7 @@ their descriptions.
 
 #+findex: consult-org-heading
 #+findex: consult-org-agenda
+#+findex: consult-org-tag
 - =consult-org-heading=: Variant of =consult-imenu= or =consult-outline= for Org
   buffers. The headline and its ancestors headlines are separated by slashes.
   Supports narrowing by heading level, priority and TODO keyword, as well as live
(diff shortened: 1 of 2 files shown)


minad:
Thanks, but this is out of scope of Consult. Consult focuses on search and jump commands, not arbitrary utilities. I think this should actually be improved in Org itself. Sorry that I've not been more clear about this.

